PR #38, opened 2020-10-07
Feat/workshop attendance api

### LINKED ISSUE #10: Expose Workshop attendance api
This is similar in function to the topic progress and Recruit Project apis. 

Please make sure you write tests for the api using the correct mixin. This task is more like configuration than code, if you find yourself writing logic then you might have fallen down a rabbit hole

Rokhuda:

## Description:

## I solemnly swear that:


backend/curriculum_tracking/api_views.py
```diff
@@ -572,3 +572,30 @@ def assign_user_as_reviewer(self, request, pk=None):
...
+            permission_classes = [
+                permissions.IsAdminUser
+                | core_permissions.IsStaffUser
+                | core_permissions.IsCurrentUserInSpecificFilter("attendee_user")
+            ]
+        return [permission() for permission in permission_classes]
+
+    def get_queryset(self):
+        return models.WorkshopAttendance.objects.order_by("pk").prefetch_related(
+            "agile_card", "content_item",
```
sheenarbw: Why are you prefetching these? You dont seem to be accessing any fields from related agile card or content item instances.

backend/curriculum_tracking/api_views.py
```diff
@@ -572,3 +572,30 @@ def assign_user_as_reviewer(self, request, pk=None):
...
+
+    serializer_class = serializers.WorkshopAttendanceSerializer
+
+    filter_backends = [DjangoFilterBackend]
+    filterset_fields = ["attendee_user"]
+
+    def get_permissions(self):
+        if self.action == "retrieve":
+            permission_classes = [
+                curriculum_permissions.IsProjectAssignee
```
sheenarbw: Are you sure that these permissions are correct? 

IsProjectAssignee checks that the current logged in user is equal to the project.recruit_user.
There is no project here, and WorkshopAttendance doesn't have a `recruit_user` field.

Try GET an instance via the api. You'll get a 500 I think.

Never guess! And take names seriously, if something says it is for projects and you try to apply it to a thing that has nothing to do with projects then it probably wont work.

sheenarbw (changes requested):
Mostly good. But I see a bug and an inefficiency.

backend/curriculum_tracking/api_views.py
```diff
@@ -572,3 +572,27 @@ def assign_user_as_reviewer(self, request, pk=None):
...
+
+    serializer_class = serializers.WorkshopAttendanceSerializer
+    queryset = models.WorkshopAttendance.objects.order_by("pk")
+    filter_backends = [DjangoFilterBackend]
+    filterset_fields = ["attendee_user"]
+
+    def get_permissions(self):
+        if self.action == "retrieve":
+            permission_classes = [
+                curriculum_permissions.IsCardAssignee
```
sheenarbw: This is the permission class you are using.

```
class IsCardAssignee(BasePermission):
    def has_permission(self, request, view):
        user = request.user
        card = view.get_object()   #   a
        return user in card.assignees.all()   # b
```
In the line lebelled a, card would be made equal to the current workshop attendance instace.
In the line labelled b, we'll get an error because WorkshopAttendance does not have an attribute called assignee.

Here's what I need you to do:
1.write a unit test. In that test create a workshop attendance instance. try to GET that as a user who should not be allowed to get it. Expect a permission error
2. write a test where the user who attended the workshop GETs the workshop instance. Expect a sucessful response

When you wrrite those tests you will see that they both fail. 

Now make them pass

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -308,3 +308,22 @@ def verbose_instance_factory(self):
...
+        content.topic_needs_review = False
+        content.save()
+        return workshop_attendance
+
+    def test_get_list(self):
+        attendee_user = UserFactory(is_superuser=True, is_staff=True)
+        self.login(attendee_user)
+        url = self.get_list_url()
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200)
```
sheenarbw, line 346: your tests can be added to this class

sheenarbw (changes requested):
You're guessing...

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -308,3 +313,43 @@ def verbose_instance_factory(self):
...
+
+        workshop_attendance = factories.WorkshopAttendanceFactory()
+        worshop_attendee = workshop_attendance.attendee_user
+        
+
+        random_recruit = UserFactory(is_superuser=False, is_staff=False)
```
sheenarbw: Rather put this extra test stuff in a seperate test funciton. Use this class but different functions. Something like:

`def test_get_instance_permissions`

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -308,3 +313,43 @@ def verbose_instance_factory(self):
...
+    SUPPRESS_TEST_POST_TO_CREATE = True
+
+    def verbose_instance_factory(self):
+        workshop_attendance = factories.WorkshopAttendanceFactory()
+        content = workshop_attendance.content_item
+        content.topic_needs_review = False
+        content.save()
+        return workshop_attendance
+
+    def test_get_list(self):
```
sheenarbw: `def test_get_list(self):` is a bunction that already exists on the base class. If you write another function of the same name then you override the behavior in the base class. Is there a good reason to override the existing test functionality with tis new function?

It looks like your `test_get_list` function tests a bunch of stuff that has nothing to do with getting any lists. And it also doesn't do a bunch of the checks set out in the base class.

sheenarbw (changes requested):
This still needs work. 

Either there is a misunderstanding about OOP (do you know that you are overriding all sorts of useful behavior ?) Or maybe there is a problem with testing (you can write extra tests without overriding existing tests. Tests are just functions whose names starte with `test_`)

backend/curriculum_tracking/api_views.py
```diff
@@ -572,3 +567,25 @@ def assign_user_as_reviewer(self, request, pk=None):
...
+    queryset = models.WorkshopAttendance.objects.order_by("pk")
+    filter_backends = [DjangoFilterBackend]
+    filterset_fields = ["attendee_user"]
+
+    def get_permissions(self):
+        if self.action == "retrieve":
+            permission_classes = [
+                permissions.IsAdminUser
+                | core_permissions.IsStaffUser
+                | curriculum_permissions.IsWorkshopAttendee
```
sheenarbw, line 583: much better!

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -308,3 +308,32 @@ def verbose_instance_factory(self):
...
+        content = workshop_attendance.content_item
+        content.topic_needs_review = False
+        content.save()
+        return workshop_attendance
+
+    def test_get_instance_permissions(self):
+
+        random_human = UserFactory(is_superuser=False, is_staff=False)
+        self.login(random_human)
+        url = self.get_list_url()
```
sheenarbw: This is a bit strange - your test name says it's about instance permissions, but you are using the `get_list_url` url with filters...

You aren't actually testing the permissions you created.

You should rather get a specific instance instead of a list of instances if you want to test your instance permissions.

Eg:   `url = f"{self.get_list_url()}/{workshop_attendance.id}"` 

If a user wants to fetch their own attendance instance then they should be able to

sheenarbw (changes requested):
Nearly there!

Your code looks good, but your tests aren't testing what they say they are testing

sheenarbw (approved):

PR #40, opened 2020-10-08
created date-time picker widget

Venusian-Mane:

## Description:

## I solemnly swear that:

### DIFF
--- a/frontend/src/components/widgets/DateTimePicker.js
+++ b/frontend/src/components/widgets/DateTimePicker.js
@@ -1,0 +1,34 @@
+import React from "react";
+import { makeStyles } from "@material-ui/core/styles";
+import TextField from "@material-ui/core/TextField";
+
+const useStyles = makeStyles((theme) => ({
+  container: {
+    display: "flex",
+    flexWrap: "wrap",
+  },
+  textField: {
+    marginLeft: theme.spacing(1),
+    marginRight: theme.spacing(1),
+    width: 200,
+  },
+}));
+
+export default function DateAndTimePickers() {
+  const classes = useStyles();
+
+  return (
+    <form className={classes.container} noValidate>
+      <TextField
+        id="datetime-local"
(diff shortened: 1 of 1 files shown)


frontend/src/components/widgets/DateTimePicker.js
```diff
@@ -0,0 +1,34 @@
...
+}));
+
+export default function DateAndTimePickers() {
+  const classes = useStyles();
+
+  return (
+    <form className={classes.container} noValidate>
+      <TextField
+        id="datetime-local"
+        label="Due Date"
```
sheenarbw, line 24: Are you certain that date time pickers will only be used for due dates? Now and forever?

Wouldn't it be better to accept the label as a prop?

And how do we handle events? Like onChange? And what if we want to populate it with a useful default value that isn't in 2017?

And what if we want to include it in a seperate form?

This code is wearing handcuffs. It isn't particularly reusable. The idea with widgets is that they need to work in different contexts. 

As this stands we can just use this to draw useless form elements that dont connect to anything. If this code were to be useful then there are sections that would need to be rewritten. So the code is not complete.

sheenarbw (changes requested):
Ok there are a few big problems here. 

The first problem is that it's always important to think about how your code will be used. Basically you took an existing  form element and made it impossible to attach it to any external form logic without major edits. Which is kindof a strange thing to do. How do you think the calling code would get access to the input value? How does one tie any dynamic behavior to the form? Like submitting the form for example? 

The other problem is that you shouldn't build things that already exist. If anyone asks you to build a thing, you should check if it exists. And if it exists then you need to say "cant we just use this thing". I googled material ui date time pickers and this is the first thing that came up: https://material-ui.com/components/pickers/

So basically what we have here is a less useful version of the ones that already exist...

Maybe just close this PR? 
And always  always aim to write code that can be used without being edited by the very first person who needs to use it.

PR #59, opened 2020-10-14
Feat/import intro to tilde course

### LINKED ISSUE #56: Write a script that can import dev_helpers/data/intro-to-tilde-course.json. Quickly please :)
In order to make developement easy, we want a quick and easy way to set up a little curriculum that demonstrates all the different card movements.

This syllabus already exists on our main instane, but loading up absolutely everything in the world into your local db is kinda a pain in the neck. 

Take a look at this: `curriculum_tracking/management/commands/export_curriculum.py`
This was used to create:
dev_helpers/data/intro-to-tilde-course.json

Make another script called `import curriculum` that takes the json and saves it into the db.

The sooner this is done, the sooner others can use it.

Andy-Nkumane:
Related issues: Write a script that can import dev_helpers/data/intro-to-tilde-course.json. Quickly please :) #56

## Description:

## I solemnly swear that:


backend/curriculum_tracking/models.py
```diff
@@ -45,10 +45,10 @@ def __str__(self):
 
 class ContentItemOrder(models.Model, Mixins):
     post = models.ForeignKey(
-        "ContentItem", on_delete=models.PROTECT, related_name="pre_ordered_content"
+        "ContentItem", on_delete=models.PROTECT, related_name="pre_ordered_content", null=True
     )
     pre = models.ForeignKey(
-        "ContentItem", on_delete=models.PROTECT, related_name="post_ordered_content"
+        "ContentItem", on_delete=models.PROTECT, related_name="post_ordered_content", null=True
```
sheenarbw: I dont think this is wise. If there is an instance of ContentItemOrder that has `pre` or `post` as None then it means that there was a bug somewhere else. 

Data integrity is super important. That's why we have constraints. If there is a piece of code that conflcts with a constraint then it means that the code should be fixed. The database shouldn't be edited in order to allow invalid data.

backend/curriculum_tracking/management/commands/import_curriculum.py
```diff
@@ -0,0 +1,74 @@
...
+    def handle(self, *args, **options):
+        json_file = "dev_helpers/data/intro-to-tilde-course.json"
+        save_curriculum_to_db(json_file)
+
+
+def save_curriculum_to_db(json_file):
+    with open(json_file) as f:
+        data = json.load(f)
+
+    df_dict = {i: pd.DataFrame(data[i]) if i != 'curriculum' else pd.DataFrame(pd.Series(data[i])).transpose() for i in data}
```
sheenarbw: I think that using pandas here makes things more complicated than is needed. `data` is really just a dictionary. The data isn't big or anything.  I think it would be very good for you to be comfortable using foundational python. Python is very powerful even when nt using pandas.

backend/curriculum_tracking/management/commands/import_curriculum.py
```diff
@@ -0,0 +1,74 @@
...
+    # dataframes present in dictionary `df_dict` (4) `content_item_orders`, `content_items`, `curriculum`, `curriculum_content_requirements`
+
+    for i in df_dict:
```
sheenarbw: If you do things in a loop like this then it's hard to make sure that stuff happens in a useful order. 

For example what happens if a ContentItemOrder is created before the ContentItems it should refer to are added to the database? 

Try this instead:

```
create_content_items(data['content_items'])  # write this as a seperate function
create_content_item_orders(.....)
create_curriculum(.....)
create_curriculum_content_requirements(.....)
```

sheenarbw (changes requested):
Good first attempt but there are a few problems. There are simpler ways to do this (remember KISS)  and also it looks a bit buggy

backend/curriculum_tracking/management/commands/import_curriculum.py
```diff
@@ -0,0 +1,78 @@
...
+    with open(json_file) as f:
+        data = json.load(f)
+
+    # keys in dictionary `data` (4) `content_item_orders`, `content_items`, `curriculum`, `curriculum_content_requirements`
+    create_content_items(data['content_items'])
+    create_content_item_orders(data['content_item_orders'])
+    create_curriculum(data['curriculum'])
+    create_curriculum_content_requirements(data['curriculum_content_requirements'])
+
+def find_url_index(link, model):
```
sheenarbw: This function can be better. 

This is the same:

```
def find_url_index(link, model):
    if link == None:
        return 

    for index, value in enumerate(model.objects.all()):
        if value.url == link:
            return model.objects.filter(url=link).first()
```

You don't use `index`, so this is better:

```
def find_url_index(link, model):
    if link == None:
        return 

    for  value in model.objects.all():
        if value.url == link:
            return model.objects.filter(url=link).first()
```

Now let's look at that loop... first you fetch every single content item that exists in the database, then you fetch the one that you just matched

```
    for  value in model.objects.all():
        if value.url == link
            also_the_value =  model.objects.filter(url=link).first()
            assert value == also_the_value, "this should never raise an exception, because the sql filter you just performed was also performed by your loop"
            return also_the_value
```

A simpler version would be:

```
return model.objects.get(url=link) 
```

So basically:

```
def find_url_index(link, model):
    if link:
        return model.objects.get(url=link) 
```

This is the sameas your version. Except your version doesn't raise DoesNotExist errors. These errors would be useful because they make us aware of bugs

Also, you only ever call this with model = ContentItem. I dont think it would make sense to call it with any other model. So a further simplification is:

```
def find_url_index(link):
    if link:
        return ContentItem.objects.get(url=link) 
```

And lastly, good code is clear and easy to read. The function name does not describe what the function does.

how about 
```
def get_content_item_from_url(url):
    if url:   # why introduce a new name if we are already calling it url. Consistency is best
           return ContentItem.objects.get(url=url) 

```

backend/curriculum_tracking/management/commands/import_curriculum.py
```diff
@@ -0,0 +1,78 @@
...
+        return 
+    else:
+        for index, value in enumerate(model.objects.all()):
+            if value.url == link:
+                return model.objects.filter(url=link).first()
+
+def create_content_items(data):
+    for content in data:
+        content_items, created = ContentItem.objects.get_or_create(
+            # available_flavours = ','.join(content['available_flavours']),
```
sheenarbw, line 37: are there no available flavours in the export?

backend/curriculum_tracking/management/commands/import_curriculum.py
```diff
@@ -0,0 +1,78 @@
...
+        name = data['name'],
+        short_name = data['short_name']
+    )
+
+def create_curriculum_content_requirements(data):
+    for requirement in data:
+        curriculum_content_requirements, created = CurriculumContentRequirement.objects.get_or_create(
+            content_item = find_url_index(requirement['content_item'], ContentItem),
+            # flavours = requirement['flavours'],
+            curriculum = Curriculum.objects.filter(short_name='tilde intro student').first(),
```
sheenarbw: The short name might not always be 'tilde intro student'. We might use this script to import other things. Rather get the name from the file

sheenarbw: Also what is more efficient:

```
for requirement in data:
    curriculum = Curriculum.objects.filter(short_name='tilde intro student').first()
    do_something_useful(curriculum)
```

or 

```
curriculum = Curriculum.objects.filter(short_name='tilde intro student').first()
for requirement in data:
    do_something_useful(curriculum)
```
Hitting a database is a very slow thing. Since it's slow it's considered expensive - it takes up time and networking capacity. It's better to  just do one fetch.

backend/curriculum_tracking/management/commands/import_curriculum.py
```diff
@@ -0,0 +1,78 @@
...
+    curriculum, created = Curriculum.objects.get_or_create(
+        name = data['name'],
+        short_name = data['short_name']
+    )
+
+def create_curriculum_content_requirements(data):
+    for requirement in data:
+        curriculum_content_requirements, created = CurriculumContentRequirement.objects.get_or_create(
+            content_item = find_url_index(requirement['content_item'], ContentItem),
+            # flavours = requirement['flavours'],
```
sheenarbw: dont we export flavours?

backend/curriculum_tracking/management/commands/import_curriculum.py
```diff
@@ -0,0 +1,78 @@
...
+)
+from curriculum_tracking.card_generation_helpers import get_ordered_content_items
+import json
+from django.core.management.base import BaseCommand
+import pandas as pd
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        json_file = "dev_helpers/data/intro-to-tilde-course.json"
```
sheenarbw: Can you rather allow the json_file to be passed in as a command line parameter?

backend/curriculum_tracking/management/commands/import_curriculum.py
```diff
@@ -0,0 +1,78 @@
+from core.models import Curriculum
+from curriculum_tracking.models import (
+    CurriculumContentRequirement,
+    ContentItemOrder,
+    ContentItem,
+)
+from curriculum_tracking.card_generation_helpers import get_ordered_content_items
+import json
+from django.core.management.base import BaseCommand
+import pandas as pd
```
sheenarbw: What is pandas for?

sheenarbw (changes requested):
Getting close! There's one really funny function and a few other pieces of polish, but this is much better than the last version

backend/curriculum_tracking/management/commands/import_curriculum.py
Andy-Nkumane: there are no flavours in the provided json file, they are all empty

backend/curriculum_tracking/management/commands/import_curriculum.py
Andy-Nkumane: funny thing, initially I started with ContenItem and changed it later on to model

backend/curriculum_tracking/management/commands/import_curriculum.py
sheenarbw: That's true. But the import script should be the opposite of this script:

curriculum_tracking/management/commands/export_curriculum.py

If flavours are added it should still work. We might want to use this to export and import other courses too. I suppose for now it's ok because we dont *need* the flavours right now. But there is a good chance we will have to revisit this work.

sheenarbw: Can you please make an issue that details all the fields that were skipped over.

I think it should just talk about flavours and tags. 

Once that issue exists let m know then I'll merge this

sheenarbw (changes requested):
We can defer the flavour stuff to another issue. Can you make that issue please. I explained more in line

Andy-Nkumane:
@sheenarbw [issue](https://github.com/Umuzi-org/Tilde/issues/71) created

sheenarbw (approved):

Issue #102, opened 2021-05-17
Make a widget that can be used to list the ReviewTrusts that a person has


sheenarbw:
This widget just needs to be demonstrated as a Story in Storybook. It doesn't need to be incorperated into the main app just yet.

The widget needs to look neat and tidy and be fit for human consuption.

Remember KISS. We just need to see the data. 

## what is review trust?

As users review each other's work we keep track of who does a good job of reviewing. If a person is usually right when they review simple-calc part 1 in js, then we might decide that we trust this person. If a person is trusted on a card then when they add a positive review, the card moves to the complete column. 

## Why display it?

People should aim to be trusted reviewers. If they have trust then it means they are doing a good job.  Trust is something worth earning so we want to show people where they are at. 

Also, the more trust we have in the people using Tilde, the better. If we can trust certain recruits then it means that staff members can do higher value tasks.

## The data 

The widget needs to be able to display data structures that looks something like this:

```
[
{
    content_item_title: "something",
    flavours: ["javascript","react"] 
}, 
{
    content_item_title: "something else",
    flavours: ["python","pytest"] 
}, 
]
```
The data structure might change a bit in the future, the APIs still need to be developed.

## looks

Here are some MaterialUI widgets that could be useful:

- Fixed Header table: https://material-ui.com/components/tables/#fixed-header (there can be one row per trust)
- Chip: https://material-ui.com/components/chips/#chip You can use chips to display the flavours

babalwa-mbolekwa:
I am planning to make a PR by the 7th of June at 5pm.

babalwa-mbolekwa:
I am planning to make a PR by the 14th of June. I couldn't make a PR by the 7th of June because of a blocker on setting up Tilde on my machine with Ubuntu and Docker. I am also planning on upskilling on Material UI.
Plans:

-  Upskill on Material UI
- Tilde repo setup on my machine [Done]
- Setup react storybook
- Create a story for the Review trusts widget.
- Push work to Tilde repo and make a PR by the 14th of June.

sheenarbw:
Goal: draw a picture using storybook.

1. Run the storybook that is already built into Tilde. You dopnt need to reimplement it or anything. Just run npm run storybook and it will launch
2. Make a file at frontend/src/components/regions/UserDetails/ReviewTrustTable/Presentation.js. Make a react component in that file that just renders <div>Helloworld</div>
3. Make a story that draws your new component. Your story should be inside the folder frontend/src/stories
[...]

PR #113, opened 2021-06-09
Review trusts list widget

babalwa-mbolekwa:
Related issues: [https://github.com/Umuzi-org/Tilde/issues/102]

## Description:

## Screenshots/videos
ReviewTrustsTable

## I solemnly swear that:

### DIFF
--- a/frontend/src/components/regions/UserDetails/ReviewTrustTable/Presentation.js
+++ b/frontend/src/components/regions/UserDetails/ReviewTrustTable/Presentation.js
@@ -53,7 +53,9 @@ const ReviewTrustTable = (props) => {
                                 </TableCell>
                             </TableRow>
                         )) :
-                        <TableRow></TableRow> 
+                        <TableRow>
+                            <TableCell colSpan="2">Nothing to display</TableCell>
+                        </TableRow> 
                 }
             </TableBody>
         </Table>
--- a/frontend/src/components/regions/UserDetails/ReviewTrustTable/Presentation.js
+++ b/frontend/src/components/regions/UserDetails/ReviewTrustTable/Presentation.js
@@ -1,13 +1,10 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { makeStyles } from '@material-ui/core/styles';
-import Paper from '@material-ui/core/Paper';
 import Table from '@material-ui/core/Table';
 import TableBody from '@material-ui/core/TableBody';
 import TableCell from '@material-ui/core/TableCell';
-import TableContainer from '@material-ui/core/TableContainer';
 import TableHead from '@material-ui/core/TableHead';
-import TablePagination from '@material-ui/core/TablePagination';
 import TableRow from '@material-ui/core/TableRow';
 import Chip from '@material-ui/core/Chip';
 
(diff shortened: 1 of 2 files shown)


frontend/src/stories/ReviewTrustTable.stories.js
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import ReviewTrustTable from "../components/regions/UserDetails/ReviewTrustTable/Presentation";
+
+export default {
+    title: "Tilde/ReviewTrustTable",
+    component: ReviewTrustTable
+}
+
+export const Primary = () => <ReviewTrustTable title="String Calculator" flavours="JavaScript, Regex"></ReviewTrustTable>
```
sheenarbw: Rather:

```
const trustInstanes = [
{
    content_item_title: "something",
    flavours: ["javascript","react"] 
}, 
{
    content_item_title: "something else",
    flavours: ["python","pytest"] 
}, 
]

export const Primary = () = <ReviewTrustTable trustInstances={trustInstances}/>
```

frontend/src/components/regions/UserDetails/ReviewTrustTable/Presentation.js
```diff
@@ -0,0 +1,60 @@
...
+
+const ReviewTrustTable = (props) => {
+
+    const { title, flavours } = props;
+    const classes = useStyles();
+
+    return (
+        <Table stickyHeader aria-label="sticky table">
+            <TableHead>
+                <TableRow>
```
sheenarbw, line 31: We need one row for each ReviewTrust. If I am trusted on 3 projects then I should have 3 rows. If I am trusted on zero projects, I should see zero rows.

sheenarbw (changes requested):
You are on the right track. But we need a the table to be able to display multiple ReviewTrust instances on multiple rows. One row per instance

Look at the issue. The Issue states that :+1: 

The widget needs to be able to display data structures that looks something like this:

```
[
{
    content_item_title: "something",
    flavours: ["javascript","react"] 
}, 
{
    content_item_title: "something else",
    flavours: ["python","pytest"] 
}, 
]
```

If the issue says that the data has a specific shape, then you need to make sure that your work takes that into account. We need a component that would actually work when we feed it the real data.  

Your component must accept an array of objects. Each object has a content_item_title and an array of flavours.

If I include a data structure in an issue description then I put it there to be used. Please when I give you a data structure, make sure that your story uses that exact structure and the entire structure gets sent to your component.

frontend/src/components/regions/UserDetails/ReviewTrustTable/Presentation.js
```diff
@@ -0,0 +1,71 @@
...
+                            <TableRow>
+                                <TableCell>{trustInstance.content_item_title}</TableCell>
+                                <TableCell>
+                                    {trustInstance.flavours.map((flavourChip) => (
+                                        <Chip className={classes.flavourChip} label={flavourChip} />
+                                    ))}
+                                </TableCell>
+                            </TableRow>
+                        )) :
+                        <TableRow></TableRow>
```
sheenarbw: If there are no instances then it would be good to tell the user something. We dont want things to appear weird or buggy. 
For exmple:

`<TableCell colSpan="2">Nothing to display</TableCell>`

sheenarbw (changes requested):
Very good. There is just one small thing that will make the table more user-friendly

sheenarbw (approved):

Issue #104, opened 2021-05-29
If there is an open PR on a card, then display the age of that PR on the card. Eg 1 day ago, 2 days ago, etc


sheenarbw:
We need to be able to see how long people are watiting for PRs to get dealt with

Requires:

- [x] #114 
- [x] #115

babalwa-mbolekwa:
Goal: Display age of open PR on Agile Card if there is an open PR.
Plan:
1. Create a utils.js file in frontend/src/components/regions/AgileBoard/AgileCard/utils.js
2. Create a function inside utils.js that uses Github API, that will return PR age string if there is an Open PR.
3. Create a component that displays PR age and its story, 
4. Add PR age component to AgileCard component and story.
4. Make a PR by the 25th June at 5pm.

babalwa-mbolekwa:
Updated plan:
1. Update user.json file in frontend/src/stories/fixtures/agileCard.json to use most recent data.
2. Create a Component in /frontend/src/components/widgets that will display age of Open PR
 - Create a function that will return difference between oldestOpenPrUpdatedTime and currentTIme.
3. Create a story for the component above.
4. Make a PR by the 25th of June at 5pm.

PR #120, opened 2021-06-24
Feat/open pr age

babalwa-mbolekwa:
Related issues: [https://github.com/Umuzi-org/Tilde/issues/104]

## Description:

## Screenshots/videos
AgeOfOpenPR

## I solemnly swear that:


sheenarbw (changes requested):
This looks very good, but what if the PR was created 3 hours ago?

This might help: https://www.npmjs.com/package/javascript-time-ago I haven't used this package before but it looks promising.

frontend/src/components/widgets/CardBadges.js
```diff
@@ -38,7 +42,17 @@ export default ({ card }) => {
     codeReviewNyCompetentSinceLastReviewRequest,
     codeReviewRedFlagSinceLastReviewRequest,
     openPrCount,
+    oldestOpenPrUpdatedTime,
   } = card;
+
+  let ageOfOpenPR = () => {
```
sheenarbw: This function can be a bit cleaner in a few ways. Firstly, use consts for constants

Now looking at this line:
```
    const age = timeAgo.format(date - 24 * 60 * 60 * 1000);
```
Why are you subtracting that number? This line deserves a comment. And this function deserves some tests.

When adding comments you can describe why you did something (so that the next programmer doesn't waste time second-guessing your work). You can also link to external documentation. Eg you can say `// based on https://something.useful`.

Unit tests serve multiple purposes: One is to test the code (dah) but also good tests act as documentation. Good tests describe the befavious of the code.  If you write tests for this function you would need to refactor your code so that this function is testable. Perhaps put it inside a utils.js script and let it accept the date as a parameter. eg:

```
getAgeString = ({dateString}) => { // notice the name changes. Why do you think I would suggest this?
      const date = new Date(dateString);
      const timeAgo = new TimeAgo('en-US');
      // useful comment goes here
      return timeAgo.format(date - 24 * 60 * 60 * 1000); 
}
```

Remember that any fool can make code that runs. A good coder writes code that people can read.

sheenarbw (changes requested):
The code looks like it works. But it is dirty and unclear in a few ways.

Try to empathise with the next programmer who will work with this code. Imagine we get a new junior on the team and they need to work with this code. They'll have a hard time and they'll have to go do some research or ask you a bunch of questions.  Let your code tell a story.

sheenarbw (changes requested):
I'm not convinced that your function will work. Please use tests to prove that it works

sheenarbw (changes requested):
These tests will pass today and fail tomorrow.  We need the tests to pass even if we run them a month from now.

sheenarbw (approved):

Issue #131, opened 2021-07-20
Update the CardDetails modal so that it shows a button for setting the due date


sheenarbw:
On the card details modal next to the due date, add a button.

The button should only show up in the following situations:

1. If the currently logged in user is assigned to the card and there is no due date yet (in other words, people can set their own due dates, but can't change them)
2. If the currently logged in user has permission "MANAGE_CARDS" on the card

## Branching stratergy 

Please branch off `feat/set-individual-card-due-dates` instead of develop.  We'll use that branch so that we can make the feature fully work before merging it into develop. We dont want to deploy a half-finished deature. 

## permissions work like this:

User1 can have permissions assigned for multiple teams. Eg they can have view access to team 1 and manage cards access for team 2

If user 2 is in team 1 then user 1 can view their cards. 
If user is in team 2 then user 1 can manage their cards
if user 2 is in both those teams then user 1 has all the permissions

babalwa-mbolekwa:
Goal: Create a button for setting due date on a card.
Plan:
1. Branch off **feat/set-individual-card-due-dates**
2. add due date button component feature in frontend/src/components/regions/CardDetailsModal/Presentation.js
3. Update story for card details modal in frontend/src/stories/CardDetailsModal.stories.js
3. Make a PR by 23 July 2021 5pm.

PR #138, opened 2021-07-23
Due date button

babalwa-mbolekwa:
Related issues: [https://github.com/Umuzi-org/Tilde/issues/131]

## Description:

## Screenshots/videos
dueDateButton

## I solemnly swear that:


frontend/src/components/regions/CardDetailsModal/Presentation.js
```diff
@@ -62,66 +64,71 @@ const CardBasicDetails = ({ card }) => {
...
-                {card.completeTime && (
-                  <TableRow>
-                    <TableCell>Complete Time </TableCell>
-                    <TableCell>{completeTime}</TableCell>
-                  </TableRow>
-                )}
-              </TableBody>
-            </Table>
-          </Paper>
-        </Grid>
```
sheenarbw, line 117: Why are you removing all the grid stuff? Surely we are just trying to add a button. I dont think you should be rewriting the whole file

frontend/src/components/regions/CardDetailsModal/Presentation.js
```diff
@@ -7,8 +7,10 @@ import {
   TableRow,
   TableCell,
   Typography,
+  Button,
 } from "@material-ui/core";
 import { makeStyles } from "@material-ui/core/styles";
+// import Button from '@material-ui/core/Button';
```
sheenarbw, line 13: please clean up unnecessary comment

frontend/src/components/regions/CardDetailsModal/Presentation.js
```diff
@@ -62,66 +64,71 @@ const CardBasicDetails = ({ card }) => {
...
+
+      <Paper className={classes.sectionPaper} variant="outlined">
+        <Table>
+          <TableBody>
+            <TableRow>
+              <TableCell>Due Time</TableCell>
+              <TableCell>
+                {dueTime}
+              </TableCell>
+              {
```
sheenarbw, line 92: generally we try to keep presentation files as dum as possible. 

Can you put this functionality in another file?

eg make a file called 
src/components/regions/CardDetailsModal/utils.js 

then make a function called `canSetDueTime({card,authUser})` and just call it as needed

frontend/src/components/regions/CardDetailsModal/Presentation.js
```diff
@@ -62,66 +64,71 @@ const CardBasicDetails = ({ card }) => {
...
+                authUser.email === card.assigneeNames[0] && !dueTime && 
+                Object.keys(authUser.permissions.teams)
+                .map((key) => authUser.permissions.teams[key].permissions[0])
+                .includes("MANAGE_CARDS") ? (
+                  <TableCell>
```
sheenarbw, line 97: This row of the table sometimes has 3 cells and sometimes has 2.

All other rows only have 2 cells. Why do we want the button to be so far to the right of the screen? Why make the table inconsistent?

If you just put the button in the same cell as the due time text then what would go wrong?

eg:

```
<TableCell>
                {dueTime} {canSetDueTime({card,authUser}) && (<Button variant="outlined">Set Date</Button>)}
</TableCell>
```

frontend/src/components/regions/CardDetailsModal/Presentation.js
```diff
@@ -53,7 +55,7 @@ const TopicProgressDetails = ({ topicProgress, reviews }) => {
   );
 };
 
-const CardBasicDetails = ({ card }) => {
+const CardBasicDetails = ({ card, authUser }) => {
```
sheenarbw, line 58: You've added an extra argument to this function.  It would be good to make sure that existing code doesn't break as a result.

Where is CardBasicDetails used in the code? Does it recieve the correct props? 

Your story will look of, but the code wont actually work in context.

sheenarbw (changes requested):
There are a few problems in the code. 

But  also, it's really importnat to read the issue carefully. There is a section of the issue that talks about branching stratergy. It states:

```
Please branch off feat/set-individual-card-due-dates instead of develop. We'll use that branch so that we can make the feature fully work before merging it into develop. We dont want to deploy a half-finished deature.
```

Does that make sense?

So you should be branching off the correct branch and merging into the correct branch too.

PR #140, opened 2021-07-28
due time button added

babalwa-mbolekwa:
Related issues: [https://github.com/Umuzi-org/Tilde/issues/131]

## Description:

## Screenshots/videos
dueTimeButton

## I solemnly swear that:

### DIFF
--- a/frontend/src/components/regions/CardDetailsModal/Presentation.js
+++ b/frontend/src/components/regions/CardDetailsModal/Presentation.js
@@ -1,5 +1,6 @@
 import React from "react";
 import {
+  Button,
   Grid,
   Paper,
   Table,
@@ -19,6 +20,7 @@ import Modal from "../../widgets/Modal";
 import ProjectDetails from "./ProjectDetails";
 import UsersTable from "./UsersTable";
 import Reviews from "./Reviews";
+import { canSetDueTime } from "./utils";
 
 const useStyles = makeStyles((theme) => ({
   paper: {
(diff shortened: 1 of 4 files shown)


frontend/src/components/regions/CardDetailsModal/Presentation.js
```diff
@@ -122,6 +126,7 @@ const CardBasicDetails = ({ card }) => {
 
 export default ({
   card,
+  authUser,
```
sheenarbw, line 129: Now the CardDetailsModal requires that it has access to the authUser.

Does it have access to the authUser? I dont think anything passes authUser to this component.

If you have a function like this:

```
function importantStuff({something}){
   ....stuff
}

```

And you change it to:

```
function importantStuff({something,and,more,stuff, authUser}){
   ....stuff
}
```

But it gets called like this in other areas of the code base:

```
importantStuff({something})
```

Will it work? 

No. It needs to be called with the correct arguments.



If a Presentation component needs an argument, then that argument needs to be passed in. Otherwise it wont work. The component will break as soon as we try to actually use it in real life. 

## what to do:

For each component you'll see that there are at least 2 files:

Presentation.js <- this has the looks
index.js <- this has the brains. This renders the presentation component with the correct props.

If you change the presentation component so it needs extra props, then make sure the index.js component actually sends the right props.

sheenarbw (changes requested):
This code will break. If a function requires new arguments then we need to pass those arguments in.

sheenarbw (approved):

PR #141, opened 2021-07-30
Due time button

### LINKED ISSUE #132: Add a click event handler to the button in #131 so that when the user clicks the button, the correct form comes up as a modal
## Branching stratergy 

Please branch off `feat/set-individual-card-due-dates` instead of develop.  We'll use that branch so that we can make the feature fully work before merging it into develop. We dont want to deploy a half-finished deature. 

## Opening the modal using redux

Please see how we open the card details modal and mimick that.

look in here: src/components/regions/AgileBoard/AgileCard/index.js

1. We are using redux 
2. It makes use of an operation defined in src/components/regions/CardDetailsModal/redux/operations.js . Take note of the file structure. You wil need to architect your redux filoes in the same way 
3. The store is defined in src/redux/store.js, add your reducer to the combined reducers
4. Add the modal underneath the other modals here: src/components/regions/AppHeaderAndMenu/Presentation.js

babalwa-mbolekwa:
Related issues: [https://github.com/Umuzi-org/Tilde/issues/132]

## Description:

## I solemnly swear that:


frontend/src/components/regions/DueTimeFormModal/Presentation.js
```diff
@@ -0,0 +1,41 @@
...
+const DueTimeForm = (props) => {
+
+  const { dueTime } = props;
+  const defaultValue = dueTime ? dueTime.split('').slice(0, 16).join('') : "";
+  const classes = useStyles();
+
+  return (
+      <form novalidate>
+        <TextField
+          id="datetime-local"
```
sheenarbw: This is in the develop branch
Copy pasting between branches is a no-go. We use git to get info from one branch to another.

sheenarbw (commented):
You are on the right track. It looks like you just need to dispatch the action to finish this off.

There is also a thing that needs to be fixed. 

I see you copied some code from the develop branch into your branch. That's not ideal. The way we get code to be represented in multiple git branches is by using `git rebase` or `git merge`. Not copy paste.

In this case, the simplest thing would be to merge develop into protected/feat/set-individual-card-due-dates. I'll do that now and push.

Then you can rebase your branch
```
git checkout due-time-button
git rebase protected/feat/set-individual-card-due-dates
```

Copy paste is hardly ever a good idea. Copy paste generally means we need to clean something up.

sheenarbw (changes requested):
Your code isn't DRY.

You should probably delete frontend/src/components/regions/AgileCardDueDateForm/Presentation.js. We are just using the Modal, right? What is this extra file for?

Always make it DRY.

frontend/src/components/regions/CardDetailsModal/index.js
```diff
@@ -80,6 +82,10 @@ function CardDetailsModalUnconnected({
     openReviewFormModal({ cardId: project.agileCard });
   };
 
+  const handleClickSetDueTime = () => {
+    openDueTimeFormModal({ cardId: project.agileCard });
```
sheenarbw: Can we not set due times for topics? Why do we need to make use of the project here?

why not just:
```
    openDueTimeFormModal({ cardId });
```

cardId is a prop. A card can be associated with a project or a topic

frontend/src/components/regions/DueTimeFormModal/Presentation.js
```diff
@@ -0,0 +1,48 @@
...
+
+    dueTimeButton: {
+        marginTop: "0.3rem",
+        marginRight: "0.3rem",
+    },
+});
+
+const DueTimeFormModal = ({ card, closeModal }) => {
+
+  const {dueTime} = card || {};
```
sheenarbw: under what circumstances would you want the due time to be an object? Shouldn't it always be a string here? 
Does your story work if there is no card passed into the presentation component?

frontend/src/components/regions/DueTimeFormModal/Presentation.js
```diff
@@ -0,0 +1,48 @@
...
+  const defaultValue = dueTime ? dueTime.split('').slice(0, 16).join('') : "";
```
sheenarbw: Can you add a comment here.  It is not clear why you have chosen slice(0,16) instead of slice(0,2345).  The next person to deal with this code will need to waste time looking things up.

Just say something like:
// dueTime is either an empty string or something like this: "put in some example time so that folks can see the format here"

frontend/src/components/regions/DueTimeFormModal/Presentation.js
```diff
@@ -0,0 +1,48 @@
...
+  const classes = useStyles();
+
+  return (
+    <Modal open={!!card} onClose={closeModal}>
+      <Paper>
+        <form novalidate>
+          <TextField
+            id="datetime-local"
+            label="Due Date:"
```
sheenarbw: What's the convention with form labels? Do we currently have colons at the ends of label strings?

frontend/src/components/regions/DueTimeFormModal/index.js
```diff
@@ -0,0 +1,37 @@
+import React from "react";
+import Presentation from "./Presentation";
+import { connect } from "react-redux";
+import operations from "./redux/operations";
+
+export const DueTimeFormModalUnconnected = ({ cardId, card, closeModal }) => {
```
sheenarbw: what is cardId used for here? Why do we need it as a prop?

sheenarbw (changes requested):
Still needs a bit of polish

sheenarbw (changes requested):
..

sheenarbw (changes requested):
canSetDueTime is buggy. IIf I am logged in as me, I should be able to set due times on anybody's cards. But as it stands, I can only set due times on my own cards.

Please write some tests on the function, and then make those tests pass.

We need:
1. If the card belongs to the current user and there is no due time set, then they should be able to see the button
2. If the card belongs to the current user and that user has no special permissions, and there is a due time set, then dont show the button
3. If the current user is looking at a card and is not the assignee and they have managment permissions for that card, then show the button,
4. if the current user is not the assignee and has no managment permissions then dont show the due time button, even if there is no due time set

Also, looking at the model itself, everthing is very squished. Please add some space around things so it is more tidy. We are on  a mission to make Tilde look better, let's not make it look worse. It's better to take care the first time

Sbonelo01:
@Babalwa01 , Good work so far, did you manage to get to the tests?

babalwa-mbolekwa:
> @Babalwa01 , Good work so far, did you manage to get to the tests?

Thank you Sbonelo. I will be addressing the changes requested today. I fixed the blocker for the env variables.

frontend/src/components/regions/DueTimeFormModal/Presentation.js
```diff
@@ -0,0 +1,53 @@
...
+    },
+
+    paper: {
+      padding: "1rem"
+    }
+});
+
+const DueTimeFormModal = ({ card, closeModal }) => {
+
+  const {dueTime} = "";
```
sheenarbw: What is this line for?

If the card has a due time that was already set (eg the card is due on Tuesday next week) then we want to pre-populate the form so that it refers to Tuesday next week. We dont want it to always be vblank.

Also, I dont think this code does what you think it does:

```
>   const {dueTime} = "";
> console.log(dueTime)
undefined
```

Your comment is still not entirely clear. Why are you choosing to slice at 16? Why not 133? 
Good code makes its intentions clear. Please make sure that the intention of this line is clear.

Eg: # if the dueTime has a value, then we need to strip out the seconds.

sheenarbw (changes requested):
There is a bug. defaultValue will always be "". This is not the correct behavior.

Also the comment can be more clear.

babalwa-mbolekwa:
@sheenarbw @Sbonelo01  I have attended to all the changes accept for the code in  _**src/components/regions/DueTimeFormModal/Presentation.js**_  I tried to use `const dueTime = card.dueTime;` instead of `const dueTime = "";` but it gave me this error: `TypeError: Cannot read properties of null (reading 'dueTime')` not quite sure how I can resolve it

frontend/src/components/regions/CardDetailsModal/CardDetailsModal.test.js
```diff
@@ -0,0 +1,90 @@
...
+        dueTime: null,
+    },
+    currentUser: {
+        email: "babalwa.mbolekwa@umuzi.org",
+    }
+}
+
+test("canSetDueTime function returns true if card belongs to current user and there's no due time set", () => {
+    expect(canSetDueTime(cardAndUserObject1.agileCard, cardAndUserObject1.currentUser)).toBe(true);
+});
```
sheenarbw, line 18: The first test would be better structured like so:

```
test("canSetDueTime function returns true if card belongs to current user and there's no due time set", () => {
    // all the data that is used in the test is IN the test

    const user = {id: 1}
    const card = {
        assignees: [user.id],  // there is no space for typos here. It is very explicit. Ids are best when comparing things
        dueTime: null,
    }
    // also note how simple the data is. Flat is better than nested.

    expect(canSetDueTime({card, user})).toBe(true);  
    // Notice the {} in the parameters. Generally it's better to use destructuring instead of having positional arguments. It is much less likely to cause bugs later
});
```

sheenarbw (changes requested):
Ok, getting there. Your tests need some polish.

frontend/src/components/regions/CardDetailsModal/CardDetailsModal.test.js
```diff
@@ -0,0 +1,74 @@
+import { canSetDueTime } from "../../widgets/utils";
+import card from "../../../stories/fixtures/agileCard.json";
+import authUser from "../../../stories/fixtures/authUser.json";
+
+/* Instance 1: Card belongs to current user and there is no due time set */
+test("canSetDueTime function returns true if card belongs to current user and there's no due time set", () => {
+    const user = {id: 1}
+    const card = {
+        assigneeNames: [user.id],
```
sheenarbw: assigneeNames are not ids.

frontend/src/components/regions/CardDetailsModal/CardDetailsModal.test.js
```diff
@@ -0,0 +1,74 @@
+import { canSetDueTime } from "../../widgets/utils";
+import card from "../../../stories/fixtures/agileCard.json";
+import authUser from "../../../stories/fixtures/authUser.json";
+
+/* Instance 1: Card belongs to current user and there is no due time set */
```
sheenarbw: what is this comment for? Surely the next line is descriptive enough?

frontend/src/components/regions/CardDetailsModal/CardDetailsModal.test.js
```diff
@@ -0,0 +1,74 @@
...
+                    active: true,
+                    permissions: [
+                        "REVIEW_CARDS"
+                    ]
+                }
+            }
+        }
+    } 
+    const card = {
+        assigneeNames: [user.id],
```
sheenarbw: this should say `assignees : [user.id]`

frontend/src/components/regions/CardDetailsModal/CardDetailsModal.test.js
sheenarbw: this should say `assignees : [user.id]`

frontend/src/components/regions/CardDetailsModal/CardDetailsModal.test.js
```diff
@@ -0,0 +1,74 @@
...
+                    active: true,
+                    permissions: [
+                        "MANAGE_CARDS"
+                    ]
+                }
+            }
+        }
+    } 
+    const card = {
+        assigneeNames: [user.id],
```
sheenarbw: this should say `assignees : [user.id]`

frontend/src/components/regions/CardDetailsModal/CardDetailsModal.test.js
```diff
@@ -0,0 +1,74 @@
...
+        dueTime: "2021-06-14T09:58:28Z",
+    }
+    expect(canSetDueTime({card, user})).toBe(false);
+});
+
+/* Instance 3: Current user is not the assigned user but has management permissions and due time is not set */
```
sheenarbw: comment probably unnecessary

frontend/src/components/regions/CardDetailsModal/CardDetailsModal.test.js
```diff
@@ -0,0 +1,74 @@
...
+        assigneeNames: [user.id],
+        dueTime: null,
+    }
+    expect(canSetDueTime({card, user})).toBe(true);
+});
+
+/* Instance 4: Current user is not the assigned user, due time is not set and
+   current user doesn't have the 'MANAGE_CARDS' permission */
+test("canSetDueTime function returns false if card doesn't belong to current user, current user doesn't have management permissions and there's no due time set", () => {
+    expect(canSetDueTime({card, authUser})).toBe(false);
```
sheenarbw: should say:
```
    expect(canSetDueTime({card, user:authUser})).toBe(false);
```

sheenarbw: Or all the other calls are incorrect

Sbonelo01 (dismissed):
Looks good to me

frontend/src/components/regions/AgileBoard/AgileCard/Presentation.js
```diff
@@ -152,7 +152,7 @@ export default ({
         <Typography variant="h6" component="h2">
           {card.title}
         </Typography>
-
+       
         {/* {card.flavourNames.map((flavour) => (
```
Vuyisanani, line 156: please remove this

frontend/src/components/regions/AppHeaderAndMenu/Presentation.js
```diff
@@ -157,6 +158,7 @@ export default function Presentation(props) {
       <CssBaseline />
       <CardDetailsModal />
       <AddCardReviewModal />
+      <DueTimeFormModal />
       {/* <MarkSingleCardAttendanceModal /> */}
```
Vuyisanani, line 162: please remove this

frontend/src/components/regions/CardDetailsModal/CardDetailsModal.test.js
```diff
@@ -0,0 +1,89 @@
...
+                }
+            }
+        }
+    }
+    const card = {
+        assignees: [authUser.userId],
+        dueTime: null,
+    }
+    const teams = authUser.permissions.teams;
+    expect(canSetDueTime({card, authUser, teams})).toBe(true);
```
sheenarbw: Here you are effectively passing in teams twice. You are passing it in on the authUser object and as a seperate object. It is not useful to pass the same piece of information into the same function twice. What is this for?

sheenarbw (changes requested):
Looks like you are misunderstanding the issue. 

The card belongs to a user. That user is in some teams. 
Does the authUser have access to those teams?

Vuyisanani (commented):
Hi @Babalwa01 

Just before I can approve, should the button  only be in the page card below :
Screenshot 2021-10-22 at 02 40 17

or should it be in any card that has a Due Time button, e.g below:
Screenshot 2021-10-22 at 02 39 54

Also please include your UI screenshot

frontend/src/components/regions/AppHeaderAndMenu/Presentation.js
babalwa-mbolekwa: > please remove this

hey @Vuyisanani  . The `DueTimeFormModal` doesn't show when I click on the set time button, when I remove it

babalwa-mbolekwa:
> Hi @Babalwa01
> 
> Just before I can approve, should the button only be in the page card below : Screenshot 2021-10-22 at 02 40 17
> 
> or should it be in any card that has a Due Time button, e.g below: Screenshot 2021-10-22 at 02 39 54

@Vuyisanani I'm not quite sure. I will enquire about it on Monday. but this part looks like it only displays the dates

babalwa-mbolekwa:
cardWithDueTime
cardWithoutDueTime
dueTimeFormModal

Vuyisanani (approved):

sheenarbw (approved):

PR #161, opened 2021-08-31
Add an extra field to the serializer to expose the latest reviewer since last request for a review/Ticket #143

### LINKED ISSUE #143: upgrade agile card api endpoint so that it exposes an extra field: a list of the ids of the users who have reviewed the card since the last review request
You will need to know a little bit about Django Rest Framework (DRF)

Look for the AgileCardSerialiser, add the extra field there
Test that your field returns the right ids

EckardBerry:

## Description:
1.  I made some changes to the models.py file for the backend - Pls see lines 1389 till the end
2.  I made some changes to the serializer.py file for the backend - Pls see AgileCardSerializer
3.  I added a test to the tests directory for the backend, test name is:  test_find_ids_of_latest_reviewers_for_agile_card_api.py

## I solemnly swear that:


backend/curriculum_tracking/api_views.py
```diff
@@ -136,22 +138,14 @@ def _get_teams_from_workshop_attendance(self, request, view):
...
-            | curriculum_permissions.IsCardReviewer
-            | core_permissions.HasObjectPermission(
-                permissions=Team.PERMISSION_VIEW, get_objects=_get_teams_from_card
-            )
-        )
-        | core_permissions.IsReadOnly
-        | core_permissions.IsStaffUser
-        | core_permissions.IsCurrentUserInSpecificFilter("assignees")
-        | core_permissions.HasObjectPermission(
+        Union[permissions.IsAdminUser, core_permissions.ActionIs(
```
sheenarbw: This wasn't necessary.  The permissions work as they are.

What are you actually trying to do here?

backend/curriculum_tracking/api_views.py
```diff
@@ -589,13 +584,11 @@ def get_permissions(self):
...
             permission_classes = [
-                permissions.IsAdminUser
-                | curriculum_permissions.IsProjectAssignee
-                | curriculum_permissions.IsProjectReviewer
-                | core_permissions.HasObjectPermission(
-                    permissions=Team.PERMISSION_VIEW,
-                    get_objects=_get_teams_from_recruit_project,
-                )
+                Union[
+                    permissions.IsAdminUser, curriculum_permissions.IsProjectAssignee, curriculum_permissions.IsProjectReviewer, core_permissions.HasObjectPermission(
```
sheenarbw: This wasn't necessary.  The permissions work as they are.

What are you actually trying to do here?

backend/curriculum_tracking/serializers.py
```diff
@@ -177,8 +177,14 @@ class Meta:
             "can_force_start",
             "open_pr_count",
             "oldest_open_pr_updated_time",
+            "users_of_most_recent_reviewers",
         ]
 
+    users_of_most_recent_reviewers = serializers.SerializerMethodField("get_reviewed_users_most_recent")
```
sheenarbw: This name doesn't make a lot of sense.  a reviewer is a user. So you are saying the users of the recent users that reviewed stuff.

How about: `users_that_reviewed_since_last_review_request`?  Does this name make sense? 

Ideally just by looking at a name we hould know it's intention, it's a hard thing to get right sometimes.

backend/curriculum_tracking/serializers.py
```diff
@@ -177,8 +177,14 @@ class Meta:
...
+
+    def get_reviewed_users_most_recent(self, instance):
```
sheenarbw: naming also should be consistent as much as possible.

If the field is called `x` then a good name for the serialiser method would be `get_x`

backend/curriculum_tracking/tests/test_find_ids_of_latest_reviewers_for_agile_card_api.py
```diff
@@ -0,0 +1,75 @@
...
+        )
+
+        self.project = RecruitProjectFactory(content_item=self.card.content_item)
+        self.user = UserFactory()
+        self.card.assignees.set([self.user])
+        self.assertIsNotNone(self.card.recruit_project)
+        self.assertEqual(self.card.status, AgileCard.IN_PROGRESS)
+        self.assertEqual(self.project.content_item, self.card.content_item)
+        self._reviewer_ids_since_latest_review_request = []
+        self.project.review_requests_made = []
```
sheenarbw: self.project.review_requests_made = []

this is strange. Is the project not being construsted properly so you need to construct it some more? 

I dont think projects need to have this attribute

backend/curriculum_tracking/tests/test_find_ids_of_latest_reviewers_for_agile_card_api.py
```diff
@@ -0,0 +1,75 @@
...
+            self._reviewer_ids_since_latest_review_request.append(review_1.reviewer_user)
+            self._reviewer_ids_since_latest_review_request.append(review_2.reviewer_user)
+            self._reviewer_ids_since_latest_review_request.append(review_3.reviewer_user)
+
+    @property
+    def showing_latest_reviewer_ids(self):
+        return self._reviewer_ids_since_latest_review_request[::1]
+
+    def test_to_find_all_recent_reviewer_users_since_latest_review_request(self):
+        self.finding_latest_reviewer_ids()
```
sheenarbw: why not put the test logic inside the test? Why put it elsewhere?

backend/curriculum_tracking/tests/test_find_ids_of_latest_reviewers_for_agile_card_api.py
```diff
@@ -0,0 +1,75 @@
...
+
+        review_2 = factories.RecruitProjectReviewFactory(
+            status=COMPETENT, recruit_project=self.project
+        )
+
+        review_3 = factories.RecruitProjectReviewFactory(
+            status=EXCELLENT, recruit_project=self.project
+        )
+
+        if len(self.project.review_requests_made) == 0:
```
sheenarbw: This is a test. Why would you have an if/else? You should already know what branch of this will get executed 100% of the time.  the other brancvh should be removed from the code.

backend/curriculum_tracking/tests/test_find_ids_of_latest_reviewers_for_agile_card_api.py
```diff
@@ -0,0 +1,75 @@
...
+            return 'No review requests made and therefore no reviews submitted'
+        else:
+            self._reviewer_ids_since_latest_review_request.append(review_1.reviewer_user)
+            self._reviewer_ids_since_latest_review_request.append(review_2.reviewer_user)
+            self._reviewer_ids_since_latest_review_request.append(review_3.reviewer_user)
```
sheenarbw: Rather set things up so that the project has a bunch of reviews and a review request timestamp. Then calculate in your head which users should be returned, then assert that those are the ones that got returned.

Your tests shouldn't be calculating things. You do the calculations and then make assertions.

pseudocode:
```
card = setup_the_card(review_request_time=t3)
review1 = setup_review_at(time = t1) # before t3
review2 = setup_review_at(time = t2) # before t3
review3 = setup_review_at(time = t4) # after t3 
review4 = setup_review_at(time = t5) # after t3
users = get_the_users_by_calling_the_function_under_test()
assert  [review3.user,review4.user] == users
```

backend/curriculum_tracking/tests/test_find_ids_of_latest_reviewers_for_agile_card_api.py
```diff
@@ -0,0 +1,75 @@
...
+                content_type=ContentItem.PROJECT,
+                project_submission_type=ContentItem.LINK,
+            ),
+            status=AgileCard.IN_PROGRESS,
+        )
+
+        self.project = RecruitProjectFactory(content_item=self.card.content_item)
+        self.user = UserFactory()
+        self.card.assignees.set([self.user])
+        self.assertIsNotNone(self.card.recruit_project)
```
sheenarbw: The card and the recruit_project are not associated with each other, even though they have the same content item.

Eg if Tshepo and Mpho both have cards for simple calc part 1, they each have a card, each card is associated with a different RecruitProject. Both the projects are associated with the same content item

This would be cleaner:

```
self.card = factories.AgileCardFactory(
            content_item=factories.ContentItemFactory(
                content_type=ContentItem.PROJECT,
                project_submission_type=ContentItem.LINK,
            ),
            status=AgileCard.IN_PROGRESS,
        )

self.project = self.card.recruit_project # this should already exist on the card  
self.user = self.card.assignees.first()    # same
```

backend/curriculum_tracking/tests/test_find_ids_of_latest_reviewers_for_agile_card_api.py
```diff
@@ -0,0 +1,75 @@
...
+        if len(self.project.review_requests_made) == 0:
+            return 'No review requests made and therefore no reviews submitted'
+        else:
+            self._reviewer_ids_since_latest_review_request.append(review_1.reviewer_user)
+            self._reviewer_ids_since_latest_review_request.append(review_2.reviewer_user)
+            self._reviewer_ids_since_latest_review_request.append(review_3.reviewer_user)
+
+    @property
+    def showing_latest_reviewer_ids(self):
+        return self._reviewer_ids_since_latest_review_request[::1]
```
sheenarbw: KISS.

Why do we need a property that just calls a different function. Why not just call the function where you need to call it? 

Why the [::1]. Are you trying to make the result into a list? Why npot just say `list(...)`, make it explicit?

backend/curriculum_tracking/models.py
```diff
@@ -1383,3 +1385,31 @@ def add_collaborator(self, user, add_as_project_reviewer):
                 if self.recruit_project:
                     self.recruit_project.reviewer_users.add(user)
         self.save()
+
+    def finding_latest_reviewer_users(self):
```
sheenarbw: If I ask you to tell me what this function does in English then the answer would be something like: 

"it gets the users that reviewed since the last review request" so a good name would be something like `get_users_that_reviewed_since_last_review_request`. It's more explicit and it's plain English.

backend/curriculum_tracking/models.py
```diff
@@ -1383,3 +1385,31 @@ def add_collaborator(self, user, add_as_project_reviewer):
...
+
+        reviewer_user_since_latest_review_request = []
+        review_requests_made = []
+
+        # Making sure to only get reviewers since the latest request for a review
+        if self.status in [
+            AgileCard.REVIEW_FEEDBACK,
+            AgileCard.IN_REVIEW,
+        ]:
+            review_requests_made.append(self.topic_progress.review_request_time)
```
sheenarbw: why topic_progress? What about project_progress?

sheenarbw: There will only ever be one review request time. Why should it be in an array?

backend/curriculum_tracking/models.py
```diff
@@ -1383,3 +1385,31 @@ def add_collaborator(self, user, add_as_project_reviewer):
...
+        else:
+            review_requests_made = []
+
+        reviews_since_latest_review_request = RecruitProjectReview.objects.filter(
+            recruit_project=AgileCard.recruit_project
+        ).filter(timestamp__lt=self.topic_progress.review_request_time)
```
sheenarbw: we need reviews that happened AFTER the last review request time. This fetches the reviews that happened before.

sheenarbw (changes requested):
Ok so this needs a lot of work. It's a good effort but much to learn. 

Can you look over it. dont treat fixing this as urgent, I think we should have a pairing session soon. This week is bananas for me but I'll try make a plan to spend some time with you.

backend/curriculum_tracking/models.py
```diff
@@ -1383,3 +1385,20 @@ def add_collaborator(self, user, add_as_project_reviewer):
...
+            ReviewClass = RecruitProjectReview
+
+        elif self.content_item.content_type == ContentItem.TOPIC:
+            ReviewClass = TopicReview
+
+        reviews = ReviewClass.objects.filter(
+            timestamp__gte=self.review_request_time
+        )
+
+        return list(map(lambda review: review.reviewer_user_id, reviews))
```
sheenarbw: There is a much simpler way to do the same thing. Remeber that the best code is usually the simplest code.
eg:
```
return [o.reviewer_user_id for o in reviews]
```

Map returns an iterable instead of calculating all the values off the bat. Sometimes this is really useful if you are trying to be memory efficient. Or if the cvallback function takes long and you want to spread te workload out over a longer time perios. But in this case you are immediately taking that iterable and turining it into alist.

A list comprehention is simpler, more time and memory efficient (AFAIK) and easier to read. 

With your way we basically:
- create an object and a function
- iterate over the reviews and call the function on each review, function calls have overhead
- inside the function we use a getter to get the user id 

With a list comprehension we
- Dont declare any extra functinos or objects
- iterate over the reviews 
- get the user id with a perfectly normal getter 
- dont introduce function call overhead

Always remember the principle of KISS

backend/curriculum_tracking/serializers.py
```diff
@@ -177,8 +177,14 @@ class Meta:
             "can_force_start",
             "open_pr_count",
             "oldest_open_pr_updated_time",
+            "users_that_reviewed_since_last_review_request",
         ]
 
+    users_that_reviewed_since_last_review_request = serializers.SerializerMethodField("get_users_that_reviewed_since_last_review_request")
+
+    def get_users_that_reviewed_since_last_review_request(self, instance):
+        return models.AgileCard.get_users_that_reviewed_since_last_review_request(instance)
```
sheenarbw: why not:

`return instance.get_users_that_reviewed_since_last_review_request()`

instance should be an AgileCard instance

backend/curriculum_tracking/tests/test_find_ids_of_latest_reviewers_for_agile_card_api.py
```diff
@@ -0,0 +1,74 @@
...
+    #card = factories.AgileCardFactory()
+
+    def setUp(self):
+        self.card = factories.AgileCardFactory(
+            status=AgileCard.IN_PROGRESS,
+        )
+
+        self.project = self.card.recruit_project
+        self.user = self.card.assignees.first()
+        self.card.assignees.set([self.user])
```
sheenarbw:
```
self.user = self.card.assignees.first()
self.card.assignees.set([self.user])   #this line doesn't do much. The assignee is already there
```

If you want to you can check that `self.card.assignees.count() == 1`

backend/curriculum_tracking/tests/test_find_ids_of_latest_reviewers_for_agile_card_api.py
```diff
@@ -0,0 +1,74 @@
...
+            status=NOT_YET_COMPETENT,
+            recruit_project=self.project,
+        )
+        self.review_4.timestamp = time_four
+        self.review_4.save()
+
+        result = self.card.get_users_that_reviewed_since_last_review_request()
+
+        # Only three of the four reviews should have been added as part of the result
+        self.assertNotEqual(len(result), 4)
```
sheenarbw: If the result was equal to `"Mary had a little lamb, its fleece was white as snow"` then this test would pass. 

What should the result be equal to? What specific user ids should be returned?

should result include `self.review_1.reviewer_user_id`? how about  `self.review_2.reviewer_user_id`?

The orderin of the result doesn't matter. The pattern I generally use here is:

```
expeted_result = [correct user ids]
self.assertEqual(sorted(result),sorted(expected_result))  # sorting thins because he order doesn't matter
```

Do you know the zen of Python?  If you don't, open a shell and `import this`
Explicit if better than implicit.

sheenarbw (changes requested):
You are very much on the right track.

Your test needs to be more explicit. And a few things can be simplified. But we are nearly there.

Getting this finished will unblock some frontend wins. Nearly there!

backend/curriculum_tracking/models.py
EckardBerry: I originally had it as return [reviewer.reviewer_user_id for reviewer in reviewers] but I kept
   on getting an error that said that Django model object is not iterable, but
   now since I've destroyed my local database and changed it I no longer get that
   error, don't know why?  But I agree with you that the list/map business is
   not the best outcome.

backend/curriculum_tracking/models.py
```diff
@@ -1383,3 +1384,20 @@ def add_collaborator(self, user, add_as_project_reviewer):
...
+            return []
+
+        if self.content_item.content_type == ContentItem.PROJECT:
+            ReviewClass = RecruitProjectReview
+
+        elif self.content_item.content_type == ContentItem.TOPIC:
+            ReviewClass = TopicReview
+
+        reviews = ReviewClass.objects.filter(
+            timestamp__gte=self.review_request_time
```
sheenarbw, line 1403: Oops, forgot something here. 

We need to make sure that we are filtering by the TopicProgress or RecruitProject entity as well. Not just timestamp.

Eg: If I have 2 projects sucht that project 1 has a bunch of reviews and project 2 des not, then `project2.get_users_that_reviewed_since_last_review_request()` shouldn't return anything user ids.

So basically there is a filter missing. 

The streucture of the code can be the same, except you can do this sort of thing:

```
        if self.content_item.content_type == ContentItem.PROJECT:
            # ReviewClass = RecruitProjectReview not this 
           reviews = RecruitProjectReview.objects.filter(recruit_project = self)

      ....
   # then we chain filters instead of defining the query from scratch
    reviews = reviews.filter(timestamp__gte=self.review_request_time)

```

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +688,66 @@ def test_cancel_workshop_attendance(self):
...
+        )
+
+        self.project = self.card.recruit_project
+        self.user = self.card.assignees.first()
+        self.assertIsNotNone(self.card.assignees)
+        self.assertIsNotNone(self.card.recruit_project)
+        self.assertEqual(self.card.status, AgileCard.IN_PROGRESS)
+        self.assertEqual(self.project.content_item, self.card.content_item)
+
+    def test_finding_latest_reviewer_ids(self):
```
sheenarbw: add one more test here to show that the right reviews count. 

Eg: make 2 projects such that:

```
len(project1.get_users_that_reviewed_since_last_review_request()) > 0
len(project2.get_users_that_reviewed_since_last_review_request()) == 0
```

sheenarbw (changes requested):
This is good work. But there is a bug we both didn't see.

backend/curriculum_tracking/tests/test_AgileCard.py
sheenarbw: Can you plkease seperate out your tests? One gigantic test function can work, but it's better if we have small specific things. 

Eg: 1 test function to make sure that the time is taken into account properly.
And then another test that makes sure that different projects have different results for `get_users_that_reviewed_since_last_review_request`.

Having one gigantic thing is very messy and it makes it very hard to see what your plan is.

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +688,83 @@ def test_cancel_workshop_attendance(self):
...
+        self.project.start_time = timezone.now() - timedelta(days=5)
+        self.project_2.start_time = timezone.now() - timedelta(days=5)
+        self.project.save()
+        self.project_2.save()
+        self.assertEqual(AgileCard.derive_status_from_project(self.project), AgileCard.IN_PROGRESS)
+        self.assertEqual(AgileCard.derive_status_from_project(self.project_2), AgileCard.IN_PROGRESS)
+
+        self.request_review_time = self.project.start_time + timedelta(1)
```
sheenarbw: why do you use self here? Just because we are in a class doesn't mean all of the variables have to be stapled to the instance.

If a variable is only used inside a single function and state doesn't need to be maintained then dont use self.

A useful rule of thumb for tests is to only assign things to instance variables if you are inside the setUp function eg:
```
def setUp(self):
    self.something_that_gets_used_in_multiple_OTHER_functions = something 
    something_that_is_only_referenced_in_this_one_function = "notice the absence of self. on this line"
```

backend/curriculum_tracking/tests/test_generate_agile_cards_for_recruits.py
```diff
@@ -235,7 +235,7 @@ def test_can_add_soft_prereq_to_cards_that_were_ready(self):
         ]
         create_or_update_content_cards_for_user(self.recruit, ordered_content_items)
 
-        cards = models.AgileCard.objects.all()
+        cards = models.AgileCard.objects.order_by("-id")
```
sheenarbw, line 238: Take note. I fixed the weird thing.

The order of the cards just kept swapping around.

backend/curriculum_tracking/tests/test_AgileCard.py
sheenarbw: Remember that good code makes it's intention clear.

If I say `self.something = anything` then I'm expressing the following intention:
1. I want to store the value in such a way as to make it accessable on the class instance
2. Because that way I can access it whenever I have acces to the class instance 
3. So I can bring the variable into scope when I'm not in the current scope 
4. because I intend to use the value while I'm not in the current scope 

But that isn't your intention at all. You are just using the variable in this one place. You dont need it anywhere else. 

Now imagine that we have a class that is actually complicated. It has 20 functions. Each of those functions just use self whenever they want to set up any new variable. How scrambled is that class going to look? How likely are bugs? How do we know what is actually needed, what is actually information we can rely on and what is jsut weird noise?

Just because `self` is available it doesn't mean we need to hit all the things with the `self` hammer

sheenarbw (changes requested):
I fixed the weird thing.

Now all that is left seems to be clean code issues.

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +688,104 @@ def test_cancel_workshop_attendance(self):
...
+
+        result = self.card_1.get_users_that_reviewed_since_last_review_request()
+        ids_which_can_be_returned = [review_2.reviewer_user.id, review_3.reviewer_user.id, review_4.reviewer_user.id]
+
+        # Making sure that reviews were done on card and not on card_2, if card_2 returns reviews then our function is
+        # returning the wrong stuff and therefore it is not working as it should.
+        self.assertEqual(sorted(result), sorted(ids_which_can_be_returned))
+        assert len(self.card_1.get_users_that_reviewed_since_last_review_request()) > 0
+
+    def test_get_users_that_reviewed_since_last_review_request_for_project_2(self):
```
sheenarbw: what does project_2 mean? 
Please make sure that the test function names are meaningful. Do you mean  Remember that people who aren't you will need to read this. If you name your functions after variables then that's not the way to go. We need to know the intention of the code.

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +688,104 @@ def test_cancel_workshop_attendance(self):
...
+        # Create a review on card_2, the review will be before the latest request for a review
+        review_1_card_2 = factories.RecruitProjectReviewFactory(
+            status=NOT_YET_COMPETENT,
+            recruit_project=self.project_2,
+            timestamp=time_1
+        )
+        review_1_card_2.timestamp = time_1
+        review_1_card_2.save()
+
+        # Review done on project_2, card_2 were done before the latest request for a review, therefore,
```
sheenarbw: You already tested the timing in the other test. What is this test for? Is it just a repeat?

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +688,104 @@ def test_cancel_workshop_attendance(self):
...
+        # get_users_that_reviewed_since_last_review_request() should return no user id's
+        project_2_return_results = self.card_2.get_users_that_reviewed_since_last_review_request()
+        assert len(project_2_return_results) == 0
```
sheenarbw: There seems to be a test missing. 

Here is some pseudocode:
```
project_1 = Project()
project_2 = Project()

reviewer_user = User()
add_ a_review_to_project_1(reviewer_user)

# the two projects have had different reviewers and therefore have different results
assert project_1.get_users_that_reviewed_since_last_review_request() == [reviewer_user.id]  
assert project_2.get_users_that_reviewed_since_last_review_request() == []

```

sheenarbw (changes requested):
There aer still some strange things in the tests: Some stuff is unclear and looks like a repetitive test. And some important stuff isn't tested

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +687,128 @@ def test_cancel_workshop_attendance(self):
...
+
+        review_4 = factories.RecruitProjectReviewFactory(
+            status=NOT_YET_COMPETENT,
+            recruit_project=self.project_1,
+        )
+        review_4.timestamp = self.time_four
+        review_4.save()
+
+        project_1_return_results = self.card_1.get_users_that_reviewed_since_last_review_request()
+        ids_which_can_be_returned = [review_2.reviewer_user.id, review_3.reviewer_user.id, review_4.reviewer_user.id]
```
sheenarbw: a better name would be `ids_that_should_be_returned` or `expected_ids`

If something `can` be returned then it sounds like it's ok if it's not returned. Eg: "you can pay your taxes" makes it seem like you dont really have to.

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +687,128 @@ def test_cancel_workshop_attendance(self):
...
+
+        # Setting a request for review and creating four different review times (No reviews done yet)
+        self.request_review_time = self.project_1.start_time + timedelta(1)
+        self.project_1.request_review(force_timestamp=self.request_review_time)
+        self.time_one = self.project_1.start_time - timedelta(days=6)
+        self.time_two = self.project_1.start_time + timedelta(days=4)
+        self.time_three = self.project_1.start_time + timedelta(days=3)
+        self.time_four = self.project_1.start_time + timedelta(days=2)
+
+    def test_functionality_of_users_that_reviewed_since_last_review_request_for_project_1(self):
```
sheenarbw: please remove the word `project_1` from the test name. I shouldn't go have to look at other pieces of code to know the purpose of the test. 

`test_that_if_there_are_reviews_before_and_after_request_times_then_the_correct_ids_are_returned` might be better. It's a bit long but at least it tells us what we are trying to achieve.

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +687,128 @@ def test_cancel_workshop_attendance(self):
...
+
+        project_1_return_results = self.card_1.get_users_that_reviewed_since_last_review_request()
+        ids_which_can_be_returned = [review_2.reviewer_user.id, review_3.reviewer_user.id, review_4.reviewer_user.id]
+
+        # Making sure that reviews were done on card and not on card_2, if card_2 returns reviews then our function is
+        # returning the wrong stuff and therefore it is not working as it should.
+        self.assertEqual(sorted(project_1_return_results), sorted(ids_which_can_be_returned))
+        assert len(self.card_1.get_users_that_reviewed_since_last_review_request()) > 0
+
+    def test_functionality_of_users_that_reviewed_since_last_review_request_for_project_2(self):
```
sheenarbw: What is the point of this test? It looks like t's testing the same thing as the previous test except with less detail. Does this test add value? If it's just a repeat then it should be removed. 

What is the purpose of using project_2 anstead of project_1 here? What does this test prove?

sheenarbw (changes requested):
There are still some strange things happening in the tests. One major thing you need to work on is just making your code's intewntion clear by choosing meaningful names. It's a skill that can be practiced and improved on.

It would also be good for you to read about unit tests and what they are for. What makes a good test.

Lastly, I still think you should clean up any git chaos on any of your branches before writing any code on any issues. Please see your other PRs for details.

EckardBerry:
I made some changes, not sure if they are correct.  I see the tag says "experienced-devs-only" and not good first issue.  I don't really know what else to do so can I ask Siphiwe or Viro to take this over, maybe when I see what they do a light will go on for me?

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +687,128 @@ def test_cancel_workshop_attendance(self):
...
+        ids_which_should_be_returned = [review_2.reviewer_user.id, review_3.reviewer_user.id, review_4.reviewer_user.id]
+
+        # Making sure that reviews were done on card and not on card_2, if card_2 returns reviews then our function is
+        # returning the wrong stuff and therefore it is not working as it should.
+        self.assertEqual(
+            sorted(self.card_1.get_users_that_reviewed_since_last_review_request()),
+            sorted(ids_which_should_be_returned)
+        )
+
+    def test_no_user_ids_should_return_since_review_done_before_latest_review_request(self):
```
TheAbstract: In `test_correct_ids_returned_before_or_after_review_request` above, the case where no ids are returned because there's no review since last request is checked. That's why `ids_which_should_be_returned` doest have the id from `review_1`, otherwise this assert would fail (which it doesn't)
```
self.assertEqual(
            sorted(self.card_1.get_users_that_reviewed_since_last_review_request()),
            sorted(ids_which_should_be_returned)
        )
```

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +687,128 @@ def test_cancel_workshop_attendance(self):
...
+        review_1_card_2.timestamp = time_of_review
+        review_1_card_2.save()
+
+        # Review done on project_2, card_2 were done before the latest request for a review, therefore,
+        # get_users_that_reviewed_since_last_review_request() should return no user id's
+        assert len(self.card_2.get_users_that_reviewed_since_last_review_request()) == 0
+
+    def test_two_different_projects_return_different_results_for_users_that_reviewed_since_last_review_request(self):
+
+        # Creating two new projects
```
TheAbstract: Since this is a method of `ReviewerIdsSinceLatestReviewRequest` we know from the context thats what it's doing. Isn't `test_two_different_projects_return_different_results` clear enough?

TheAbstract (commented):
@EckardBerry I ran the tests and everything it looks fine. Now that I've looked though things, would like to have a session were we walk through the code and then the tests.

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +687,128 @@ def test_cancel_workshop_attendance(self):
...
+            status=NOT_YET_COMPETENT,
+            recruit_project=self.project_2,
+            timestamp=time_of_review
+        )
+        review_1_card_2.timestamp = time_of_review
+        review_1_card_2.save()
+
+        # Review done on project_2, card_2 were done before the latest request for a review, therefore,
+        # get_users_that_reviewed_since_last_review_request() should return no user id's
+        assert len(self.card_2.get_users_that_reviewed_since_last_review_request()) == 0
```
sheenarbw: why did you choose to use assert instead of `self.assertEqual`?

We are using the unittest framework, it has certain conventions. Those conventions lead to nice logging of errors.

Eg if this assertion were to fail as it was then the traceback would say something like this: 

```
  File "/home/sheena/workspace/Tilde/backend/curriculum_tracking/tests/test_AgileCard.py", line 790, in test_no_user_ids_should_return_since_review_done_before_latest_review_request
    assert stuff
AssertionError
```
As you can see, that is not super informative.

If you replaced your assert with self.assertEqual then the error would look like this:

```
  File "/home/sheena/workspace/Tilde/backend/curriculum_tracking/tests/test_AgileCard.py", line 790, in test_no_user_ids_should_return_since_review_done_before_latest_review_request
    self.assertEquals(... things)
AssertionError: 1 != 2     # isn't that nice?
```

Can you please switch over to use the convenions in place instead of using standard assert statements

backend/curriculum_tracking/tests/test_AgileCard.py
sheenarbw: That would be clear enough.

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +687,128 @@ def test_cancel_workshop_attendance(self):
...
+        )
+
+        self.card_2 = factories.AgileCardFactory(
+            status=AgileCard.IN_PROGRESS,
+        )
+
+        self.project_1 = self.card_1.recruit_project
+        self.project_2 = self.card_2.recruit_project
+        self.user = self.card_1.assignees.first()
+        self.assertIsNotNone(self.card_1.assignees)
```
sheenarbw: Another thing to consider: What is the purpose of the setUp?

Is the setup a test? And if it is not a test why does it have so many assertions?

It would be much cleaner to just have setup things happen in the setup function, then have all your test assertions in appropriate test functions. 

Are these assertions even necessary? They dont seem to have much to do with the functionality under test. So it's not immediately obvious what you are trying to do here.

Is this just sanity checking?

Also you are checking a few odd things, eg `self.assertIsNotNone(self.card_1.assignees)`.  I dont think that will ever be none because the any_card.assignees gives us a thing that we can use to create a queryset.

Eg even if the assignees was empty then `any_card_ever.assignees` would not be None.  Because `any_card_ever.assignees.all()` should always work and `None.all()` will error.

Can you please clean up your setup method. If any of these assertions are necessary we'll need to find a place for them. It would probably be good to remove most of them.

sheenarbw (changes requested):
Getting there.

The code works though. Just needs a bit more polish and we are good to go

TheAbstract (approved):
Ran everything, all tests passing.

sheenarbw (approved):

PR #162, opened 2021-09-09
#154 TeamsList and ReviewTrustList component api and redux wiring

### LINKED ISSUE #154: Teams list component: Api and redux wiring
Basically what we need to do is wire up the component in #103 and make sure that it actually works.

Create an index.js file if it does not exist. then add this component to the Dashboard page's described in #155. 

At this point you should be able to view the trusts table if you navigate to the dashboard in the frontend

Make a PR at this point just to get a bit of feedback about structuring

Connect the revirew trust list index component to Redux and make it work. You'll need to use a few things correctly:

# mapStateToProps: 

This should be used for fetching the user from the redux store

We fetch all the known users from the store then use the user id to get the right user instance

# mapDispatchToProps:

If the user that we are trying to look at is not in the store, then we'll need to fetch the user from the backend.  

You'll want to make use of:  **apiReduxApps.FETCH_SINGLE_USER.operations.maybeStart**

# useEffect

This is where you'll actyually call the function that dispatches the action that causes the data to be fetched.

babalwa-mbolekwa:
Related issues: [https://github.com/Umuzi-org/Tilde/issues/154]

## Description:

## Screenshots/videos
dashboard_teams_and_reviewTrusts

## I solemnly swear that:

### DIFF
--- a/frontend/src/components/regions/Dashboard/Presentation.js
+++ b/frontend/src/components/regions/Dashboard/Presentation.js
@@ -12,19 +12,24 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-export default ({ user }) => {
+export default ({ 
+  user, 
+  userId 
+}) => {
   const classes = useStyles();
-  const teams =  user.teamMemberships;
-  return (
-    <React.Fragment>
-      <Grid container spacing={1}>
-        <Grid item xs={12}>
-          <Paper className={classes.paper}>
-            <Typography>You are a member of these teams:</Typography>
-            <TeamsTable teams={teams} />
-          </Paper>
+  const teams =  user ? user.teamMemberships : {};
+  if(userId)
+    return (
+      <React.Fragment>
+        <Grid container spacing={1}>
+          <Grid item xs={12}>
+            <Paper className={classes.paper}>
+              <Typography>You are a member of these teams:</Typography>
+              <TeamsTable teams={teams} />
+            </Paper>
+          </Grid>
         </Grid>
-      </Grid>
-    </React.Fragment>
-  )
+      </React.Fragment>
+    );
+  return <React.Fragment />;  
 };
--- a/frontend/src/components/regions/Dashboard/Presentation.js
+++ b/frontend/src/components/regions/Dashboard/Presentation.js
@@ -1,8 +1,7 @@
 import React from "react";
 import { Grid, Paper, Typography } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
-import ReviewTrustTable from "../UserDetails/ReviewTrustTable/Presentation";
-import TeamsTable from "../UserDetails/TeamsTable/Presentation";
+import TeamsTable from "../UserDetails/TeamsTable";
 
 const useStyles = makeStyles((theme) => ({
   paper: {
@@ -13,22 +12,16 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-export default () => {
+export default ({ user }) => {
   const classes = useStyles();
-  
+  const teams =  user.teamMemberships;
   return (
     <React.Fragment>
       <Grid container spacing={1}>
         <Grid item xs={12}>
           <Paper className={classes.paper}>
-            <Typography>TEAMS</Typography>
-            <TeamsTable />
-          </Paper>
-        </Grid>
-        <Grid item xs={12}>
-          <Paper className={classes.paper}>
-            <Typography>REVIEW TRUSTS</Typography>
-            <ReviewTrustTable />
+            <Typography>You are a member of these teams:</Typography>
+            <TeamsTable teams={teams} />
           </Paper>
         </Grid>
       </Grid>
--- a/frontend/src/components/regions/Dashboard/Presentation.js
+++ b/frontend/src/components/regions/Dashboard/Presentation.js
@@ -1,5 +1,5 @@
 import React from "react";
 
 export default () => {
-  return <div>TODO</div>;
-};
+  return <div>TODO</div>
+}
--- a/frontend/src/components/regions/Dashboard/Presentation.js
+++ b/frontend/src/components/regions/Dashboard/Presentation.js
@@ -1,5 +1,37 @@
 import React from "react";
+import { Grid, Paper, Typography } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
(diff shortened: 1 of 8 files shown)


frontend/src/components/regions/UserDetails/TeamsTable/Presentation.js
```diff
@@ -29,21 +29,23 @@ const TeamsTable = (props) => {
...
-                {   
-                    Object.keys(teams).length > 0 ?
-                        Object.values(teams).map((team) => (
-                            <TableRow key={Object.values(teams).indexOf(team)}>
-                                <TableCell>
-                                    {team.name}
-                                </TableCell>
+            {
+                <TableBody>
+                    {
```
sheenarbw, line 45: It looks like you have some extra brackets.

frontend/src/components/regions/Dashboard/index.js
```diff
@@ -1,6 +1,37 @@
...
+      <Grid container spacing={1}>
+        <Grid item xs={12}>
+          <Paper className={classes.paper}>
+            <Typography>TEAMS</Typography>
+            <TeamsTable />
+          </Paper>
+        </Grid>
+        <Grid item xs={12}>
+          <Paper className={classes.paper}>
+            <Typography>REVIEW TRUSTS</Typography>
```
sheenarbw: Frontends are about communication. We need to choose words wisely. Dont just give things the same name as the api.

How about "You are a trusted reviewer for these cards:"

frontend/src/components/regions/Dashboard/index.js
```diff
@@ -1,6 +1,37 @@
...
 export default () => {
-  return <Presentation />;
+  const classes = useStyles();
+  
+  return (
+    <React.Fragment>
+      <Grid container spacing={1}>
+        <Grid item xs={12}>
+          <Paper className={classes.paper}>
+            <Typography>TEAMS</Typography>
```
sheenarbw: WHY ARE YOU SHOUTING.

You could just use a heading component instead. Look at all the varients you can choose from: https://material-ui.com/components/typography/ 

It's also important to be consistent. Are there any other headings in Tilde that are in all caps? Nope. Not on the board, not on the actions  page, not on the login page...

Who will go update all those other headings to make them all capital letters?  Is that even necessary? Is there a reason why this page should have more capitalisation than the others?

frontend/src/components/regions/Dashboard/index.js
```diff
@@ -1,6 +1,37 @@
 import React from "react";
-import Presentation from "./Presentation";
+import { Grid, Paper, Typography } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
+import ReviewTrustTable from "../UserDetails/ReviewTrustTable/Presentation";
+import TeamsTable from "../UserDetails/TeamsTable/Presentation";
```
sheenarbw: Please make use of the index.js files. If we just draw the PResentation components then none of the redux wiring will actually execute.

Rather:
```
import ReviewTrustTable from "../UserDetails/ReviewTrustTable";
import TeamsTable from "../UserDetails/TeamsTable";
```

If you just import the Presentation then the components wont actually work. They'll just be pictures without functionality. We need the Dashboard to function. We dont just want a bunch of empty tables.

frontend/src/components/regions/Dashboard/index.js
```diff
@@ -1,6 +1,37 @@
...
+            <Typography>TEAMS</Typography>
+            <TeamsTable />
+          </Paper>
+        </Grid>
+        <Grid item xs={12}>
+          <Paper className={classes.paper}>
+            <Typography>REVIEW TRUSTS</Typography>
+            <ReviewTrustTable />
+          </Paper>
+        </Grid>
```
sheenarbw: All this layout stuff should be in dashboard/presentation.js. 
Index.js is for smarts, not for layout.

frontend/src/components/regions/Dashboard/index.js
sheenarbw: Of course these iomports should be in the Dashboard/Presentation.js file. This is not the place for layout

frontend/src/components/regions/Dashboard/Presentation.js
```diff
@@ -1,5 +1,5 @@
 import React from "react";
 
 export default () => {
-  return <div>TODO</div>;
-};
\ No newline at end of file
+  return <div>TODO</div>
```
sheenarbw: All layout should go in here. 
Please import child components and make sure this Presentation file lays them out correctly.

sheenarbw (changes requested):
The file structure needs work.

Also, I see no mention of redux in the code you submitted in this PR.  But the PR title mentions redux. Is the title incorrect or is there code missing?

frontend/src/components/regions/UserDetails/TeamsTable/redux/operations.js
```diff
@@ -0,0 +1,9 @@
+import Creators from "./actions.js";
+
+export default {
+    ...Creators,
+    openTeamsTable: ({ userId }) =>
```
sheenarbw: What is this for? I dont think the teams table is a modal.

sheenarbw (changes requested):
You included a bunch of weird stuff in your PR. Please remove the unnecessary things.

babalwa-mbolekwa:
teamsDashboard

Vuyisanani (commented):
Hi @Babalwa01 

Can you please make sure you have the requested changes resolved, mark them resolve or they should be outdated

Vuyisanani (commented):
Don't you also have to write unit tests?

babalwa-mbolekwa:
> You included a bunch of weird stuff in your PR. Please remove the unnecessary things.

attended to changes

frontend/src/components/regions/UserDetails/TeamsTable/redux/operations.js
babalwa-mbolekwa: > What is this for? I dont think the teams table is a modal.

attended to changes requested

babalwa-mbolekwa:
> Don't you also have to write unit tests?

@Vuyisanani not quite sure how I can structure them for this ticket. Kindly assist with the approach

Vuyisanani (approved):

sheenarbw (approved):

PR #167, opened 2021-09-13
Email password login update

### LINKED ISSUE #153: Create/update a story that displays the login page. Users should be able to login with email/password
Please use the branch protected/feat/email-password-login as a base branch.

Sbonelo01:
Related issues: [https://github.com/Umuzi-org/Tilde/issues/153]

## Description:

## I solemnly swear that:

The idea is to keep things simple and minimalistic

Screenshot from 2021-09-14 00-24-27


babalwa-mbolekwa (changes requested):
Hey Sbonelo the UI looks good on storybook but can be improved:
- Please also include and use the props that were originally used.
- Please also keep the alerts and the heading that was used, also take note of the conditions used with the alerts
- I get this error when running it on Tilde localhost `./src/components/regions/Login/Presentation.js`
`Attempted import error: 'createTheme' is not exported from '@material-ui/core/styles'.` please check if **createTheme** is imported correctly.

babalwa-mbolekwa (dismissed):

sheenarbw (changes requested):
Why not just add the form to the login page that we already have? Why do we need 2 seperate login pages? If we allowed login with Github or Rocketchat then would we have 4 login pages?

We could have a page that has the form and also has a "login with google" button.  Why not?

Sbonelo01:
> Why not just add the form to the login page that we already have? Why do we need 2 seperate login pages? If we allowed login with Github or Rocketchat then would we have 4 login pages?
> 
> We could have a page that has the form and also has a "login with google" button. Why not?

@sheenarbw  That makes sense. Looking at the new updated Prioritised backlog frontend issues I was thinking of creating separate components as stories and once you approve how they look, we can glue everything together. I could be thinking about this in the wrong way?

sheenarbw:
@Sbonelo01 
Please always update the code that is meant to be updated.  If we do it your way then it's at least double the work. Imagine:

1. You write a login form frontned and a story for it (that takes some doing) Note that we will need to delete all these files later.
2. I tell you it's missing the "login with google" button so you add that in and eventuially get it looking good.
3. Now it looks perfect so you update the existing login page with the new layout. You also make sure that the story that illustrates the final corect login page. Unfortunately the code you wrote wasn't designed to be integrated into the existing machine so this process is error prone
4. delete everything you did from 1 since it has now been copy pasted into the correct place. 
5. Get another code review from me - making me do a repetitive job
6. I find that your code isa actually incompatable with what is in place or that you forgot to purge all the useless and unintegratable code. You are likely to make mistakes when actually wtriting code that is meant to be integrated with what is in place. You have also increased the chance of human error significantly
7. You clean up and eventually we get to merge. 

The better way:
1. Updaste the page you were meant to update. Work on the story you were meant to work on. Put things in the final place. Dont put the code in arbitrary locations. Dont write files that are just meant to be deleted.
2. Ask for a code review
3. I tell you about any integration problems and you fix those
4. get to merge smoothly

Sbonelo01:
Screenshot from 2021-09-21 19-55-34

babalwa-mbolekwa (changes requested):
Awesome work Sbonelo! 
Please also put the _Please log in_ heading and also remove the `{ useState }`  import that's causing a warning on the Login Presentation.js file

babalwa-mbolekwa (dismissed):
Well done Sbonelo!

frontend/src/stories/Login.stories.js.orig
```diff
@@ -0,0 +1,9 @@
+import React from "react"
+import LoginForm from "../components/regions/Login/LoginForm"
```
sheenarbw: please remove this file.

When you do a self-review you should always check that there aren't any weird extra files that slipped into the PR.

sheenarbw (changes requested):
I haven't reviewed everything. There is at least one thing to clean up here

babalwa-mbolekwa (changes requested):
Very good work Sbonelo. Please also add the needed props to the Login story, e.g :
since the Login component makes use of `loading`, `error`, `handleLoginWithGoogle` then you can create your story like:
`export const Primary = (args) => <LoginForm {...args} />`
`Primary.args = {`
`loading: false,`
`error: "",`
`handleLoginWithGoogle: () => {},`
`}`

depending on each prop's datatype. For props that are objects you can use the ones from `src/stories/fixtures` or create a different sample object within the story file.

You'll see the props it uses in the bottom "Controls" section on storybook

babalwa-mbolekwa (approved):
Well done Sbonelo!

sheenarbw (approved):

PR #183, opened 2021-09-20
Create api exposing user stats/157

### LINKED ISSUE #157: Create apis that expose the following User stats
add a "stats" action to the user viewset. Expose the following info:

total number of cards complete, in review, etc (each card status)
total number of Reviews done on Tilde
total number of PR reviews done

total number of reviews done in last 7 days 
total nimber of PR reviews done in last 7 days
number of cards completed in the last 7 days
number of cards started in last 7 days

If you think of other user-specific stats that should be drawn on a dashboard, please feel free to make suggestions

EckardBerry:

## Description:

1. I added some functions to core/models.py for the api endpoint
2. I added a new serialisezer class to core/serializers
3. I created a file where my 'tests' are in, once I am happy with everything this file will be deleted and a proper test will be added to the file test_AgileCard.py
4. I added a bit of code to core/views/UserViewSet for the added functionality of stats per user

## I solemnly swear that:


backend/core/models.py
```diff
@@ -159,6 +169,106 @@ def github_name(self):
...
+            assignees=self.id
+        )
+        return len([card.id for card in cards_in_completed_column_amount])
+
+    @property
+    def user_cards_in_review_column(self):
+        cards_in_review_column_amount = curriculum_tracking.models.AgileCard.objects.filter(
+            status=curriculum_tracking.models.AgileCard.IN_REVIEW,
+            assignees=self.id
+        )
```
sheenarbw: use `.count()` instead of `len`

As in `SomeModel.objects.filter(whatever).count()` 
This does a count in SQL. It's much faster because it it doesn't need to transfer all the data to Python, it just sends one number.

backend/core/models.py
```diff
@@ -9,11 +9,21 @@
...
 
 # from rest_framework.generics import RetrieveAPIView
+import curriculum_tracking
+import git_real
 from model_mixins import Mixins
 from django_countries.fields import CountryField
 from django.contrib.auth.models import Group as AuthGroup
-
 from django.contrib.auth.models import PermissionsMixin
+from curriculum_tracking.constants import (
```
sheenarbw: I worry about circular imports happening.

As much as possible, we need to allow curriculum_tracking to import from core, but not the other way around.

If you do need to import curriculum_tracking from within core, then please do it inside a function. Just in time imports reduce the risk of circular imports causing problems.

backend/core/serializers.py
```diff
@@ -136,3 +136,63 @@ class Meta:
...
     class Meta:
         fields = []
+
+
+class UserStatsPerWeekSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.User
+        fields = [
+            "id",
+            "user_email",
```
sheenarbw: why user_email?  If we are looking at a user on the frontend then we will already know their email. 

If this is used by another process then the user email can be fetched using the standard users api.

backend/core/views.py
```diff
@@ -17,6 +18,13 @@
 from core import permissions as core_permissions
 from core.filters import ObjectPermissionsFilter
 from core.models import Team
+from curriculum_tracking.models import (
```
sheenarbw: Are you using these imports anywhere? I can't see these models mentioned in this file.

backend/curriculum_tracking/tests/factories.py
```diff
@@ -1,5 +1,7 @@
 from factory.django import DjangoModelFactory
 import factory
+
+import git_real.models
```
sheenarbw: what is this for?

backend/curriculum_tracking/tests/for_testing_purposes.py
```diff
@@ -0,0 +1,278 @@
+import unittest
```
sheenarbw: all the tests are for testing purposes. Can you think of a more descriptive file name?

Generally it's best to try to choose good names off the bat, going back and having to clean stuff up later isn't ideal

backend/curriculum_tracking/tests/test_creating_cards_and_reviews.py
```diff
@@ -0,0 +1,58 @@
...
+
+from curriculum_tracking.constants import (
+    NOT_YET_COMPETENT,
+    COMPETENT,
+    EXCELLENT,
+)
+from django.utils import timezone
+
+
+class CreatingCardsAndReviews(TestCase):
```
sheenarbw: Are you sure this test runs?

backend/curriculum_tracking/tests/for_testing_purposes.py
```diff
@@ -0,0 +1,278 @@
...
+            user_id=self.repo_card_one.assignees.first().id
+        )
+        pr_review_two_weeks_ago.save()
+
+        pr_review_two_days_before_yesterday = PullRequestReview.objects.create(
+            html_url="https://github.com/Umuzi-org/PullRequest-Review(4)",
+            pull_request_id=self.pr_two_days_before_yesterday.id,
+            author_github_name="DeadManWalking",
+            submitted_at=two_days_before_yesterday,
+            body="This review happened no more than two days ago",
```
sheenarbw: This test is very long and complicated.

Would it be possible to test 1 field at a time instead?

sheenarbw (changes requested):
There are a bunch of things mentioned inline. And then there's one weird thing that might be a bit hard to explain. But I'll try:

When it comes to architecting complex apps, you always need to think of the flow of knowledge (via imports). With core and curriculum_tracking, think of core as something foundational, and curriculum as something built on top of core. 

Ideally we will keep the imports going in one direction. So curriculum_tracking can import from core. But core should not import from curriculum_tracking.

Why? 
1. Reduces the chance of circular imports. This is important because it might jsut break the code
2. It makes the code simpler to reason about. If a needs parts of b, and b needs pars of a and you draw out all the depenencies as arrows, you want the pictrure to be simple. If we aren't disciplined here then things can very quickly turn into a big pile of spagetti.

How can we fix this?

I would put the new serialiser inside curriculum_tracking. I would also make the serialiser smart enough to figure out what it needs to know without having to make changes to the UserManager class. Ideally there should be no changes to core.models to make this functionality work. 

The tricky parts are:
1. testing: you would then need to make use of api testing to make sure the data pulls through. Basically simulate GET requests.
2. there is still a circular part to the imports, but at least it is only one import in one place. It has a smaller blast radius if things go wrong later. It's easier to reason about.

backend/curriculum_tracking/tests/factories.py
EckardBerry: I removed the import, I was thinking about creating a review factory for PR's but then decided against it and forgot to remove the import.

backend/curriculum_tracking/tests/for_testing_purposes.py
EckardBerry: I should actually just add the file to .gitignore because it's not actually what I want at the the end of the day, it's just something I use to fool around with and check that stuff works as I expect them to work.  That's why it doesn't have very descriptive name seeing that I am going to delete it anyway.

backend/curriculum_tracking/tests/test_creating_cards_and_reviews.py
EckardBerry: Yeah this test runs but it will be deleted cause it's not an actual 'test' test, it's just for fooling around and checking how stuff works.  I will delete this once I am happy everything works as it should.

backend/curriculum_tracking/tests/for_testing_purposes.py
EckardBerry: I agree, I will make it a lot shorter, I created a lot of data just to fool around with but I agree with you that to have this as a test for the new api endpoint is not good cause it is way too long.

backend/core/models.py
EckardBerry: Did this now, it works well, thanks for this.

backend/core/models.py
EckardBerry: Don't need these constants anymore so I removed the import.

backend/core/serializers.py
EckardBerry: I got rid of the user_email, it was actually an easy way for me to see that I am getting the correct user and their stats.  But it's gone now.

backend/core/views.py
EckardBerry: Got rid of this import as well.

EckardBerry:
Hi Sheena

Will you please have a look at the function "get_cards_in_completed_column" in curriculum_tracking/serializers.py.  You will see that the functionality for this function alone use to exist in core/models.py.  I commented it out there for now and moved all the functionality to the serialiser.  Is this what you were thinking about when you mentioned in your comments about moving my new serializer to curriculum_tracking/serializers and making the serializer smart enough to do what it should do?

If so, my intention is to do the same to the rest of the functions as well.

backend/core/models.py
```diff
@@ -9,11 +9,12 @@
 # import re
 
 # from rest_framework.generics import RetrieveAPIView
+import git_real
```
sheenarbw: what is this import for? You just import git_real later on.

Can you please tidy up your imports?

backend/core/models.py
```diff
@@ -159,6 +160,132 @@ def github_name(self):
...
+        cards_in_completed_column_amount = AgileCard.objects.filter(
+            status=AgileCard.COMPLETE,
+            assignees=self.id
+        ).count()
+
+        return cards_in_completed_column_amount
+    """
+
+    @property
+    def user_cards_in_review_column(self):
```
sheenarbw: I still think this stuff should be implemented in the curriculum_tracking app instead of inside `core`.

Why do you want it here? Why do we want things to be so tangled up?

backend/curriculum_tracking/serializers.py
```diff
@@ -323,3 +323,69 @@ class Meta:
...
+
+        self.cards_in_completed_column_amount = AgileCard.objects.filter(
+            status=AgileCard.COMPLETE,
+            assignees=instance.id
+        ).count()
+
+        return self.cards_in_completed_column_amount
+
+    def get_cards_in_review_column(self, instance):
+        return instance.user_cards_in_review_column
```
sheenarbw: rather implement things here instead of on the User model

backend/curriculum_tracking/tests/for_testing_purposes.py
sheenarbw: ok cool.

There's a trick I use. If you make a directory named `gitignore` then you  can put your rough work files in there and they'll get ignored. That way you dont need to add each file individualy

sheenarbw (changes requested):
progress :)

backend/core/models.py
EckardBerry: https://github.com/Umuzi-org/Tilde/pull/183#issuecomment-925282516

Hi Sheena, this relates to my question (https://github.com/Umuzi-org/Tilde/pull/183#issuecomment-925282516).  You will see I moved only the 1st function and all of it's functionality inside core/serializers.py, should it be in curriculum_tracking/serializers.py?

backend/core/models.py
EckardBerry: I cleaned up, hope I did not miss anything.

backend/core/serializers.py
```diff
@@ -135,4 +135,4 @@ class Meta:
 
 class UserStatsSerializer(serializers.Serializer):
```
sheenarbw: can you please remove this class? I think the one you made is better, thos one is just technical debt now

backend/core/views.py
```diff
@@ -146,6 +148,19 @@ class UserViewSet(viewsets.ModelViewSet):
     queryset = models.User.objects.all().order_by("last_name")
     serializer_class = serializers.UserSerializer
 
+    @action(
+        detail=True,
+        methods=['get'],
+        serializer_class=UserStatsPerWeekSerializer
+    )
+    def stats(self, request, pk=None):
```
sheenarbw, line 178: good

backend/core/views.py
```diff
@@ -1,3 +1,5 @@
+import backend.settings
```
sheenarbw: why are you importing backend.settings? Is it used in this file? I cant see it being used anywhere

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +694,199 @@ def test_cancel_workshop_attendance(self):
         self.assertIsNone(self.card.workshop_attendance)
         self.assertEqual(WorkshopAttendance.objects.count(), 0)
         self.assertEqual(self.card.status, AgileCard.READY)
+
+
+class TestingForTheStatsAPI(TestCase):
```
sheenarbw: This is not a good place for these tests.  This file is for testing AgileCard methods.

 You are testing a class named `UserStatsPerWeekSerializer` so it would be better to have a file naemed test_UserStatsPerWeekSerializer or test_serializers

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +694,199 @@ def test_cancel_workshop_attendance(self):
...
+
+    def test_creating_cards_and_testing_functionality_of_stats_api(self):
+
+        # One card with 4 reviews to follow on it
```
sheenarbw: instead of having one gigantic test that does all of the things, it would be better to break things down.

How about one little test function per serialiser field? Then the next person reading this code will be able to see what you are trying to do. And if something goes wrong then they will be able to figure it out.

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -688,3 +694,199 @@ def test_cancel_workshop_attendance(self):
...
+        self.card_1 = AgileCardFactory(
+            status=AgileCard.IN_PROGRESS,
+        )
+
+        self.project_1 = self.card_1.recruit_project
```
sheenarbw: If a variable is only used inside one function then there is no reason to use self. In fact using `self` when you shouldnt is bad practice.

Imagine if you have a complicated class that has 50 functions. Inside each function, whenever there is a variable declared then it is added to the instance `self.thingy = whatever`.

Now since there are 50 functions some names get reused in different functions. So calling one function can ffect how other functions behave in unintended ways. This would lead to bugs.

It would also lead to confusion - imagine a new developer joining the team and needing to understand the code. There is no way of seeing which of the variables are only meant to be in scope inside the function or which variables are meant to be accessed on the instance. This becomes like spegetti which leads to wasted time and more bugs.

Only use self if: You need the variable to be in scope outside the current function.

You might want to look into the principle of least privilage

sheenarbw (changes requested):
Getting there. The tests need some wqork still.

There are somne lessons to learn still. Let me know if I said anything confusing. Some stuff is hard to explain.

backend/core/views.py
EckardBerry: When I use a module from somewhere and it is not imported then Pycharm imports it automatically without me knowing about it.  I removed it, I will go and check in Pycharm's settings to see where I can turn off automatic imports.

backend/curriculum_tracking/tests/test_AgileCard.py
EckardBerry: Created a separate file named test_UserStatsPerWeekSerializer

backend/curriculum_tracking/tests/test_AgileCard.py
EckardBerry: I created a bunch of separate tests and added comments inside each to try and explain what is happening.

backend/curriculum_tracking/tests/test_AgileCard.py
EckardBerry: I removed most usages of 'self' outside the setUp function.

.gitignore
```diff
@@ -3,3 +3,4 @@ gitignore/
 venv/
 *.log
 certbot.out
+backend/curriculum_tracking/tests/for_testing_purposes.py
```
sheenarbw: Why not just delete the file? 

I'm worried that out gitignore will become very noisy if we just put lots of random stuff in there.

backend/core/serializers.py
sheenarbw: Same comment as before

backend/curriculum_tracking/tests/test_UserStatsPerWeekSerializer.py
```diff
@@ -0,0 +1,234 @@
...
+        # the completed column
+        assert UserStatsPerWeekSerializer.get_cards_in_completed_column(
+            UserStatsPerWeekSerializer, User=self.user
+        ) == 0
+
+        # In total we did start new cards within the last 7 days but we did not complete any cards in the last seven
+        # days
+        assert UserStatsPerWeekSerializer.get_cards_started_last_7_days(
+            UserStatsPerWeekSerializer, User=self.user
+        ) > 0
```
sheenarbw: Can we be more specific? What exact number are we expecting here?

Most of the time it makes sense to assert specific things

backend/curriculum_tracking/tests/test_UserStatsPerWeekSerializer.py
```diff
@@ -0,0 +1,234 @@
...
+
+        content_item_2 = ContentItemFactory(
+            content_type=ContentItem.TOPIC,
+            project_submission_type=ContentItem.REPOSITORY,
+        )
+        card_2 = AgileCardFactory(
+            content_item=content_item_2,
+            status=AgileCard.COMPLETE,
+        )
+        assignee__ = SocialProfileFactory().user
```
sheenarbw: Why `assignee__` instead of just `assignee`? What are you trying to tell us with this snew naming convention?

sheenarbw (changes requested):
Getting there.

There are a couple of little issues left that I spotted. And one unresolved comment from my last review.

I highly recommend that you sort out your git feature branching BEFORE making any changes to the code. Otherwise stuff is just going to get more tangled up and weird. Please see your other PR for more details.

EckardBerry:
I made the requested changes, I mentioned in the other PR about the tag that says "experienced-devs-only", can I ask Siphiwe or Viro to take this over then maybe when I see what they did things will become clearer?

backend/core/serializers.py
sheenarbw: Can you please remove this class?

backend/core/views.py
```diff
@@ -146,6 +147,19 @@ class UserViewSet(viewsets.ModelViewSet):
...
 
+    @action(
+        detail=True,
+        methods=['get'],
+        serializer_class=UserStatsPerWeekSerializer
+    )
+    def stats(self, request, pk=None):
+        serializer = self.get_serializer(data=request.data)
+        if serializer.is_valid():
+            card = self.get_object()
```
sheenarbw: Are you sure this is a card? We are dealing with the UserViewSet here aren't we?

backend/curriculum_tracking/tests/test_creating_cards_and_reviews.py
sheenarbw: So you dont actually want this PR to be merged? Can you please delete this file?

backend/curriculum_tracking/tests/test_UserStatsPerWeekSerializer.py
```diff
@@ -0,0 +1,219 @@
...
+
+        self.pull_request = git_real_factories.PullRequestFactory(
+            repository=self.repo_card_one.recruit_project.repository, updated_at=self.today
+        )
+        self.pull_request.save()
+
+    def test_one_card_in_progress_before_any_reviews(self):
+
+        # One card, no reviews just yet, function 'get_cards_in_progress_column' should return one card
+        assert self.card_1.status == AgileCard.IN_PROGRESS
```
sheenarbw: can yo uplease use `self.assertEquals`

We aren't using pytest. This unit testing framework works differently

backend/curriculum_tracking/tests/test_UserStatsPerWeekSerializer.py
```diff
@@ -0,0 +1,219 @@
...
+            author_github_name="DeadManWalking",
+            submitted_at=self.two_weeks_ago,
+            body="This review happened a long time ago.",
+            commit_id=self.repo_card_one.assignees.first().id,
+            state="closed",
+            user_id=self.repo_card_one.assignees.first().id
+        )
+        pr_review_two_weeks_ago.save()
+
+        pr_review_two_days_before_yesterday = PullRequestReview.objects.create(
```
sheenarbw: All these object creations seem quite verbose. It would be better to use factoryboy to create a factory and then use the factory. Then other people can also use the factory in their tests. It saves time in the medium term

sheenarbw (changes requested):
Getting there - there is still some tidying up to do. But close

backend/core/views.py
EckardBerry: card was the wrong variable name to put there, changed it now to user_object.

backend/curriculum_tracking/tests/test_UserStatsPerWeekSerializer.py
EckardBerry: Made the change now.

backend/curriculum_tracking/tests/test_UserStatsPerWeekSerializer.py
EckardBerry: I've now created a PullRequestReview factory and adjusted the tests accordingly.

backend/curriculum_tracking/tests/factories.py
```diff
@@ -1,5 +1,7 @@
...
 from curriculum_tracking import models
```
TheAbstract: If `git_real.models` isn't being used, I think it's better to remove it. Especially since `models` from `curriculum_tracking`  is being used through out, can be a bit confusing.

backend/curriculum_tracking/tests/test_UserStatsPerWeekSerializer.py
```diff
@@ -0,0 +1,196 @@
...
+        )
+        review_1.timestamp = time_one
+        review_1.save()
+
+        # review_1 had a status of NYC, so we should have at least one card in the 'REVIEW FEEDBACK' column
+        self.assertEqual(UserStatsPerWeekSerializer.get_cards_in_review_feedback_column(
+            UserStatsPerWeekSerializer, User=self.user), 1
+        )
+
+    def test_no_card_in_completed_column_after_competent_review_reviewer_not_trusted_reviewer(self):
```
TheAbstract:
```suggestion
    def test_no_card_in_completed_column_after_untrusted_competent_review(self):
```

TheAbstract (changes requested):
@EckardBerry this test is failing: `test_pr_reviews_card_in_different_column_due_to_review_status`. 

```cmd
ERROR: test_pr_reviews_card_in_different_column_due_to_review_status (curriculum_tracking.tests.test_UserStatsPerWeekSerializer.TestingForTheStatsAPI)
  File "/umuzi/Tilde/backend/venv/lib/python3.8/site-packages/django/db/backends/utils.py", line 82, in _execute
    return self.cursor.execute(sql)
psycopg2.errors.ForeignKeyViolation: insert or update on table "git_real_pullrequestreview" violates foreign key constraint "git_real_pullrequest_pull_request_id_a7d0c00c_fk_git_real_"
DETAIL:  Key (pull_request_id)=(3) is not present in table "git_real_pullrequest".
```

Also, we're testing to see that no reviews older than a week actually get counted here right? If so, the test name doesn't really make that clear.

backend/curriculum_tracking/tests/factories.py
EckardBerry: I updated the name of the function a bit.  I see it is not the test that fails, it has something to do with the test_db that django creates when running tests.  The error is weird, I will have to ask for Sheena's help on it.

backend/core/serializers.py
```diff
@@ -131,8 +131,3 @@ class Meta:
         fields = ["message"]
 
     message = serializers.CharField()
-
-
-class UserStatsSerializer(serializers.Serializer):
```
sheenarbw, line 136: thanks :)

backend/curriculum_tracking/tests/factories.py
```diff
@@ -336,3 +338,18 @@ class Meta:
     attendee_user = factory.SubFactory(UserFactory)
     content_item = factory.SubFactory(ProjectContentItemFactory)
     timestamp = timezone.now()
+
+
+class PullRequestReviewFactory(DjangoModelFactory):
```
sheenarbw: shouldn't this be inside git_real.tests.factories?

backend/curriculum_tracking/tests/factories.py
```diff
@@ -336,3 +338,18 @@ class Meta:
...
+    class Meta:
+        model = git_real.models.PullRequestReview
+
+    html_url = factory.lazy_attribute(lambda *args, **kwargs: next(_content_url_iterator))
+    pull_request = factory.SubFactory(PullRequestFactory)
+    author_github_name = factory.SubFactory(UserFactory)
```
sheenarbw: Please double check the field types. The author name is a CharFiend, it's a string. You are treating it like a foreign key to the user model.

sheenarbw (changes requested):
I was doing some constructive procrastination and decided to review your stuff. You are getting there, your factory is mostly looking good but I think it's in the wrong file. And it's very important to be familiar with the model that you are trying to instantiate, at least one of the fields is using the wrong type. 

But we are getting there...

This will be very useful once it is done

backend/curriculum_tracking/tests/factories.py
EckardBerry: I moved it there now

backend/curriculum_tracking/tests/factories.py
EckardBerry: Made some changes to the fields:  author_github_name, html_url and commit_id.  I created some generators for commit_id and html_url

backend/curriculum_tracking/serializers.py
```diff
@@ -334,6 +335,149 @@ class Meta:
...
+    cards_in_review_feedback_column = serializers.SerializerMethodField("get_cards_in_review_feedback_column")
+    cards_in_progress_column = serializers.SerializerMethodField("get_cards_in_progress_column")
+    cards_completed_last_7_days = serializers.SerializerMethodField("get_cards_completed_last_7_days")
+    cards_started_last_7_days = serializers.SerializerMethodField("get_cards_started_last_7_days")
+    total_number_of_tilde_reviews = serializers.SerializerMethodField("get_total_number_of_tilde_reviews")
+    tilde_reviews_done_last_7_days = serializers.SerializerMethodField("get_tilde_reviews_done_last_7_days")
+    total_number_of_pr_reviews = serializers.SerializerMethodField("get_total_number_of_pr_reviews")
+    pr_reviews_done_last_7_days = serializers.SerializerMethodField("get_pr_reviews_done_last_7_days")
+
+    def get_cards_in_completed_column(self, User):
```
sheenarbw: is User a class or an instance of a class? Because this name says it is a class

backend/curriculum_tracking/serializers.py
```diff
@@ -334,6 +335,149 @@ class Meta:
...
+
+        cards_in_completed_column_amount = AgileCard.objects.filter(
+            status=AgileCard.COMPLETE,
+            assignees=User.id
+        ).count()
+
+        return cards_in_completed_column_amount
+
+    def get_cards_in_review_column(self, User):
+        from curriculum_tracking.models import AgileCard
```
sheenarbw: why import this over and over again? Can't we import it once? Or just use models.AgileCard?

backend/curriculum_tracking/serializers.py
```diff
@@ -334,6 +335,149 @@ class Meta:
...
+        ).count()
+
+        return cards_completed_past_seven_days
+
+    def get_cards_started_last_7_days(self, User):
+        from curriculum_tracking.models import AgileCard
+
+        cards_started_past_seven_days = AgileCard.objects.filter(
+            assignees=User.id,
+            recruit_project__start_time__gte=datetime.now() - timedelta(days=7)
```
sheenarbw: use timezone instead of datetime. Otherwise we'll get all sorts of warnings. Our db is timezone aware

backend/curriculum_tracking/serializers.py
```diff
@@ -334,6 +335,149 @@ class Meta:
...
+        ).count()
+
+        return cards_started_past_seven_days
+
+    def get_total_number_of_tilde_reviews(self, User):
+        from curriculum_tracking.models import RecruitProjectReview, TopicReview
+
+        tilde_project_reviews_done_to_date = RecruitProjectReview.objects.filter(
+            reviewer_user_id=User.id
+        ).all().count()
```
sheenarbw, line 426: you don't need .all() here.

backend/curriculum_tracking/serializers.py
```diff
@@ -334,6 +335,149 @@ class Meta:
...
+            recruit_project__start_time__gte=datetime.now() - timedelta(days=7)
+        ).count()
+
+        return cards_started_past_seven_days
+
+    def get_total_number_of_tilde_reviews(self, User):
+        from curriculum_tracking.models import RecruitProjectReview, TopicReview
+
+        tilde_project_reviews_done_to_date = RecruitProjectReview.objects.filter(
+            reviewer_user_id=User.id
```
sheenarbw: you could filter by `reviewer_user = user`, you dont need to mention ids, it will know what you mean

sheenarbw (changes requested):
some clean code things

TheAbstract (dismissed):

backend/curriculum_tracking/serializers.py
TheAbstract: @sheenarbw I think there's some misunderstanding about this. When we discussed this, Eckard said you wanted it like this to avoid circular imports?

sheenarbw (approved):

PR #202, opened 2021-10-05
Django admin delete teams which are inactive/110

EckardBerry:

## Description:
- I've made some changes to core/admin by adding the extra function needed for this new functionality
- I've added a new test to core/tests, it is called "test_added_admin_functionality.py"

## I solemnly swear that:


backend/core/admin.py
```diff
@@ -1,12 +1,25 @@
...
 from guardian.admin import GuardedModelAdmin
+from rest_framework.decorators import action
+from core.permissions import IsStaffUser
 
 
 class UserSetInline(admin.TabularInline):
     model = models.User.groups.through
     raw_id_fields = ("user",)
 
+@action(
```
sheenarbw: This is odd. You are mixing up DRF actions with admin panel things. 

The django admin panel has nothing to do with DRF. 

If you think it would be better to expose this functionality as an api endpoint then it would be better to configure it in the api_views.

The admin panel shouldn't know about the apis. If we deleted drf then this functiopnality should work

backend/core/admin.py
```diff
@@ -1,12 +1,25 @@
...
+    detail=True,
+    permission_classes=[IsStaffUser]
+)
+def make_members_inactive_for_inactive_teams(TeamAdminInstance, request, queryset: object):
```
sheenarbw: The issue says: "Expose functionality on the Django admin panel that would allow a user like Ryan to deactivate or activate all members of a specific team "

"a specific team" means 1 team, not many teams. 

That said, I'm not certain how this will look on the admin panel. Are there checkboxes on the teams and then an option to take an action on all those teams at once?

backend/core/tests/test_added_admin_functionality.py
```diff
@@ -0,0 +1,37 @@
+from core.tests import factories
```
sheenarbw: I dont think this is a good name for the file. What if we add other admin functionality at a later date? Would we make a new file called "test_more_added_admin_functionality_that_we_Added_later.py"?

Please focus on names. If I look at this file name then I should know what is actually happening inside the file. What functionality are we testing specifically? We are testing something very specific here.

backend/core/admin.py
```diff
@@ -26,8 +39,8 @@ class TeamAdmin(GuardedModelAdmin):
             },
         ),
     )
-
     inlines = [UserSetInline]
+    actions = [make_members_inactive_for_inactive_teams]
```
sheenarbw: I dont think that this is a good name considering what we want to achieve. Just because a team is inactive doesn't mean we always want to deactivate all the users.

We need a function where we can 
1. Select a team (any team at all)
2. Deactivate all users in the team.

A better name would be `deactivate_team_members`

backend/core/tests/test_added_admin_functionality.py
```diff
@@ -0,0 +1,37 @@
...
+    def test_make_members_inactive_for_inactive_teams_should_make_all_members_inactive_of_inactive_team(
+            self, mocked_queryset
+    ):
+        # Making sure the team is inactive
+        assert mocked_queryset.return_value[0].active == False
+        assert mocked_queryset.return_value[0].active_users != None
+
+        # Now calling make_members_inactive_for_inactive_teams will make all members in the team inactive
+        make_members_inactive_for_inactive_teams(TeamAdmin, HttpRequest, mocked_queryset)
+        assert len(mocked_queryset.return_value[0].active_users) == 0
```
sheenarbw: This isn't how the function is meant to work. Also I dont think that this team has any members at all

backend/curriculum_tracking/models.py
```diff
@@ -1383,3 +1384,24 @@ def add_collaborator(self, user, add_as_project_reviewer):
                 if self.recruit_project:
                     self.recruit_project.reviewer_users.add(user)
         self.save()
+
+    def get_users_that_reviewed_since_last_review_request(self):
```
sheenarbw, line 1388: Why is this code in this PR? 

It looks like you are using one branch for 2 features. Please remembert that we are using feature branching. 
I think you'll probably need to do some work to fix your branches. 

Look at the commit network here: https://github.com/Umuzi-org/Tilde/network. The head of one of your branches is stapled onto the tail of another one. That's really not good. 

Those should be 2 seperate branches. Each branch sould come out of develop.

If you need help understanding feature branching and I am not avaiolable I know Ryan understands this stuff well. And I'm sure Vuyi can help too.

sheenarbw (changes requested):
There are a lot of different things wrong here. 

The main thing you need to sort out before anything else is yout git branches. It's critical that we follow good pratices around git branching. Please read up on feature branching.

If you are unsure of what is going wrong with your branches please reach out to Ryan or Viro. I know that Ryan knows this stuff, and I'm pretty sure Viro knows it. 

I'll set up a session next week

backend/core/admin.py
EckardBerry: The function will still only work on a "specific team" because the user has only two options, either select all teams by clicking in the block next to "NAME" or clicking on a  specific team.  The fact that the function checks if a team is in itself marked as inactive is an added check I placed which I will now remove.  The function will not indiscriminately make team members inactive for teams it should not.

PR #208, opened 2021-10-20
#184 Add repo url to card api endpoints

### LINKED ISSUE #184: add repo_url to card api endpoints
Update the api endpoints for agile_card and agile_card summary so that there is an extra field called repo_url. If there is a repo associated with the card, then the repo url should be there.

Andy-Nkumane:

## Description:

## I solemnly swear that:


EckardBerry (changes requested):
Hello

Not bad at all, the overall functionality works.  There are some things which I think you will have to maybe change not because your work is wrong but because we work as a team and we need to understand things collectively.  Here are my remarks for now:

1. You will have to write a test/s that proves your work.  The test will probably exist
   in curriculum_tracking.tests.test_AgileCard.py  Just note that this is my opinion,
   Sheena might very well say something different but I am pretty sure she will say you
   have to have a test since we follow the TDD process.
   
2. I know what you mean by repo_url cause I read through everything, but I think that
   someone else who does not know the background might not understand, they might think
   that the function gets you a URL for some repository regardless.  Just a suggestion,
   maybe your function name should make it clear that the function returns the URL for
   a repository linked to a specific card or project?  If you do change your function name,
   remember to also change the serializer fields to the exact same name.
   
3. At the moment your function will return something only if there is something to return
   Perhaps a good idea might be that the field is still visible from the API even if there
   is no linked URL to a specific card, perhaps if no repository is linked it should return
   a message of some kind, for example:
   
   def repo_url(self):
     if self.recruit_project:
       return self.recruit_project.link_submission
     if self.repository:
       return self.repository.ssh_url
     else:
       return 'No URL linked to card'
       
    Again, this is just my opinion, most of the things I've done on my TIlde tickets ended
    up being changed so my suggestion might be useless.

backend/curriculum_tracking/models.py
```diff
@@ -1010,6 +1011,10 @@ def get_teams(self, assignees_only=False):
             user_ids.extend([user.id for user in self.reviewers.all()])
         return Team.get_teams_from_user_ids(user_ids)
 
+    @property
+    def repo_url(self):
+        return self.recruit_project.link_submission or self.repository.ssh_url
```
sheenarbw: A link submission isn't the same as a repo_url.

Also, what if there is no recruit_project on the card? It happens...

sheenarbw (changes requested):
getting there :P

EckardBerry:
Hello Andy

Sheena, Siphiwe and myself updated your repo_url function (pls have a look at it).  I've also started you off on a small test which exists in "backend/curriculum_tracking/tests/test_AgileCard.py", the class is currently called "return_repository_url_linked_to_card", you can update the test and add more tests as you see fit.

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -788,4 +788,17 @@ def test_request_review_but_no_review_done_since_time_of_review_request(self):
         # so I manually created a 'recruit_project' attribute.
         project_two.recruit_project = project_two
 
-        self.assertEqual(AgileCard.get_users_that_reviewed_since_last_review_request(project_two), [])
\ No newline at end of file
+        self.assertEqual(AgileCard.get_users_that_reviewed_since_last_review_request(project_two), [])
+
+
+class return_repository_url_linked_to_card(TestCase):
```
sheenarbw: this is a funny name. 

Are we testing `repo_url` ? Why not say `repo_url_Tests`? A useful convention to follow when testing specific functions is:

class `the_functions_actual_name_Tests`

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -788,4 +788,17 @@ def test_request_review_but_no_review_done_since_time_of_review_request(self):
...
+
+    def setUp(self):
+        self.card = factories.AgileCardFactory()
+        self.card.recruit_project.repository = git_real_factories.RepositoryFactory()
+        self.card.save()
+        self.repository = self.card.recruit_project.repository
+
+    def test_repository_link_is_returned_and_that_it_is_an_actual_repository(self):
```
sheenarbw: We need a little more here.

What happens if there is no project associated with the card? What if there is a project but no repo? 


You are basically tsting that when everything is perfect then everything will work. But stuff usually isn't perfect, we need to make sure that the code is robust.

sheenarbw (changes requested):

backend/curriculum_tracking/tests/test_AgileCard.py
EckardBerry: Extra test added now to ensure that the function repo_url returns None for a card with no linked repository.

backend/curriculum_tracking/tests/test_AgileCard.py
EckardBerry: Class name has been changed.

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -788,4 +788,26 @@ def test_request_review_but_no_review_done_since_time_of_review_request(self):
         # so I manually created a 'recruit_project' attribute.
         project_two.recruit_project = project_two
 
-        self.assertEqual(AgileCard.get_users_that_reviewed_since_last_review_request(project_two), [])
\ No newline at end of file
+        self.assertEqual(AgileCard.get_users_that_reviewed_since_last_review_request(project_two), [])
+
+
+class RepoUrlTests(TestCase):
```
sheenarbw: The convention when testing a specific function is to name the test:

f"{the_actual_function_name}_Tests"

eg look at "derive_status_from_project_Tests".  There is a function named 'derive_status_from_project'. Naming things exactly the same makes it very easy to search for related code.

Even though this convention is odd it is important to know what naming conventiuons are for.

Good code makes it's intentions clear. This code is a test for the `repo_url` function.  Also good code sticks to the conventions in play as much as possible instead of using a different convention.

If you think that this new convention is better and can make an argument for it then it might be  good to switch.

sheenarbw (changes requested):

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -788,4 +788,26 @@ def test_request_review_but_no_review_done_since_time_of_review_request(self):
...
+                            content_type=ContentItem.WORKSHOP),
+                            status=AgileCard.READY,
+                            recruit_project=None
+                            )
+
+    def test_repository_link_is_returned_and_that_it_is_an_actual_repository(self):
+        self.assertEqual(self.card.repo_url, self.repository.ssh_url)
+
+    def test_none_is_returned_for_card_not_linked_to_a_repository(self):
+        self.assertEqual(self.card_no_repo.repo_url, None)
```
sheenarbw: If `card_no_repo` is only used in this one function, why is it defined so far away in the code? 

If something is used only in one place, then define it in that place.

backend/curriculum_tracking/tests/test_AgileCard.py
```diff
@@ -788,4 +788,26 @@ def test_request_review_but_no_review_done_since_time_of_review_request(self):
...
 
-        self.assertEqual(AgileCard.get_users_that_reviewed_since_last_review_request(project_two), [])
\ No newline at end of file
+        self.assertEqual(AgileCard.get_users_that_reviewed_since_last_review_request(project_two), [])
+
+
+class RepoUrlTests(TestCase):
+
+    def setUp(self):
+        self.card = factories.AgileCardFactory()
```
sheenarbw: setup should be used for common setup. MEaning, for setting up things that are needed for multiple tests. 

you have defined tewo cards and two tests.  card 1 is only used in test 1, card 2 is only used in test 2.  So card 1 should be defined inside test 1 and card 2 should be defined inside test 2.

This is because cohesive code is good.

Also it makes your code much less leaky and makes the intentions clear.  If you put card 1 in setup then you are sayin "I intend to use this card in multiple tests". But that is not your intention.  Why should test 2 have access to card 1 if it actually doersn't need it? (principle of least privilage)

sheenarbw (changes requested):
And some more clean code things.

Please look into:
- cohesive code
- the principle of least privilage

Andy-Nkumane:
approve

sheenarbw (approved):

PR #209, opened 2021-10-20
Heading: Populate total number of cards assigned to a user per column

### LINKED ISSUE #191: Tilde board column headings: display the number of cards in the column in the heading area. eg "Backlog (53)"
Get the info from the relevent API logs in the redux store. 

eg FETCH_PERSONALLY_ASSIGNED_AGILE_CARDS_PAGE 

Note that when we do api calls to something that returns an array of values, the response has a count. 

Make sure you add up all the cards that are in the column. 

Eg the backog shows all blocked and ready cards, the review column shows cards belonging to the current user as well as cards that the current user is meant to review.

Vuyisanani:

## Description:

## Screenshots/videos

Screenshot 2021-10-20 at 08 39 39

## I solemnly swear that:


Sbonelo01:
I think Sheena already shared what is not going okay with this solution. You are displaying the number of cards that are already known by the front-end

babalwa-mbolekwa (approved):

kaleem99 (approved):

sheenarbw (changes requested):
It looks like you aren't counting the review cards. If I have 5 yellow cards and 3 blue card in a column then the total cards is 5+3. 

So we need to take lastReviewerCall into account.

Can you please write a test that calls boardFromCards. 
The test needs to pass some api call info to the function and expect the correct column call counts for the different columns.

Sbonelo01 (commented):
Hello @Vuyisanani, so far the code looks okay. I will assume you are still working on the tests

kaleem99 (approved):
looks good.

frontend/src/components/regions/AgileBoard/AgileBoard.test.js
```diff
@@ -0,0 +1,57 @@
+import  { boardFromCards }  from "../AgileBoard";
+import { cards, latestCalls} from "./mockedData";
+
+test("boardFromCards function returns correct total length 5 of the array", () => {
+   let boardFromCard  = boardFromCards({ cards, latestCalls });
```
Sbonelo01, line 5: I know that these are just tests. Is there a specific reason they are declared with let instead of const? `const` seems more fiting.

Sbonelo01 (commented):
Everything else looks good ...

frontend/src/components/regions/AgileBoard/AgileBoard.test.js
sheenarbw: const should be used here. Remember that we hold the learners to a high standard with this sort of thing.

sheenarbw (approved):
One let versus const issue, it would be good to use the right thing going forward. But I'm keen to get this merged so we can move on

frontend/src/components/regions/AgileBoard/AgileBoard.test.js
Vuyisanani: I will change to const

PR #215, opened 2021-10-27
Has card been reviewed 

### LINKED ISSUE #144: Add an icon to theAgileCard component that shows a user if they have reviewed the card since the last review request
NEEDS #143 

If we are looking at a user's board then we will need to see checkboxes to show whether that user has reviewed the card or not.

The checkboxes should only be displayed if the card is in the Review column.

Eg if we are looking at the board of a user with id=555:

If there is a card in the review column, and the users who reviewed the card since the last review request = [123,456,7], then display an unchecked checkbox

If there is a card in the review column, and the users who reviewed the card since the last review request = [123,555,456,7], then display achecked checkbox

## NB: 

Please make sure @Vuyisanani  is happy with the design before you start doing things.
Please update the presentation layer of the AgileCard component so that this functionality will show up in storybook.
You'll notice that the agile card story uses some json fixtures. Those fixtures are out of date - there is a new field available on the agile card api endpoint. Eckard added it in #143 so you can ask him about it if you get confused. 

Babalwa has done similar tasks. She will also be able to give advice here.

Sbonelo01:
Related issues: [https://github.com/Umuzi-org/Tilde/issues/144]

## Description: 
**This small PR seeks to get confirmation whether the icon is placed in the correct place or not, need feedback from  @Vuyisanani as per the issue details. The rest of the code is still under construction.**

## Screenshots/videos

Screenshot from 2021-10-27 08-30-10

## I solemnly swear that:


frontend/src/components/regions/AgileBoard/AgileCard/Presentation.js
```diff
@@ -148,7 +150,10 @@ export default ({
           {card.contentTypeNice} {card.projectSubmissionTypeNice}
         </Typography>
         <Typography variant="caption"> [card id:{card.id}]</Typography>
-
+        <Checkbox
+          checked={true}
+          color="primary"
```
Vuyisanani: Instead of introducing a new colour to the Card, can you rather consider using this color: #3f51b5;

Vuyisanani:
The check box,  position and size is fine, but just the colour,  make it be like the one below, #3f51b5;
Screenshot 2021-10-27 at 23 52 02

Vuyisanani (approved):

frontend/src/components/regions/AgileBoard/AgileCard/Presentation.js
Vuyisanani: I see this is the same colour when pulled this on my branch :)

babalwa-mbolekwa (changes requested):
Hey Sbonelo. The UI looks great and all tests pass, but the box is not ticked when the viewed user's ID is part of the IDs in the `usersThatReviewedSinceLastReviewRequest` list from Storybook. could you please fix that part. 
- The card titles are also repeated when the card is not in the review column, could you please fix them too:
titles

Sbonelo01:
Hey @Babalwa01, Thanks for the review; It does change.

This is the card when the viewed users id is not in the reviewers list:

Screenshot from 2021-11-10 18-34-07

and this is the card when the viewed user id is found in the card:

Screenshot from 2021-11-10 18-35-25

babalwa-mbolekwa (approved):
Well done Sbonelo!

frontend/src/components/regions/AgileBoard/AgileCard/AgileCard.test.js
```diff
@@ -52,3 +53,35 @@ test("getAgeString function returns correct string when age is less than a minut
...
+    status: "IR",
+    usersThatReviewedSinceLastReviewRequest: [777, 555, 3332],
+  }
+  const viewedUser = {
+    id: 28,
+  }
+  expect(showCheckedBox({ viewedUser, card })).toBe(false)
+})
+
+test("showCheckedBox should return false if the card is not in the review column", () => {
```
sheenarbw: The checkbox should be shown if the card is in the review column. This makes it look like you only want to show the checkbox if the user has already reviewed the card. 

Names are important. 

And also, how do you know we wont use other checkboxes later on? And maybe we'll decide to use a different way to visualise this stuff later on instreead of a checkbox.  Why does your function even need to know that it has anything to do with a checkbox? 

How about calling the function: `userReviewedSinceLastReviewRequest`, it would return true oif the user did review the card since the last review request. See how the function name fully describes what the function actually does?

frontend/src/stories/fixtures/agileCard.json
```diff
@@ -1,61 +1,62 @@
 {
-    "id": 107015,
-    "contentItem": 550,
-    "contentItemUrl": "http://syllabus.africacode.net/projects/memory-game/part-2/",
-    "status": "RF",
-    "recruitProject": 5738,
+    "id": 130508,
```
sheenarbw: Can you please avoid changing fixtures unless you need to? There are all sorts of changes here that are irrelevent to your code.  Other people use these fixtures, did you make sure you didn't break any stories?

frontend/src/components/regions/AgileBoard/AgileCard/utils.js
```diff
@@ -0,0 +1,14 @@
+export function checkIfCardIsInReviewColumn({ card }) {
+  return card.status === "IR";
+}
+
+const checkIfUserIdExistsInReviewers = ({ viewedUser, card }) => {
+  return card.usersThatReviewedSinceLastReviewRequest.includes(viewedUser.id);
+};
+
+export function showCheckedBox({ viewedUser, card }) {
```
sheenarbw: Is this unction supposed to decide if the checkbox should be shown or not? OR is it supposed to decide if the checkbox should be checked or not? 

It should do only one of these things. Not bothe of these things.

sheenarbw (changes requested):
It looks like it works but the code follows a few bad practices. 

Good names and single purpose functions make code a lot clearer. 

AgileCard Presentation directly calls checkIfCardIsInReviewColumn and showCheckedBox. 
Then ShowCheckedBox calls checkIfCardIsInReviewColumn and checkIfUserIdExistsInReviewers 

That is a complicated way to solve a simple problem. 

AgileCard Presentation  should call one function to know if it should draw the checkbox or not and another function to decide if the box should be checked. These 2 functions shouldn't know about each other. You only need 2 functions and they are both very simple and independent. 

Please work on making your code as simple as possible. Otherwise when you get a complicated problem to solve it's going to be really a lot of spagetti.

sheenarbw (approved):

PR #221, opened 2021-10-28
Assignment exporter script

### LINKED ISSUE #142: Assignment exporter script
We need a managment script that we can use to give Liezel a csv with people's writing assignments. We can make it a bit more versatile though.

The script should live inside curiculum_tracking/managment/commands

Just poke around in there to get an idea of how things work.

The script needs inputs:
- team name
- content item title 

The script should generate a csv with the following columns:
learner_email_address, link_to_project

have some pseudocode: 
```
For each active user in the team: 
   Get the user's ProjectProgress instance with the matching title
   get the url of the repo or if there is a link submission then get that url
   add a row to the csv output
```

Generally when generating csv files and that kind of thing, we just put them inside the gitignore folder so git ignores them.

Testing: 
The test coverage on this project is not as extensive as it could be.  Try to seperate your code out into easily testable functions and test them. At some point we are likely to refactor this code and turn it into an api. 

We are making it a managment command first because Liez needs the data soon and making it work as an api will take longer.

Take a look at how FactoryBoy is used in the tests

TheAbstract:
Related issues: [#142 ]

## Description:

## I solemnly swear that:


backend/curriculum_tracking/tests/test_command_export_project_urls.py
```diff
@@ -0,0 +1,19 @@
...
+
+class TestCommandExportProject(TestCase):
+
+    def test_project_not_submitted(self):
+        self.team = TeamFactory()
+        self.user = UserFactory()
+        self.team.users.add(self.user)
+
+        project = ProjectContentItemFactory()
+        result = get_email_and_url(self.team)
```
EckardBerry: So I see that result correctly calls 'get_email_and_url' which then in turn returns a list as it should with the email of the active user but 'None' for the link submitted.  So I am guessing that is where your blocker is, that you need it to return a liink?

EckardBerry (commented):
I created a stupid little test just to show you how to get a team with a user but also a repository, your function 'get_email_and_url' will still only get the user email and not the url but that is because it is looking for link_submission or git_url and not repository.  Please see my test below, hope it helps:

    def test_project_with_a_link_submission(self):
        from curriculum_tracking.tests import factories
        from git_real.tests import factories as git_real_factories

        self.team = TeamFactory()
        self.card = factories.AgileCardFactory()
        self.card.recruit_project.repository = git_real_factories.RepositoryFactory()
        self.card.save()

        repository = self.card.recruit_project.repository  # This will return <Repository: https://something.com/like_this4>
        user = self.card.recruit_project.repository.user  # This will return <User: foo.5@example.com>
        self.team.users.add(user)
        res = get_email_and_url(self.team)

backend/curriculum_tracking/management/commands/export_project_urls.py
```diff
@@ -0,0 +1,29 @@
...
+from curriculum_tracking.models import RecruitProject
+import csv
+
+
+def get_email_and_url(team):
+    results = []
+    for user in team.active_users:
+        project = RecruitProject.objects.filter(recruit_users__in=[user]).first()
+        results.append([user.email, project and (project.link_submission or project.git_url)])
+    return results[0]
```
sheenarbw: why are you only returning the first one?

backend/curriculum_tracking/management/commands/export_project_urls.py
```diff
@@ -0,0 +1,29 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument("team_name", type=str)
+        parser.add_argument("content_item_title", type=str)
+
+    def handle(self, *args, **options):
+        print(args, options)
+        team = Team.objects.get(name=options["team_name"])
+        results = get_email_and_url(team)
+
+        with open('gitignore/test.csv', 'w') as f:
```
sheenarbw: can you choose a better file name?

backend/curriculum_tracking/management/commands/export_project_urls.py
```diff
@@ -0,0 +1,29 @@
...
+            writer = csv.writer(f)
+            writer.writerow(['user_eamil', 'url'])
```
sheenarbw: typo `email`

backend/curriculum_tracking/management/commands/export_writing_assignments.py
```diff
@@ -0,0 +1,18 @@
+from django.core.management.base import BaseCommand
+from core.models import Team
+from curriculum_tracking.models import AgileCard, ContentItem
+import pandas as pd
+
+class Command(BaseCommand):
```
sheenarbw: We only need one command. Why are there 2?

backend/curriculum_tracking/tests/test_command_export_project_urls.py
```diff
@@ -0,0 +1,19 @@
+from core.tests.factories import TeamFactory, UserFactory
+from curriculum_tracking.tests.factories import ProjectContentItemFactory
+from django.test import TestCase
+from curriculum_tracking import models
+from curriculum_tracking.management.commands.export_project_urls import get_email_and_url
+
+class TestCommandExportProject(TestCase):
+
+    def test_project_not_submitted(self):
+        self.team = TeamFactory()
```
sheenarbw: Why are you using self? Are you using `team` outside this method?

Try not to let information leak out a function unless you need it to.

backend/curriculum_tracking/tests/test_command_export_project_urls.py
sheenarbw: It would also be good to test multiple rows returned.

sheenarbw (changes requested):

backend/curriculum_tracking/tests/test_command_export_project_urls.py
TheAbstract: @EckardBerry for this test it is meant to return `None`, for when someone hasn't submitted a link.

backend/curriculum_tracking/management/commands/export_project_urls.py
TheAbstract: The returned object is a list of lists, where each inner list is a row that we'll write to the csv. So I'm getting just the actual lists, rather than the whole thing.

sheenarbw (changes requested):
There's an extra file: backend/curriculum_tracking/management/commands/export_writing_assignments.py  

Can you please delete it?

TheAbstract:
> There's an extra file: backend/curriculum_tracking/management/commands/export_writing_assignments.py
> 
> Can you please delete it?

@sheenarbw that file was already in `develop` here https://github.com/Umuzi-org/Tilde/blob/develop/backend/curriculum_tracking/management/commands/export_writing_assignments.py but I've already deleted it on this branch. When this gets merged the one in `develop` will get deleted.

EckardBerry (changes requested):
Hello

Everything looks and works like it should, just one thing, the import of 'get_projects' (line 4) of export_project_urls.py is not used.  I ran your tests and then tried to break them but they worked.

EckardBerry (approved):

sheenarbw (changes requested):
Can you please delete thios file:
backend/curriculum_tracking/management/commands/export_writing_assignments.py 

That file should not be mentioned in the PR.  If you look at your Files Changed for this pr there is an extra file. It should be removed. Otherwise someone else has to tidy up after you and that's not the best. It's better just to merge what you need merged.

sheenarbw (approved):

PR #222, opened 2021-10-28
Moved the buttons below the navBar

Vuyisanani:

## Description: Moved the buttons below the navBar

What are you up to? Fill us in :) This is still ongoing, just have questions regarding changing the two buttons style to use Tabs, will that be for only these 2 buttons or should we consider for all other buttons?

## Screenshots/videos

Screenshot 2021-10-28 at 23 54 56

## I solemnly swear that:

### DIFF
--- a/frontend/src/components/widgets/BaseButtonLink.js
+++ b/frontend/src/components/widgets/BaseButtonLink.js
@@ -1,5 +1,6 @@
 import React from "react";
 import { Link } from "react-router-dom";
+import Tab from "@material/react-tab";
 
 import {
     Button,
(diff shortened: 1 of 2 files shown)


frontend/src/components/widgets/BaseButtonLink.js
```diff
@@ -1,5 +1,6 @@
 import React from "react";
 import { Link } from "react-router-dom";
+import Tab from "@material/react-tab";
```
sheenarbw, line 3: Please don't edit this. We use Buttons all over the place.

The issue says: Dont use buttons in the app bar, use tabs instead in the app bar.

The issue does not say: Never use buttons anywhere ever again. We are jkust changing one little part of the application so it works differently.

sheenarbw (changes requested):
Looks like this issue is not understood. 

And there are merge conflicts. It might be good to just start over

Vuyisanani:
This is no longer relevant, new PR on a different branch been created

PR #238, opened 2021-11-08
Api endpoint exposes review trust/166

### LINKED ISSUE #166: Create an api endpoint that exposes ReviewTrust instaces
Start off by writing a test. You can look at `TopicProgressViewsetTests`, it's a nice simple one.
Allow filtering by user id

You'll also need to make changes to: backend/urls.py

Make a PR once you have the tests passing so you can get some feedback.

Next up we need to work on making sure the permissions are correct:

- Only GET requests are ok. This is read only.
- admin/super users can read everything
- other users can always see their own stuff - s if I am user 1 and I am filtering for user 1 then that's ok, but I can't look at user 2's stuff.
- If I have VIEW permission over a user then I can see their stuff

EckardBerry:

## Description:
1. Made some changes to backend/curriculum_tracking/api_views.py where I added the class based view of 
    ReviewTrustsViewset (starting from line 887)

2. Made some changes to backend/backend/urls.py where I  included the registration of this api endpoint url, see lines   
    122-126

3. Made some changes to backend/curriculum_tracking/serializers.py where I added a new serializer for the api endpoint,
    see lines 614-627

4. Made some changes to backend/curriculum_tracking/tests/test_api_views.py where I added tests for this api endpoint,
    see lines 377-416

## I solemnly swear that:


backend/git_real/tests/factories.py
```diff
@@ -73,16 +57,3 @@ class Meta:
     title = "title"
     body = "body"
     state = PullRequest.OPEN
-
-
-class PullRequestReviewFactory(DjangoModelFactory):
-    class Meta:
-        model = PullRequestReview
```
sheenarbw: This is a little strange. Why are you removing this code? 

Always remember to do a self-review

sheenarbw (changes requested):
You are deleting the PullRequestReviewFactory class. I'm not sure what happened there

backend/git_real/tests/factories.py
EckardBerry: I think the issue might be that I did not pull the changes in develop before I started on this ticket.  I have now merged develop into my branch and resolved the merge conflicts, PullRequestReviewFactory now exists for branch (api_endpoint_exposes_review_trust/166) as well.

backend/curriculum_tracking/serializers.py
```diff
@@ -611,3 +611,19 @@ def get_oldest_card_in_review_time(self, instance):
...
         return self._get_review_agile_cards(instance).count()
+
+
+class ReviewTrustSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.ReviewTrust
+        fields = [
+            "id",
+            "content_item",
+            "user",
```
sheenarbw, line 639: flavours are also relevent. Check what other endpoints expose flavours, you can use the same pattern.

Eg: A person might be trusted for the Animals project, but only in Python (not in Javascript or Java)

sheenarbw (changes requested):
Looking very good. You are testing your code in a much better way, it's nice and clear to read. Nice work!

The only thing left is a missing field. Flavours are meaningful and should be exposed. Let me know if you need a description of what a flavour is and how it is used, I dont mind explaining.

Andy-Nkumane (approved):

backend/curriculum_tracking/serializers.py
```diff
@@ -611,3 +611,21 @@ def get_oldest_card_in_review_time(self, instance):
...
+    class Meta:
+        model = models.ReviewTrust
+        fields = [
+            "id",
+            "content_item",
+            "flavours",
+            "user",
+        ]
+
+    content_item = serializers.SerializerMethodField('get_name_not_id_of_content_item')
```
sheenarbw: wouldn't it make more sense with slightly different wording:

```
content_item_title = serializers.SerializerMethodField('get_content_item_title')
```

backend/curriculum_tracking/serializers.py
```diff
@@ -611,3 +611,21 @@ def get_oldest_card_in_review_time(self, instance):
...
+    flavours = serializers.CharField(help_text="comma separated list of flavours")
+
+    def get_name_not_id_of_content_item(self, review_trust: object):
+        content_item_name = review_trust
+        return str(content_item_name)
```
sheenarbw: This seems a little bit strange. Firstly, there is an extra line of code:
```
def get_name_not_id_of_content_item(self, review_trust: object):
        content_item_name = review_trust
        return str(content_item_name) 
```

This does the same thing:

```
def get_name_not_id_of_content_item(self, review_trust: object):
        return str(review_trust) 
```

But then you are relying on the str method of the review trust, so if anyone wants to see how this actually works they have to go on a treasure hunt. Also, what if someone changes how ReviewTrust.__str__ works? Then your api endpoint would randomly change and then the frontend would randomly act differently. 

It's best to be explicit when dealing with data.

```
def get_content_item_title(self, review_trust):
    return review_trust.content_item.title 
```

sheenarbw (changes requested):
Nearly there

Andy-Nkumane (approved):

backend/curriculum_tracking/serializers.py
```diff
@@ -611,3 +611,20 @@ def get_oldest_card_in_review_time(self, instance):
...
     def get_total_cards_in_review(self, instance):
         return self._get_review_agile_cards(instance).count()
+
+
+class ReviewTrustSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.ReviewTrust
+        fields = [
+            "id",
+            "content_item",
```
sheenarbw, line 636: is this the content item or the content item title?  It's a little misleading.

It might be good to call a title a title.

Usually foreign key relationships are exposed as integers.

This would probably be the best version of the code:

```
fields = [
            "id",
            "content_item", # the default behaviour here is nice and consistent with the rest of the application.
            "content_item_title", # this is the title of a content item - see how descriptive the name is?
             "flavours",
            "user",
]

...

    content_item_title = serializers.SerializerMethodField('get_content_item_title')


```

sheenarbw (changes requested):

TheAbstract (changes requested):
@EckardBerry code LGTM but we have conflicts to resolve here 🙈 .

EckardBerry:
Merge conflict resolved, sending back for a review.

Andy-Nkumane (approved):

TheAbstract (approved):

backend/curriculum_tracking/api_views.py
```diff
@@ -902,3 +902,22 @@ class BurnDownSnapShotViewset(viewsets.ModelViewSet):
...
+        & (
+            core_permissions.IsCurrentUserInSpecificFilter("user")
+            | core_permissions.HasObjectPermission(
+                permissions=Team.PERMISSION_VIEW,
+                get_objects=_get_teams_from_user_filter("user"),
+            )
+        )
+    ]
+
+    queryset = models.ReviewTrust.objects.all()
```
sheenarbw: one small change. Can you always add an order_by on the queryset? Otherwise weird random things happen.

if there is no obvious field to order by then just do `order_by('pk')`

backend/curriculum_tracking/serializers.py
```diff
@@ -625,4 +625,21 @@ class Meta:
...
+            "content_item_title",
+            "flavours",
+            "user",
+        ]
+
+    content_item_title = serializers.SerializerMethodField('get_content_item_title')
+    flavours = serializers.CharField(help_text="comma separated list of flavours")
```
sheenarbw: This implies that someone will be filling in a form to add some flavours.

Can you rather look at how AgileCardSerialiser exposes flavours? 

I think there is some flavour technical debt that may have mislead you a bit

sheenarbw (changes requested):
Nearly there. I noticed a few more little things

backend/curriculum_tracking/serializers.py
EckardBerry: The AgileCardSerializer just has "flavour_names" in the fields list.  I changed things so that it looks like the AgileCardSerializer but I am not sure if the change is what you wanted?

EckardBerry: Individually my tests pass but when I try to run the class inside which my tests exist I get the following:

AssertionError: False is not true : flavour_names = []. The value must be truthy

I have a hack to get around the problem but I am fairly sure it won't be allowed.  What should I do?

TheAbstract (approved):

sheenarbw (approved):

PR #251, opened 2021-11-16
Tilde weekly reviews bug

### LINKED ISSUE #225: Bug tilde_project_reviews_done_in_past_seven_days returns incorrect value
curriculum_tracking/serializers.py 

get_tilde_reviews_done_last_7_days has bad logic. 

```
if tilde_project_reviews_done_in_past_seven_days == None:
            return tilde_topic_reviews_done_in_past_seven_days
        else:
            return tilde_project_reviews_done_in_past_seven_days
```

tilde_project_reviews_done_in_past_seven_days will never be None. It will always be a number, that umber might be 0. 

And if I did 5 topic reviews and 3 project reviews then my total reviews is 8. 

get_total_tilde_reviews_done has the same problem. 

## The best wqay to fix bugs is TDD 

Step 1: write a test that will fail because of the bug 
Step 2: Fix the code so that the test passes

TheAbstract:
Related issues: #225 

## Description:

In **backend/curriculum_tracking/serializer.py**:
- Fixed logic for `get_tilde_reviews_done_last_7_days` and `get_total_tilde_reviews_done`.

In **backend/curriculum_tracking/tests/test_UserStatsPerWeekSerializer.py**:
- Added tests and cases to make sure functions count both project and topic reviews. For example if someone reviews a project and a topic within a week their reviews done should be 2 not 1.
- Replaced the rather long `test_get_tilde_reviews_done_last_7_days` with smaller specific tests; 
    1. `test_no_tilde_reviews_done_in_last_7_days`
    2. `test_multiple_tilde_reviews_done_in_last_7_days`
- Added tests for `UserStatsPerWeekSerializer.get_total_tilde_reviews_done`;
    1. `test_no_tilde_reviews_done`
    2. `test_multiple_tilde_reviews_done`

## I solemnly swear that:

### DIFF
--- a/backend/curriculum_tracking/serializers.py
+++ b/backend/curriculum_tracking/serializers.py
@@ -457,10 +457,6 @@ class UserStatsPerWeekSerializer(serializers.ModelSerializer):
         )
 
         return project_reviews_done_to_date + topic_reviews_done_to_date
-        # if tilde_project_reviews_done_to_date == None:
-        #     return tilde_topic_reviews_done_to_date
-        # else:
-        #     return tilde_project_reviews_done_to_date
 
     def get_tilde_cards_reviewed_in_last_7_days(self, user):
         tilde_project_reviews_done_in_past_seven_days = (
--- a/backend/curriculum_tracking/serializers.py
+++ b/backend/curriculum_tracking/serializers.py
@@ -399,7 +399,6 @@ class UserStatsPerWeekSerializer(serializers.ModelSerializer):
     )
 
     def get_cards_in_complete_column_as_assignee(self, user):
-
         cards_in_completed_column_as_assignee_amount = models.AgileCard.objects.filter(
             status=models.AgileCard.COMPLETE, assignees=user.id
         ).count()
@@ -407,7 +406,6 @@ class UserStatsPerWeekSerializer(serializers.ModelSerializer):
         return cards_in_completed_column_as_assignee_amount
 
     def get_cards_in_review_column_as_assignee(self, user):
-
         cards_in_review_column_as_assignee_amount = models.AgileCard.objects.filter(
             status=models.AgileCard.IN_REVIEW, assignees=user.id
         ).count()
@@ -415,7 +413,6 @@ class UserStatsPerWeekSerializer(serializers.ModelSerializer):
         return cards_in_review_column_as_assignee_amount
 
     def get_cards_in_review_feedback_column_as_assignee(self, user):
-
         cards_in_review_feedback_column_as_assignee_amount = (
             models.AgileCard.objects.filter(
                 status=models.AgileCard.REVIEW_FEEDBACK, assignees=user.id
@@ -425,7 +422,6 @@ class UserStatsPerWeekSerializer(serializers.ModelSerializer):
         return cards_in_review_feedback_column_as_assignee_amount
 
     def get_cards_in_progress_column_as_assignee(self, user):
-
         cards_in_progress_column_as_assignee = models.AgileCard.objects.filter(
             status=models.AgileCard.IN_PROGRESS, assignees=user.id
         ).count()
@@ -433,7 +429,6 @@ class UserStatsPerWeekSerializer(serializers.ModelSerializer):
         return cards_in_progress_column_as_assignee
 
     def get_cards_completed_last_7_days_as_assignee(self, user):
-
         cards_completed_past_seven_days = models.AgileCard.objects.filter(
             status=models.AgileCard.COMPLETE,
             assignees=user.id,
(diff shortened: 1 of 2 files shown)


backend/curriculum_tracking/serializers.py
```diff
@@ -452,24 +446,19 @@ def get_cards_started_last_7_days_as_assignee(self, user):
         return cards_started_past_seven_days
 
     def get_total_tilde_reviews_done(self, user):
-
-        tilde_project_reviews_done_to_date = (
+        project_reviews_done_to_date = (
             models.RecruitProjectReview.objects.filter(reviewer_user_id=user.id)
             .all()
             .count()
         )
```
EckardBerry, line 454: Do you need .all().count().  I am pretty sure just .count() would work

backend/curriculum_tracking/serializers.py
```diff
@@ -496,8 +485,7 @@ def get_tilde_cards_reviewed_in_last_7_days(self, user):
...
```
EckardBerry, line 485: These two functions "get_tilde_cards_reviewed_in_last_7_days" and :"get_tilde_reviews_done_last_7_days" seem to me to be doing roughly the same thing.  They both have tilde project reviews and tilde topic reviews and they both filter according to user.id and a timestamp of nothing less than 7 days ago from current date.  Hence, the two querysets are roughly the same.

backend/curriculum_tracking/serializers.py
EckardBerry, line 453: Do you need .all().count(). I am pretty sure just .count() would work

backend/curriculum_tracking/tests/test_UserStatsPerWeekSerializer.py
```diff
@@ -215,13 +211,58 @@ def test_get_tilde_cards_reviewed_in_last_7_days(self):
...
 
         # add another review, now the number of cards reviewed is 2
-
         review4 = RecruitProjectReviewFactory(
             timestamp=self.yesterday, reviewer_user=user
         )
         review4.timestamp = self.yesterday
         review4.save()
         AgileCardFactory(recruit_project=review4.recruit_project)
+        count = self.stats_serializer.get_tilde_cards_reviewed_in_last_7_days(user)
```
EckardBerry, line 227: You added count = self.stats_serializer.get_tilde_cards_reviewed_in_last_7_days(user), above it (line 210) the same assertion takes place and then later on again.  I think that either there should be one assertion or this test should be divided into three separate tests.  But seeing that they all test the same thing just at different stages I think the assertions are duplicated.

EckardBerry (changes requested):
Hello

Some good changes made, however, I think that some of the changes make things a bit cluttered and I think that at least one of the functions is duplicated in it's functionality.  The one test also "test_get_tilde_cards_reviewed_in_last_7_days" is a bit lengthy after the changes, I think you should rather create new tests.  Anyway, hear what Sheena says, this is just my opinion.

backend/curriculum_tracking/tests/test_UserStatsPerWeekSerializer.py
EckardBerry: Forgot to mention one thing that helps a lot:

When you make changes, in the description of the PR, it is a good idea to list
   the file names where you made changes and also the lines in which you made 
   changes, I saw that on one of Sheena's PR's and it makes reviewing a lot easier
   and quicker.  For instance, in your description you say you made changes to
   get_tilde_reviews_done_last_7_days and get_total_tilde_reviews_done but actually
   there are a lot of other small changes in other functions as well.

TheAbstract: I'll make the description more specific, but I'm not too sure about the specifying lines part. Isn't that what the diff on GitHub is for?  
Also, every file that has any changes is automatically shown in **Files changed**, why duplicate the process manually by mentioning files you made changes in?

backend/curriculum_tracking/serializers.py
```diff
@@ -383,9 +383,6 @@ class Meta:
     tilde_reviews_done_last_7_days = serializers.SerializerMethodField(
         "get_tilde_reviews_done_last_7_days"
     )
-    tilde_cards_reviewed_in_last_7_days = serializers.SerializerMethodField(
```
sheenarbw: Reviews done versus cards reviewed are 2 seperate things.

Maybe I reviewed the same card 3 times. So I reviewed 1 card only, but did 3 reviews

sheenarbw (changes requested):
That thing you removed was actually there on purpose.

Number of reviews done != number of cards reviewed

backend/curriculum_tracking/serializers.py
TheAbstract: @EckardBerry there was no duplication like we thought, I restored the original `get_tilde_cards_reviewed_in_last_7_days` and its test.

EckardBerry (approved):

backend/curriculum_tracking/serializers.py
sheenarbw: No. Cards reviewed != Reviews done.

I can do 12 reviews on one card.

sheenarbw (approved):

PR #271, opened 2021-11-25
Expose git_real/models/Push as an api

### LINKED ISSUE #149: expose git_real/models/Push as an api. Allow filter by repository
Permissions: Anyone with any view access to a user should be able to see their push events
Exposed fields: For now all the usernames, the timestamps, id, repo id, head commit url

TheAbstract:
Related issues: #149 

## Description:

1. Added `PushSerializer` in git_real/serializers.py
2. Added `PushViewSet` in git_real/api_views.py
3. Registered in backend/backend/urls.py

Note: The following functions/classes where added to preserve the direction of imports inline with this feedback from Sheena http://rocketchat.umuzi.org/group/tilde-dev-team?msg=EN477th54cFa9Kq3y

4. In git_real/serializers.py:
- `_get_teams_from_repository_instance`
- `_get_teams_from_repository_filter`

## I solemnly swear that:


TheAbstract:
@sheenarbw first time working on the **git_real** application so I'm actually not sure what should be tested for this?

backend/git_real/permissions.py
```diff
@@ -22,6 +22,25 @@ def has_permission(self, request, view):
         correct = recieved_digest == "sha1=" + digest
         if not correct:
             return False
-            # breakpoint()
 
         return correct
+
+def user_can_see_repository(repo, user):
```
sheenarbw: This is only used in one class, right? 

Why not put it in that class? Or are you intending it to be used elsewhere as well?

backend/git_real/permissions.py
```diff
@@ -22,6 +22,25 @@ def has_permission(self, request, view):
...
+        if user in project.recruit_users.all():
+            return True
+        if user in project.reviewer_users.all():
+            return True
+    return False
+
+class IsFilteredByRepoAttachedToProjectICanSee(BasePermission):
+
+    def has_permission(self, request, view):
+        if request.method != "GET":
```
sheenarbw: The name of this filter class doesn't imply that the request must be a GET. Why not just check that the user has permission to see the repo? You cal check the request method elsewhere. 

In fact you do check the request method elsewhere. If the ActionIs(list) then it's definately a GET request.

sheenarbw (changes requested):
This is good but there is some odd things happening around cohesive code and highly coupled code.

Also, I can't see any tests...

ry-oc:
@TheAbstract fyi: merge conflicts

EckardBerry:
Hello

Please checkout backend/git_real/tests/test_api_views.py, you will see I started a test there for you which basically creates a repository and a push event, it then goes to the api end point to see if the push event can be retrieved from the repository details.  I left some comments there for you also.

EckardBerry (changes requested):
Pls see comment I made about the test I started.

TheAbstract:
@EckardBerry about https://github.com/Umuzi-org/Tilde/pull/271#issuecomment-1013018451 like I mentioned on RC thats basically the idea I initially had but I don't think it's correct. I'll set something up for us to chat.

backend/git_real/tests/test_api_views.py
```diff
@@ -1,13 +1,68 @@
...
+        self.assertEqual(response.status_code, 200)
+
+    @mock.patch.object(IsWebhookSignatureOk, "has_permission")
+    def test_push_event_without_permission(self, has_permission):
+        has_permission.return_value = False
+        body, headers = get_body_and_headers("push")
+        url = reverse(views.github_webhook)
+        response = self.client.post(url, format="json", data=body, extra=headers)
+        self.assertEqual(Push.objects.count(), 0)
+        self.assertEqual(response.status_code, 401)
```
EckardBerry: I think the test "test_push_event_without_permission" should assert
   that the response code from the django server is 403, ergo, the request was
   understood but permission was not granted.  What is the difference?
   401 => unauthorised response for missing or bad authentication.
   403 => Forbidden response, the user is authenticated but is not authorized to
   perform the request.

EckardBerry: I still think you should have a test where you assert that the correct Push
   event gets retrieved from the view itself.  It is kind of like a test where
   you act like you are a student who makes use of the api via a curl command or
   something in order to check/retrieve all your push events.  But hear what Sheena says?

EckardBerry (changes requested):

TheAbstract:
@sheenarbw some clarify on what Eckard raised above:
1. About this https://github.com/Umuzi-org/Tilde/pull/271#discussion_r797338045 I thought the required permissions is that if a user isn’t staff or super user or the assignee, we should be getting a 401 not 403. Also a bad request would be 400 not 401, I checked the docs https://www.django-rest-framework.org/api-guide/status-codes/

2. On this https://github.com/Umuzi-org/Tilde/pull/271#discussion_r797338542 I'm not sure I get what he means.

backend/git_real/tests/test_api_views.py
```diff
@@ -1,13 +1,68 @@
...
+        head_commit_url = head_commit["url"]
+        commit_timestamp = dateutil.parser.isoparse(head_commit["timestamp"])
+        ref = body["ref"]
+        pushed_at_time = github_timestamp_int_to_tz_aware_datetime(
+            int(repository["pushed_at"])
+        )
+
+        self.client.post(url, format="json", data=body, extra=headers)
+        self.assertEqual(Push.objects.count(), 1)
+        push = Push.objects.first()
```
sheenarbw: This seems very complicated. Can't you use a factory to make a Push instance? Just `push = PushFactory()`

Here you are testing that the webhook works. We already tested the webhook elsewhere.  And it's a totally different api endpoint.

backend/git_real/tests/test_api_views.py
sheenarbw: I would accept a 401

backend/git_real/tests/test_api_views.py
```diff
@@ -1,13 +1,68 @@
...
 class RepositoryViewsetTests(APITestCase, APITestCaseMixin):
     LIST_URL_NAME = "repository-list"
     SUPPRESS_TEST_POST_TO_CREATE = True
 
     def verbose_instance_factory(self):
-        return factories.RepositoryFactory(archived=True)
+        return RepositoryFactory(archived=True)
+
+
+class PushEventTests(APITestCase):
```
sheenarbw: It's important to also check that the view actually exposes data. The api endpoint isn't just concerned with permissions, it neds to return actual data.

Please make use of the api mixins to test that. Just use SUPPRESS_POST_TO_CREATE

backend/git_real/tests/test_api_views.py
```diff
@@ -1,13 +1,68 @@
...
+        self.assertEqual(push.message, message)
+        self.assertEqual(push.commit_timestamp, commit_timestamp)
+        self.assertEqual(push.pushed_at_time, pushed_at_time)
+        self.assertEqual(push.ref, ref)
+
+        response = self.client.post(url, format="json", data=body, extra=headers)
+        self.assertEqual(response.status_code, 200)
+
+    @mock.patch.object(IsWebhookSignatureOk, "has_permission")
+    def test_push_event_without_permission(self, has_permission):
```
sheenarbw: This tests that permission can be denied. But it would also be useful to show what kinds of users actually have access.

Eg does the reviewer have access? How about the assignee?

sheenarbw (changes requested):
The tests do need a bit more work.

backend/git_real/api_views.py
```diff
@@ -1 +1,43 @@
...
+
+    return _get_teams_from_repository_instance(repo)
+
+class PushViewSet(viewsets.ModelViewSet):
+    serializer_class = PushSerializer
+    queryset = Push.objects.order_by("-pushed_at_time")
+    filter_backends = [DjangoFilterBackend]
+    filterset_fields = ["repository"]
+
+    permission_classes = [
```
sheenarbw, line 37: Here's your problem.

If you have a 'detail' type action (such as the `retrieve` action used for GETting an object) then `view.get_object()` works. But if you are using a list view then `view.get_object` wont work.

There also seems to be some misunderstanding about what the object would even be. 

Eg if we had a detail view accessible at:  `/api/push/{pk}` then this would refer to a specific Push instance. 
If we applied your `IsCardAssignee` permission to that then here is what happens:

```
class IsCardAssignee(BasePermission):
    def has_permission(self, request, view):
        user = request.user
        card = view.get_object() # This is not a card, it is a Push instance
        return user in card.assignees.all()  # this will error because Push has no attribute named assignees
```

What you need is a permission that doesn't rely on get_object at all.

It should do something like this:

```
class Something(...):
    repo_id = get the id from the URL, it's a GET parameter 
    get the project that is associated with the repo
    return True if the current user an assignee or reviewer of the project
```

sheenarbw: There is another issue to think about:

The git_real app should not know about the curriculum tracking app. Reason being: We detest spaghetti.  We should always try to keep all our imports going in the same direction. Curriculum tracking can import from git_real, git real can't import from curriculum tracking. If we were to take the whole git_real app and paste it into another django project  that has nothing to do with learner management then it should just work. 

Probably the best way to do this would be to implement the Push viewset inside the curriculum_tracking.api_views.   Git Real shouldn't need to know that projects even exist. 

That said, the SErializer class can be implemented in this app

sheenarbw (changes requested):
Found the problems

PR #272, opened 2021-11-27
Correct colors in pie chart

### LINKED ISSUE #233: Make use of correct colours in pie chart here src/components/regions/UserDashboard/UserDetailedStats.js
Please use feat/user-cart-stats-pie-chart as a base branch. The pie chart is already there.

If you look at a team card summary page then you can see what colour is assiciated with cards that are in progress, blocked, ready etc. Please make use of those colours for the different cells of the pie chart.

Please note: You definately should not be copy-pasting colours. Code should be DRY. 

Also please note that the pie chart will not draw a cell for all of the different card statuses. If there are no cards of a specific status then the cell will not be drawn. So you need to be careful to choose the correct colour for each cell.

Sbonelo01:
Related issues: [https://github.com/Umuzi-org/Tilde/issues/233]

## Description: The pie chart makes use of the correct colors. If there is no card to draw the data from ( when the card value is 0 ), the pie chart does not draw it. To test this, you can edit the `UserDetailedStats.json` file 

## Screenshots/videos

Screenshot from 2021-11-27 22-24-24

## I solemnly swear that:


babalwa-mbolekwa (approved):

FaithMo (approved):

frontend/src/components/regions/UserDashboard/UserDetailedStats.js
```diff
@@ -7,10 +7,15 @@ import {
   TableRow,
   Typography,
 } from "@material-ui/core";
-
+import { colors } from "@material-ui/core";
 import { PieChart, Pie, Cell, Tooltip } from "recharts";
 
-const COLORS = ["#0088FE", "#00C49F", "#FFBB28", "#FF8042"];
+const COLORS = [
+  colors.yellow[600],
```
sheenarbw: Where did you get these colours from? Should you not be importing them?  I'm sure we use the colours in multiple places?

sheenarbw (changes requested):
This doesn't look DRY. Did you copy paste the colours?

babalwa-mbolekwa (approved):

FaithMo (approved):

frontend/src/colors.js
```diff
@@ -0,0 +1,17 @@
...
+  READY,
+  BLOCKED,
+  COMPLETE,
+  IN_REVIEW,
+  IN_PROGRESS,
+  REVIEW_FEEDBACK,
+} from "./constants";
+import { colors } from "@material-ui/core";
+export const cardColors = {
+  [COMPLETE]: colors.yellow[600],
```
sheenarbw, line 11: where else are these colours used? Surely they should be imported in more than one place? Where did you get the colours from originally? Are the colours still defined twice?

sheenarbw (changes requested):
Still not DRY.

Why are the colours still set up in src/components/widgets/CardStatusChip.js instead of imported?

frontend/src/colors.js
Sbonelo01: @sheenarbw I was not working with `src/components/widgets/CardStatusChip.js` 

Does the issue require me to fix all the colors and not only draw the correct colors in the pie chart?

Vuyisanani (commented):
Hi  @EckardBerry 

Can you also  please review this PR

Vuyisanani (changes requested):
Please fix your merge conflicts

frontend/src/components/widgets/CardStatusChip.js
```diff
@@ -0,0 +1,41 @@
...
+  BLOCKED,
+  READY,
+  IN_PROGRESS,
+  REVIEW_FEEDBACK,
+  IN_REVIEW,
+  COMPLETE,
+} from "../../constants";
+import { cardColors } from "../../colors";
+const useStyles = makeStyles((theme) => {
+  console.log(cardColors.RF);
```
Vuyisanani: Please remove this

frontend/src/components/regions/UserDashboard/UserDetailedStats.js
```diff
@@ -7,10 +7,10 @@ import {
   TableRow,
   Typography,
 } from "@material-ui/core";
-
+import { cardColors } from "../../../colors";
 import { PieChart, Pie, Cell, Tooltip } from "recharts";
 
-const COLORS = ["#0088FE", "#00C49F", "#FFBB28", "#FF8042"];
+const [...COLORS] = Object.values(cardColors);
```
sheenarbw: Take a look at how the colours are accessed. Do you think just grabbing the colours by index makes sense? Should we not be writing code that explicittly grabs the colour by status?  (Line 79)

Right now, if I were to randomly rearrange the colours in the colours.js file then the colours in this file will start to mean differnt things. So this file needs to know about the order of the colours in a completely different file. This is brittle.

sheenarbw (changes requested):

babalwa-mbolekwa (approved):

Vuyisanani (approved):

sheenarbw (approved):

PR #279, opened 2021-12-02
Set due dates in bulk/106

EckardBerry:

## Description:

1. Added some code to the ManagmentActions view, see backend/curriculum_tracking/api_views.py (lines 844 - 913)
2. Added some code to the SetDueDate serializer, see backend/curriculum_tracking/serializers.py (lines 248 - 253)
3. Added some tests to test_api_views, see backend/core/tests/test_api_views.py (lines 130 - 223)

## I solemnly swear that:


backend/curriculum_tracking/api_views.py
```diff
@@ -838,12 +841,18 @@ class WorkshopAttendanceViewset(viewsets.ModelViewSet):
...
+
     serializer_class = serializers.NoArgs
+    queryset = models.Team.objects.filter(active=True)
 
     def list(self, request):
         return Response([])
 
+    def get_object(self):
+        queryset = self.get_queryset()
+        return queryset
```
TheAbstract: We can just return it;
```suggestion
    def get_object(self):
        return self.get_queryset()
```

backend/core/tests/test_api_views.py
```diff
@@ -124,3 +125,99 @@ def test_staff_users_cant_see_all_teams(self):
         url = self.get_list_url()
         response = self.client.get(url)
         self.assertEqual(len(response.data), 0)
+
+
+class TestBulkSetDueDatesApi(APITestCase, APITestCaseMixin):
```
TheAbstract: Is there a reason this isn't in **curriculum_tracking/tests**?

backend/curriculum_tracking/management/commands/set_bulk_due_dates.py
```diff
@@ -0,0 +1,27 @@
+from django.core.management.base import BaseCommand
+from curriculum_tracking.models import AgileCard, Team
+from ..helpers import get_team_cards
+
+
+class Command(BaseCommand):
+    """
+    Generic command structure: python manage.py set_bulk_due_dates "team_name" "content_item_id" "flavour" "YYYY-MM-DD"
+    command example: python manage.py set_bulk_due_dates bomb_squad 2 Python 2021-02-18
+    """
```
TheAbstract: Why do we have both a an endpoint and a command? If we've got one, isn't the other redundant?

TheAbstract (changes requested):
@EckardBerry good work here, a few questions below.

Also, a test I think is missing here: suppose we have `blue_team` and set the due date for a project to **date x**, can we check that the due date for that project is actually **date x** not **date y**.

backend/curriculum_tracking/api_views.py
EckardBerry: I changed this to your way.

backend/core/tests/test_api_views.py
EckardBerry: In the instructions of this ticket it says this at the bottom "Tests can go in core/tests/test_api_views.py"

backend/curriculum_tracking/management/commands/set_bulk_due_dates.py
EckardBerry: The instructions mentioned something about a command and something about an API endpoint, so I created both.  I can remove the command if it is not needed?

backend/curriculum_tracking/api_views.py
EckardBerry: I hear you about the extra test but at the moment I can't do that unless that is what this new function should be doing.  The instructions say that we select the correct cards according to team, content_item and flavours.  Therefore, we are updating the due times of cards and not of projects.  The functionality is built to update cards and not projects.  If that should change then I can have a test that updates due_times for projects?

Andy-Nkumane (changes requested):
two tests fail `test_set_project_card_due_time_permissions` and `test_set_topic_card_due_time_permissions` - getting a status code 400 instead of 200

EckardBerry:
@Andy-Nkumane You are running the wrong tests, I don't have tests with those names.  Here are my tests and they all pass:

test_team_members_cannot_bulk_set_due_dates_for_the_team_they_belong_to
test_team_members_cannot_bulk_set_due_dates_for_teams_they_dont_belong_to
test_super_user_can_bulk_set_due_dates_for_a_team
test_bulk_set_due_date_happened_for_every_card_with_the_same_content_item_for_the_team

Andy-Nkumane (approved):

backend/curriculum_tracking/api_views.py
```diff
@@ -883,6 +891,26 @@ def auto_assign_reviewers(self, request, pk=None):
...
+    def bulk_set_due_dates(self, request, pk=None):
+        serializer = self.get_serializer(data=request.data)
+        if serializer.is_valid():
+            teams = self.get_object()   # This will get you all the teams, we only need the teams in the request data
+            for team in teams:
+                if team.name in request.data.get('team'):
+                    team_cards = get_team_cards(team, request.data.get('content_item'))
+
+            for card in team_cards:
+                AgileCard.set_due_time(card, request.data.get('due_time'))
```
sheenarbw: Please use oop in conventional ways: `card.set_due_time(request.data.get('due_time'))`

`SomeClass.any_instance_method` is always bad practice.
`SomeClass.a_class_method` is good practice
`SomeClass.a_static_method` is also good

backend/curriculum_tracking/api_views.py
```diff
@@ -883,6 +891,26 @@ def auto_assign_reviewers(self, request, pk=None):
...
+            return Response(f'Due dates set for these team cards: {[card for card in team_cards]}')
```
sheenarbw: The frontend expects a JSON response.  
Generally when we have an api endopoint that edits an instance we have the endpoint return that instance. That way the frontend can know what alterations were made just by looking at the response instead of having to implement any intelligence.

Eg. Take a look at the "start_project" action on the AglieCard viewset. 

It would be good to have a response that the frontend (or any client application) could find useful.  I'd recommend making use of the `CardSummarySerializer`

backend/curriculum_tracking/api_views.py
```diff
@@ -883,6 +891,26 @@ def auto_assign_reviewers(self, request, pk=None):
...
+        serializer_class=serializers.SetDueTimeSerializer,
+        permission_classes=[HasObjectPermission(permissions=Team.PERMISSION_MANAGE_CARDS)]
+    )
+    def bulk_set_due_dates(self, request, pk=None):
+        serializer = self.get_serializer(data=request.data)
+        if serializer.is_valid():
+            teams = self.get_object()   # This will get you all the teams, we only need the teams in the request data
+            for team in teams:
+                if team.name in request.data.get('team'):
+                    team_cards = get_team_cards(team, request.data.get('content_item'))
```
sheenarbw: Should we worry about flavour_matching?

backend/curriculum_tracking/api_views.py
```diff
@@ -883,6 +891,26 @@ def auto_assign_reviewers(self, request, pk=None):
...
+    @action(
+        detail=False,
+        methods=["POST"],
+        serializer_class=serializers.SetDueTimeSerializer,
+        permission_classes=[HasObjectPermission(permissions=Team.PERMISSION_MANAGE_CARDS)]
+    )
+    def bulk_set_due_dates(self, request, pk=None):
+        serializer = self.get_serializer(data=request.data)
+        if serializer.is_valid():
+            teams = self.get_object()   # This will get you all the teams, we only need the teams in the request data
```
sheenarbw: There should only be one team.

If we make this an action on the TeamVIewSet and have `detail=True` then the loop wont be needed.

backend/curriculum_tracking/management/commands/set_bulk_due_dates.py
sheenarbw: I think this can be removed.

It's important to read instructions critically. 

Initially this issue was designed to allow a new person on the team to learn some useful skills while adding value. Managment scripts are easy to make as compared to api endpoints. But if we can access functionality via the api we dont need the managment script, it wont get used

sheenarbw (changes requested):
I would move stuff around a bit.

I think the api endpoint makes more sense on the Team viewset because the the endpoint should set the due date for a team.

And you have a very strange habit around calling instance methods. I've seen you do the same thing a few times and whenever I notice it I ask you to fix it. It would be good to break that habit. Also please mark learners as Not Yet Competent if they make that same mistake.

backend/curriculum_tracking/api_views.py
EckardBerry: I don't think so, take for instance the team provisionally accepted java, each team member will only be doing java, so the cards we get from calling 'get_team_cards' should therefore only return 'java cards'.

backend/curriculum_tracking/api_views.py
EckardBerry: i tried setting detail=True but the result is that it removes the option to bulk set due dates from the @actions list.  I found a work around though.

backend/curriculum_tracking/api_views.py
```diff
@@ -838,8 +841,10 @@ class WorkshopAttendanceViewset(viewsets.ModelViewSet):
     ]
 
 
-class ManagmentActionsViewSet(viewsets.ViewSet):
+class ManagmentActionsViewSet(viewsets.ViewSet, GenericAPIView):
+
     serializer_class = serializers.NoArgs
+    queryset = models.Team.objects.filter(active=True)
```
sheenarbw: What is this for? Why the queryset?

backend/curriculum_tracking/management/commands/set_bulk_due_dates.py
sheenarbw: This managment script wont be used. So it can be removed.

backend/core/views.py
```diff
@@ -144,6 +149,20 @@ def summary_stats(self, request, pk=None):
     # def shuffle_reviewers(self, request, pk=None):
     #     return Response("TODO")
 
+    @action(
+        detail=False,
```
sheenarbw: detail=True

backend/core/views.py
```diff
@@ -104,6 +106,9 @@ def get_queryset(self):
         )
         return queryset
 
+    def get_object(self):
```
sheenarbw: What is this for?

backend/core/views.py
```diff
@@ -144,6 +149,20 @@ def summary_stats(self, request, pk=None):
     # def shuffle_reviewers(self, request, pk=None):
     #     return Response("TODO")
 
+    @action(
+        detail=False,
+        methods=["POST"],
+        serializer_class=CardSummarySerializer,
```
sheenarbw: This is not the right serialiser. Please just accept the arguments that the view needs.

We know the team from the url. We need to know some other stuff as well.

backend/core/views.py
```diff
@@ -144,6 +149,20 @@ def summary_stats(self, request, pk=None):
...
+        permission_classes=[HasObjectPermission(permissions=Team.PERMISSION_MANAGE_CARDS)]
+    )
+    def bulk_set_due_dates(self, request, pk=None):
+
+        team: models.Team = models.Team.objects.filter(name=request.data.get('team')).first()
```
sheenarbw: Since this should be a detail view, getting the team should be much simpler than this.

backend/core/views.py
```diff
@@ -144,6 +149,20 @@ def summary_stats(self, request, pk=None):
...
+        team_cards = get_team_cards(team, request.data.get('content_item'))
```
sheenarbw: Please make use of the serialiser to identify the cards. Get the content item from the serialiser.

Do flavours matter here? They might

backend/core/views.py
```diff
@@ -144,6 +149,20 @@ def summary_stats(self, request, pk=None):
...
+        [card.set_due_time(request.data.get('due_time')) for card in team_cards]
+
+        return Response(AgileCardSerializer([card for card in team_cards][0]).data)
```
sheenarbw: I think the card summary serialiser can be used at the end. It returns less data so it's more efficient. Also, chances are that this endpoint will be accessed from the card summary page so it would make sense to make use of serialisers used for that page

sheenarbw (changes requested):
Some stuff here seems a little odd. Some of the choices made dont make a lot of sense

backend/curriculum_tracking/api_views.py
EckardBerry: Should have removed it, something I used previously but is no longer necessary.

backend/core/views.py
EckardBerry: If I set detail=True I can no longer see the option 'bulk_set_due_dates' under the 'Extra actions' tab.

backend/core/views.py
EckardBerry: If that is not there then some of the tests will fail cause they are expecting you to call the viewset with pk assigned to a team id, but we are not selecting teams like that, we are not doing something like this:

http://127.0.0.1:8000/api/teams/2/bulk_set_due_dates/ 

we are only doing this:
http://127.0.0.1:8000/api/teams/bulk_set_due_dates/

The error I get if I don't have get_object(self) is this:

    assert lookup_url_kwarg in self.kwargs, (
AssertionError: Expected view TeamViewSet to be called with a URL keyword argument named "pk". Fix your URL conf, or set the `.lookup_field` attribute on the view correctly.

backend/core/views.py
EckardBerry: I am not sure I understand, you previously suggested I make use of the CardSummarySerializer.

backend/core/views.py
EckardBerry: I tried using the CardSummarySerializer here but I keep on getting this error:

Got AttributeError when attempting to get a value for field `team` on serializer `CardSummarySerializer`.
The serializer field might be named incorrectly and not match any attribute or key on the `AgileCard` instance.
Original exception text was: 'AgileCard' object has no attribute 'team'

The error doesn't make sense to me since I did add an extra field 'team' to the CardSummarySerializer?

backend/core/views.py
EckardBerry: I see what the problem is here.  If you set detail=True (Getting a single object) but you don't select a team via team id, ergo, you don't have a value for 'pk' then you are not able to make use of the 'bulk_set_due_dates' under the 'Extra actions' tab.  So the fix was to simply select a team by team id.

backend/core/views.py
EckardBerry: I am pretty sure I've invariably fixed this now, I don't think I will need "def get_object(self)" anymore as I will now be passing a value to pk.

backend/core/views.py
EckardBerry: I've changed this a bit so that I get the content_item from the serializer.  Not so sure about the flavours part though cause all members of a team will all be Python devs or Javascript devs etc, do we then still need to check flavours?

backend/core/views.py
EckardBerry: Sorted this error out, I now make use of CardSummarySerializer only.

Andy-Nkumane (changes requested):
hi, this looks good to me, however, there are a few things to look:
* this branch is behind develop by a number of commits
* there is a merge conflict

backend/core/views.py
TheAbstract: @EckardBerry has this been resolved; https://github.com/Umuzi-org/Tilde/pull/279/files#r769351094?

backend/curriculum_tracking/api_views.py
```diff
@@ -842,7 +843,8 @@ class WorkshopAttendanceViewset(viewsets.ModelViewSet):
     ]
 
 
-class ManagmentActionsViewSet(viewsets.ViewSet):
+class ManagmentActionsViewSet(viewsets.ViewSet, GenericAPIView):
```
TheAbstract: Where does `GenericAPIView` come from? I don't see it imported anywhere.

backend/core/tests/test_api_views.py
```diff
@@ -124,3 +125,116 @@ def test_staff_users_cant_see_all_teams(self):
...
+        card.assignees.add(self.user_one_red)
+        self.login(self.user_one_blue)
+        url = f'{self.get_instance_url(pk=self.red_team.id)}bulk_set_due_dates/'
+        response = self.client.post(path=url, format="json", data={
+            'due_time': '2021-12-03T14:17', 'content_item': '2', 'team': str(self.red_team),
+            'flavour_names': 'JAVASCRIPT', 'status': card.status
+        })
+        self.assertEqual(response.status_code, 404)
+
+    def test_super_user_can_bulk_set_due_dates_for_a_team(self):
```
TheAbstract: If only a super user can set due dates then does it matter which team they belong to? I think it's enough to have `test_super_user_can_bulk_set_due_dates_for_a_team` and then one `test_non_super_user_cannot_set_due_dates` instead of the two different ones above.

TheAbstract (changes requested):
The tests don't run because of a missing import. See comments below.

backend/core/views.py
EckardBerry: @TheAbstract No sure brother, I've changed serializers and everything works but I am not sure if we can call it resolved.

backend/curriculum_tracking/api_views.py
EckardBerry: Forgot to remove this, it is gone now.

backend/core/tests/test_api_views.py
EckardBerry: The permission is permissions=Team.PERMISSION_MANAGE_CARDS.  Therefore, correct me if I am wrong, I think that if a team member is granted access then that team member can do it.  So I am basically checking that if a team member has not been given permission that that team member cannot bulk set due dates.  I am also checking that with members from different teams.  Do you still think I should remove these tests?

TheAbstract: Okay cool didn't release that about the permissions.

TheAbstract (changes requested):
While the tests you've added for this themselves seem fine when all the tests are run there are some issues. So it looks like backward compatibility is being broken.
```cmd
FAIL: test_set_project_card_due_time_permissions (curriculum_tracking.tests.test_api_views.AgileCardViewsetTests)
  File "/Users/umuzi/Tilde/backend/curriculum_tracking/tests/test_api_views.py", line 110, in test_set_project_card_due_time_permissions
    self._test_set_due_time_permissions(card, lambda card: card.recruit_project)
  File "/Users/umuzi/Tilde/backend/curriculum_tracking/tests/test_api_views.py", line 140, in _test_set_due_time_permissions
AssertionError: 400 != 200

FAIL: test_set_topic_card_due_time_permissions (curriculum_tracking.tests.test_api_views.AgileCardViewsetTests)
  File "/Users/umuzi/Tilde/backend/curriculum_tracking/tests/test_api_views.py", line 116, in test_set_topic_card_due_time_permissions
    self._test_set_due_time_permissions(card, lambda card: card.topic_progress)
  File "/Users/umuzi/Tilde/backend/curriculum_tracking/tests/test_api_views.py", line 140, in _test_set_due_time_permissions
AssertionError: 400 != 200

FAIL: test_get_list (curriculum_tracking.tests.test_api_views.CardSummaryViewsetTests)
  File "/Users/umuzi/Tilde/backend/test_mixins.py", line 62, in test_get_list
    self.assertTrue(bool(value), f"{key} = {value}. The value must be truthy")
AssertionError: False is not true : flavour_names = []. The value must be truthy

Ran 228 tests in 89.166s

FAILED (failures=3)
```
Is everything passing on your end when you run `python manage.py test`?

EckardBerry:
@TheAbstract 

When I run python manage.py test to test all the tests that exist I get the same three errors, however, even if I comment out all my code and tests I still get those three errors.  I suspect the issue is therefore not with my code/tests.

TheAbstract (commented):
Okay then let's let Sheena take a look and advice what should be done.

backend/core/tests/test_api_views.py
```diff
@@ -124,3 +125,116 @@ def test_staff_users_cant_see_all_teams(self):
...
         self.assertEqual(len(response.data), 0)
+
+
+class TestBulkSetDueDatesApi(APITestCase, APITestCaseMixin):
+
+    LIST_URL_NAME = "team-list"
+    SUPPRESS_TEST_POST_TO_CREATE = True
+    SUPPRESS_TEST_GET_LIST = True
+
+    def verbose_instance_factory(self):
```
sheenarbw: Do we need the verbose instance factory? It is only used in tests that have been suppressed. Does deleting this function make your tests fail?

backend/core/tests/test_api_views.py
```diff
@@ -124,3 +125,116 @@ def test_staff_users_cant_see_all_teams(self):
...
+    def test_team_members_cannot_bulk_set_due_dates_for_the_team_they_belong_to(self):
+
+        project = RecruitProjectFactory(due_time=None)
+        card = AgileCardFactory(recruit_project=project)
+        card.reviewers.add(self.user_one_red)
+        card.assignees.add(self.user_two_red)
+        self.login(self.user_two_red)
+        url = f'{self.get_instance_url(pk=self.red_team.id)}bulk_set_due_dates/'
+        response = self.client.post(path=url, format="json", data={
+            'due_time': '2021-12-03T14:17', 'content_item': str(card.content_item.id), 'team': str(self.red_team),
```
sheenarbw: Are you using the black autoformatter? 

Also, are you sure it is necessary to convert the ids to strings?

backend/core/tests/test_api_views.py
```diff
@@ -124,3 +125,116 @@ def test_staff_users_cant_see_all_teams(self):
...
+        response = self.client.post(path=url, format="json", data={
+            'due_time': '2021-12-03T14:17', 'content_item': '2', 'team': str(self.red_team),
+            'flavour_names': 'JAVASCRIPT', 'status': card.status
+        })
+        self.assertEqual(response.status_code, 404)
+
+    def test_super_user_can_bulk_set_due_dates_for_a_team(self):
+
+        project = RecruitProjectFactory(due_time=None)
+        card = AgileCardFactory(recruit_project=project, flavours=['JAVASCRIPT'])
```
sheenarbw: WHY ARE YOU SHOUTING ABOUT JAVASCRIPT? 

you might have seen it in capital letters in a diferent file. But notice that in that case it doesn't have quotation marks, its a constant.  This sort of thing:

```
JAVASCRIPT = 'js'
```

sheenarbw: This isn't a big deal, but if you are pattern matching then it would be good to look at things closely

backend/core/tests/test_api_views.py
```diff
@@ -124,3 +125,116 @@ def test_staff_users_cant_see_all_teams(self):
...
+        self.login(self.super_user)
+        url = f'{self.get_instance_url(pk=self.blue_team.id)}bulk_set_due_dates/'
+        response = self.client.post(path=url, format='json', data={
+            'due_time': '2021-12-03T14:17', 'content_item': str(card.content_item.id), 'team': str(self.blue_team),
+            'flavour_names': 'JAVASCRIPT', 'status': card.status
+        })
+
+        self.assertEqual(response.status_code, 200)
+        card.refresh_from_db()
+        self.assertIsNotNone(card.due_time)
```
sheenarbw: What exactly should the due time be equal to? Surely we know the exact value it should have?

backend/core/tests/test_api_views.py
```diff
@@ -124,3 +125,116 @@ def test_staff_users_cant_see_all_teams(self):
...
+        self.assertEqual(response.status_code, 200)
+
+        team_cards = get_team_cards(self.red_team, card_1.content_item)
+        self.assertIn(card_1, team_cards)
+        self.assertIn(card_2, team_cards)
+        self.assertNotIn(card_3, team_cards)
+
+        for card in team_cards:
+            card.refresh_from_db()
+            self.assertIsNotNone(card.due_time)
```
sheenarbw: What should the value be?

backend/core/views.py
```diff
@@ -104,6 +106,9 @@ def get_queryset(self):
         )
         return queryset
 
+    #def get_object(self):
```
sheenarbw: can you please remove this? Do does it serve a purpose?

backend/core/views.py
sheenarbw: When you go to the browsable api "http://localhost.... /api" and navigate to this endpoint, then you will see a form where you can post data to this endpoint. 

It should be the correct form. It should work.

Use a serialiser that accepts the correct fields so that the form works.

Just use the CardSummarySerializer when returning the response.

backend/core/views.py
```diff
@@ -144,6 +149,25 @@ def summary_stats(self, request, pk=None):
...
+    def bulk_set_due_dates(self, request, pk=None):
+
+        team: models.Team = models.Team.objects.get(pk=pk)
+        serializer = self.get_serializer(data=request.data)
+
+        if serializer.is_valid():
+            team_cards = get_team_cards(team, serializer.validated_data.get('content_item'))
+            [
+                card.set_due_time(request.data.get('due_time')) for card in team_cards if
+                card.flavour_names in [[serializer.validated_data.get('flavour_names')]]
```
sheenarbw: Does your function work if there are multiple flavours passed in?  

I would recommend a more robust flavour checking mechanism. 

You could use `card.flavours_match(serializer.validated_data.get('flavour_names'))`

backend/curriculum_tracking/serializers.py
```diff
@@ -221,8 +222,12 @@ class Meta:
             "open_pr_count",
             "oldest_open_pr_updated_time",
             "repo_url",
+            "flavour_names"
```
sheenarbw, line 225: This shouldn't be here

backend/core/tests/test_api_views.py
sheenarbw: Remember explicit is better than implicit

backend/curriculum_tracking/serializers.py
```diff
@@ -244,16 +249,16 @@ class Meta:
         fields = ["timestamp"]
 
 
-class AddReviewerUserSerializer(serializers.ModelSerializer):
+class SetDueTimeSerializer(serializers.ModelSerializer):
```
sheenarbw, line 252: Maybe a better name would include the word "bulk".  It would be good to be explicit aboput what this serialser is for. 

This should also be the serializer class associated with the new action.

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -29,6 +29,7 @@ class CardSummaryViewsetTests(APITestCase, APITestCaseMixin):
         "start_time",
         "open_pr_count",
         "oldest_open_pr_updated_time",
+        "flavour_names",
```
sheenarbw, line 49: this change shouldn't be here

backend/core/tests/test_api_views.py
sheenarbw: So it would be good to test that PERMISSION_MANAGE_CARDS works in this case.  Give someone that permission and make sure that it has the desired effect.

sheenarbw (changes requested):
This still needs some work

backend/core/tests/test_api_views.py
```diff
@@ -4,7 +4,8 @@
 from core.models import Team
 from guardian.shortcuts import assign_perm
 from django.contrib.auth.models import Group
-
+from curriculum_tracking.tests.factories import RecruitProjectFactory, AgileCardFactory
+from curriculum_tracking.management.helpers import get_team_cards
```
TheAbstract: We don't need these anymore.
```suggestion
```

backend/core/views.py
```diff
@@ -104,6 +106,9 @@ def get_queryset(self):
         )
         return queryset
 
+    #def get_object(self):
+        #return self.get_queryset()
```
TheAbstract: Since this isn't necessary anymore (https://github.com/Umuzi-org/Tilde/pull/279#discussion_r777852234) it can be removed.

backend/curriculum_tracking/api_views.py
```diff
@@ -25,6 +24,8 @@
 from django.db.models import Q
 
 from rest_framework import filters
+from curriculum_tracking.management.helpers import get_team_cards
```
TheAbstract: Not used anywhere.
```suggestion
```

TheAbstract (changes requested):
Mostly small changes. There seems to be some items from Sheena that haven't been sorted yet. I probably missed some stuff while I was in the "dark ages", so if there's something that was discussed and is no longer relevant could you just mark it as **resolved**.

There's still bunch of tests failing here, was this discussed during the pairing session this week? How are we handling this?

backend/curriculum_tracking/api_views.py
EckardBerry: Hello

I made the requested changes.  As for the tests that fail, I don't know, we worked on Andy's ticket last time.  I see that the tests that fail have to do with WorkshopAttendanceViewsetTests, I did not work on those things so I really don't know whats up with them but I will see if maybe I can figure out something:

```
  File "/home/eckard/PycharmProjects/Tilde/backend/test_mixins.py", line 48, in test_get_list
    instance = self.verbose_instance_factory()
  File "/home/eckard/PycharmProjects/Tilde/backend/curriculum_tracking/tests/test_api_views.py", line 349, in verbose_instance_factory
    workshop = factories.ContentItemFactory(content_type=ContentItem.WORKSHOP)
AttributeError: module 'core.tests.factories' has no attribute 'ContentItemFactory'
```

EckardBerry: The issue with all those failing tests is that the person who wrote them are trying to import from the wrong factories.py file.  I've now gone and imported everything and now all those tests pass.  However, there are some other tests like this one "def _test_set_due_time_permissions(self, card, get_progress):" that still have errors, I don't want to go and fool around with them because they are not mine.

EckardBerry:
Closing this PR, created a new branch and new PR which shows all the previous commits that took place.

PR #295, opened 2021-12-09
Setup project repo api end point/277

### LINKED ISSUE #277: Turn the following managment script into an api endpoint: setup_project_repo
This is a managment script that is required from time to time.

example usage:

python manage.py setup_project_repo "semitone difference - basic algorithm" "luke.scott@umuzi.org" "javascript"

Note that long running requests should make use of dramatiq

The api action should be added to the AgileCard viewset

EckardBerry:

## Description:

- Added @action to AgileCardViewset in curriculum_tracking/api_views.py in order to call a previously built function 'setup_repository()' but this time from an api endpoint.
- Added a dramatiq actor in backend/long_running_request_actors.py
- Added a test to test if the function is indeed called from the api endpoint, test is in curriculum_tracking/tests/test_api_views.py in the class AgileCardViewsetTests

## I solemnly swear that:


Andy-Nkumane (approved):

backend/git_real/helpers.py
```diff
@@ -45,7 +45,7 @@ def upload_readme(api, repo_full_name, readme_text):
             json=False,
         )
 
-    assert str(response.status_code).startswith("2"), f"{response}\n {response.json()}"
+    #assert str(response.status_code).startswith("2"), f"{response}\n {response.json()}"
```
sheenarbw: Please put this back. We are assertting that the api call was sucessful. We do need it to be successful

backend/long_running_request_actors.py
```diff
@@ -72,3 +72,11 @@ def auto_assign_reviewers():
     )
 
     work()
+
```
sheenarbw: There is another dramatiq actor in the same file that looks like this:

```
@dramatiq.actor()
def recruit_project_setup_repository(project_id):
    from curriculum_tracking.models import RecruitProject
    project = RecruitProject.objects.get(pk=project_id)
    project.setup_repository()

```

I'm sure you can see that this code is not DRY. Why not make use of the existing actor? Juist pass it the project id instead of the card id.

sheenarbw (changes requested):
Close. There's some stuff that's not DRY, and you commented out something that matters

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -234,6 +238,51 @@ def test_set_project_link(self):
...
+        card = factories.AgileCardFactory(
+            content_item=content_item,
+            flavours=[JAVASCRIPT],
+            status=AgileCard.READY,
+            recruit_project=None,
+        )
+
+        assignee = SocialProfileFactory().user
+        card.assignees.set([assignee])
+        card.start_project()
```
sheenarbw: This line also calls setup_repository. It might be good to make sure that setup_repository has been called the correct number of times.

sheenarbw (changes requested):
I think that the test would pass if the view didn't do anything.

backend/curriculum_tracking/tests/test_api_views.py
EckardBerry: This dramatiq thing is strange to me, if I for instance put a breakpoint in the view and then run the test, once I skip past this part "response = recruit_project_setup_repository.send(card.recruit_project.id)" of the view and then print response it shows this:

Message(queue_name='default', actor_name='recruit_project_setup_repository', args=(1,), kwargs={}, options={}, message_id='41f05619-7664-4701-90d4-0fb4074b403a', message_timestamp=1639572949391)

which means it worked, however, in the test itself when I get to this "self.assertTrue(project.setup_repository.called)" it will return False saying setup_repository was never called which is not the case when you consider that the view tells you the message was sent to an actor called 'recruit_project_setup_repository' with the correct arguments. 

In fact, if I change the view so that I no longer make use of dramatiq then everything works perfectly, even the test, but the moment I use dramatiq it says setup_repository wasn't called, is there not another way to test dramatiq actors?

EckardBerry: Finally got the test to work with dramatiq.

Andy-Nkumane (approved):

backend/curriculum_tracking/api_views.py
```diff
@@ -25,6 +24,7 @@
 from django.db.models import Q
 
 from rest_framework import filters
+import dramatiq
```
sheenarbw: Why are you importing dramatiq?

backend/curriculum_tracking/api_views.py
```diff
@@ -561,6 +561,25 @@ def cancel_workshop_attendance(self, request, pk=None):
         card.delete_workshop_attendance()
         return Response(serializers.AgileCardSerializer(card).data)
 
+    @action(
+        detail=True,
+        methods=["POST"],
+        serializer_class=serializers.NoArgs,
+        permission_classes=[(curriculum_permissions.IsCardAssignee) | IsStaffUser | HasObjectPermission(
```
sheenarbw: HasObjectPermission is misused. Since the get_objects argument was used it will use the default behavior. So that means we will check if the current user has Team.PERMISSION_MANAGE_CARDS on the current card. This will always be false, nobody is ever given any kind of permission over individual cards. 

You need to check that the logged in user has specific permissions on any of the teams that the card is associated with.

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -234,6 +239,45 @@ def test_set_project_link(self):
...
+        )
+
+        assignee = SocialProfileFactory().user
+        card.assignees.set([assignee])
+        project = card.recruit_project
+        url = f"{self.get_list_url()}{card.id}/setup_project_repo/"
+        self.login(super_user)
+        response = self.client.post(path=url, data={"card_id": card.id})
+        self.assertEqual(response.status_code, 200)
+        self.assertTrue(project.setup_repository.assert_called)
```
sheenarbw: project.setup_repository.assert_called is a function. Asserting that it is True is meaningless because functions are truthy.

```
def foo():pass
self.assertTrue(foo) # this never fails
```

If we use the method correctly then this is what happens:

```
project.setup_repository.assert_called()
*** AssertionError: Expected 'setup_repository' to have been called.
```

So the tests think it wasn't called. 

It was called in a seperate worker process. That seperate worker process doesn't know it is unit testing anything, it's just a worker. So it doesn't know that you are trying to mock anything. So it doesn't even record what got called and what didn't.

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -234,6 +239,45 @@ def test_set_project_link(self):
         project.refresh_from_db()
         self.assertEqual(project.link_submission, link_2)
 
+    @mock.patch.object(RecruitProject, "setup_repository")
+    def test_setup_project_repo_call_from_api_action_option(self, setup_repository):
+
+        JAVASCRIPT = "js"
+        super_user = factories.UserFactory(is_superuser=True)
+
+        def get_repo_mock(
```
sheenarbw: Where is this used? Are you sure you need it?

sheenarbw (changes requested):
It's a good start but there are some permission errors and the test assertions are incorrect. 

Testing distributed processes can be tricky though.

backend/curriculum_tracking/tests/test_api_views.py
EckardBerry: I used it before, but you're right, I don't need it anymore so I got rid of it.

Andy-Nkumane (approved):

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -235,6 +247,47 @@ def test_set_project_link(self):
...
+        self.card.assignees.set([self.assignee])
+        self.project = self.card.recruit_project
+        self.url = f"{self.get_list_url()}{self.card.id}/setup_project_repo/"
+
+    @mock.patch.object(long_running_request_actors.recruit_project_setup_repository, 'send')
+    def test_send_called_from_api_view_action_option(self, send):
+        self.login(self.super_user)
+        try:
+            response = self.client.post(path=self.url, data={"card_id": self.card.id})
+        except RecursionError:
```
sheenarbw: There really shouldn't be a recursion error.

sheenarbw (changes requested):
Some dodgy things in the test.

Andy-Nkumane (approved):

backend/long_running_request_actors.py
```diff
@@ -13,28 +13,34 @@
 PRIORITY_HIGH = 0
 
 
-os.environ.setdefault("DJANGO_SETTINGS_MODULE", "backend.settings")
```
sheenarbw: Please put this back. It's very very important that the code actually runs. 

If you dont want this to run during unit testing then just use an if statement.

If you dont know what a piece of code is for then it's not good to just delete it. This would break prod completely, none of our dramatiq stuff would work without it because the worker wouldn't be able to access the db.

backend/long_running_request_actors.py
```diff
@@ -52,7 +58,9 @@ def test_long_running_request():
 @dramatiq.actor()
 def recruit_project_setup_repository(project_id):
     from curriculum_tracking.models import RecruitProject
+    from django.db import connection
```
sheenarbw: You can remove the print and the import.

sheenarbw (changes requested):
These changes would break our production dramatiq actor.

backend/long_running_request_actors.py
EckardBerry: I am not sure why it got deleted, also, the print statement and the import I did not put there.  I apologise for this, it is definitely not something I deliberately did.

Andy-Nkumane (approved):

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -235,6 +246,47 @@ def test_set_project_link(self):
         self.assertEqual(project.link_submission, link_2)
 
 
+class TestLongRunningActorsCalledFromAgileCardAPIView(APITestCase, APITestCaseMixin):
```
sheenarbw: Which long running actor specifically? There are a few. Can we be more explicit?

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -235,6 +246,47 @@ def test_set_project_link(self):
...
+        self.card = factories.AgileCardFactory(
+            content_item=self.content_item,
+            flavours=[JAVASCRIPT],
+            status=AgileCard.READY,
+        )
+
+        self.assignee = SocialProfileFactory().user
+        self.card.assignees.set([self.assignee])
+        self.project = self.card.recruit_project
+        self.url = f"{self.get_list_url()}{self.card.id}/setup_project_repo/"
```
sheenarbw: How about get_instance_url?

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -235,6 +246,47 @@ def test_set_project_link(self):
...
+
+    @mock.patch('curriculum_tracking.api_views.Response')
+    @mock.patch.object(long_running_request_actors.recruit_project_setup_repository, 'send')
+    def test_send_called_from_api_view_action_option(self, send, Response):
+        self.login(self.super_user)
+        Response.return_value = HttpResponse({"status": "OK"})
+        response = self.client.post(path=self.url, data={"card_id": self.card.id})
+        send.assert_called()
```
sheenarbw: what arguments should it have been called with?

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -235,6 +246,47 @@ def test_set_project_link(self):
...
+        )
+
+        self.assignee = SocialProfileFactory().user
+        self.card.assignees.set([self.assignee])
+        self.project = self.card.recruit_project
+        self.url = f"{self.get_list_url()}{self.card.id}/setup_project_repo/"
+
+    @mock.patch('curriculum_tracking.api_views.Response')
+    @mock.patch.object(long_running_request_actors.recruit_project_setup_repository, 'send')
+    def test_send_called_from_api_view_action_option(self, send, Response):
```
sheenarbw: No need to mock path response. And if there was a need to patch it then that string isn't very specific, the response is not defined in curriculum_tracking, we import it.

This works:

```
from dramatiq.message import Message

    @mock.patch.object(
        long_running_request_actors.recruit_project_setup_repository, "send"
    )
    def test_send_called_from_api_view_action_option(self, send):  

        self.login(self.super_user)
        send.return_value = Message(
            queue_name="queue", actor_name="actor", args=(), kwargs={}, options={}
        )   # send it meant to return a Message instance. So we just return one. Now it behaves in a predictable way. 
        self.client.post(path=self.url, data={"card_id": self.card.id})
        send.assert_called()
```

sheenarbw (changes requested):
Some test fixes

backend/curriculum_tracking/api_views.py
```diff
@@ -561,6 +560,27 @@ def cancel_workshop_attendance(self, request, pk=None):
...
+        serializer_class=serializers.NoArgs,
+        permission_classes=[
+            curriculum_permissions.IsCardAssignee | IsStaffUser | HasObjectPermission(
+                permissions=Team.PERMISSION_MANAGE_CARDS,
+                get_objects=_get_teams_from_card
+            )
+        ]
+    )
+    def setup_project_repo(self, request, pk=None):
+        from long_running_request_actors import recruit_project_setup_repository
```
TheAbstract, line 575: Is there a reason why this is being imported here?

TheAbstract (changes requested):
There's just this one failing test
```cmd
FAIL: test_send_called_from_api_view_action_option (curriculum_tracking.tests.test_api_views.TestRecruitProjectSetupRepositoryActorCalledFromAgileCardAPIView)
  File "/Users/umuzi/Tilde/backend/venv/lib/python3.9/site-packages/mock/mock.py", line 1346, in patched
    return func(*newargs, **newkeywargs)
  File "/Users/umuzi/Tilde/backend/curriculum_tracking/tests/test_api_views.py", line 281, in test_send_called_from_api_view_action_option
    send.assert_called_with(self.card.id)
  File "/Users/umuzi/Tilde/backend/venv/lib/python3.9/site-packages/mock/mock.py", line 913, in assert_called_with
    raise AssertionError(_error_message()) from cause
AssertionError: expected call not found.
Expected: send(130)
Actual: send(196)

Ran 226 tests in 99.097s

FAILED (failures=1)
```

ry-oc:
Looks like this was something that Eckard was working on - to be reprioritised

PR #296, opened 2021-12-13
Get neg card reviews/274

### LINKED ISSUE #274: Add a "stats" action to the AgileCard api. For now just expose the number of negative reviews on that card.

Samantha-Hampton:
@EckardBerry as just discussed, any assistance on api_viewset would be great.

I'm getting the message on my action that:
   "detail": "Method \"GET\" not allowed."

Related issues: #274 

## Description:

## I solemnly swear that:


Samantha-Hampton:
@EckardBerry 
I decided to not use the 
REVIEW_VALIDATED_STATUS_CHOICES for adding up the negative reviews - since it seems like not all reviews get validated so it might leave neg reviews out?
So I added up the NYC and R reviews instead. As mentioned, when playing around on the Shell I can get the correct neg reviews total from the reviews I created on my end. But getting them to show up on the api_viewset is proving challenging...

EckardBerry (changes requested):
Hello

I've added small comments here and there in your code and also pushed some code.  We can discuss some of the things today after stand up if you can?

Samantha-Hampton:
@EckardBerry 
See latest changes - I decided to work with the AgileCardSerializer in the end so that when the action is called it includes the rest of the card data. Can of course go back to not using a serilalizer....I also changed the naming to be explicit about what we are doing. 
Not sure exactly where to from here? Other changes? Some testing?

backend/curriculum_tracking/api_views.py
```diff
@@ -559,6 +559,45 @@ def cancel_workshop_attendance(self, request, pk=None):
...
+        #***below if done without using a serializer
+        #negative_reviews = models.RecruitProjectReview.objects.filter(status__in=['NYC', 'R']).filter(recruit_project__agile_card=card_object).count()
+
+        #return Response(negative_reviews)    # Notice my response is no longer serialized so therefore I don't need to worry about any serializers anymore
+
+        card_object = self.get_object() 
+        return Response(serializers.AgileCardSerializer(card_object).data)
+    
+
+
```
EckardBerry, line 597: Ok, so there is a lot going on here which is ok considering you are trying to explain your way of thinking.  However, I think it would be better/easier for whoever is reviewing your work if you rather write block comments and keep inline comments to a minimum.  Also, it would probably be a good idea to get rid of all the commented out code, if you think you might need it later then rather paste it inside some readme file or some personal file on your computer.  So in general, just a bit of cleaning up here.

Then also, like we discussed, we don't know yet what exactly constitutes a 'negative' review, so depending on the feedback we get from Sheena your code might change drastically.

backend/curriculum_tracking/serializers.py
```diff
@@ -77,12 +77,20 @@ class Meta:
...
     title = serializers.SerializerMethodField("get_title")
     reviewed_user_emails = serializers.SerializerMethodField("get_reviewed_user_emails")
     reviewed_user_ids = serializers.SerializerMethodField("get_reviewed_user_ids")
+    #neg_reviews = serializers.SerializerMethodField("get_neg_reviews") #**
+
+    #**
+    # def get_neg_reviews(self, instance):
+    #     nyc_reviews = models.RecruitProjectReview.objects.filter(status='NYC', recruit_project=instance.recruit_project).count()
+    #     rf_reviews = models.RecruitProjectReview.objects.filter(status='R', recruit_project=instance.recruit_project).count()
+    #     return nyc_reviews + rf_reviews
```
EckardBerry, line 93: Same thing here with the commented out code, rather store it somewhere but not here in the code base as commented out code.

backend/curriculum_tracking/serializers.py
```diff
@@ -184,15 +192,37 @@ class Meta:
...
+    total_nyc_and_rf_reviews = serializers.SerializerMethodField("get_total_nyc_and_rf_reviews") #**
 
     def get_users_that_reviewed_since_last_review_request(self, instance):
         return instance.get_users_that_reviewed_since_last_review_request()
 
+    #**
+    def get_total_nyc_and_rf_reviews(self, card):
+         negative_reviews = models.RecruitProjectReview.objects.filter(status__in=['NYC', 'R']).filter(recruit_project__agile_card=card).count()
+         return negative_reviews
+
```
EckardBerry, line 210: The way you use the serializer and how you set up your function "get_total_nyc_and_rf_reviews" is a bit strange.  For instance, if you look at the function above "get_users_that_reviewed_since_last_review_request()", you will notice that the function returns details of the instance of the class, ergo, an instance/object of the AgileCard class.

In your function you first gather info about RecruitProjectReview objects which match your criteria.  This is not wrong since you are getting the data you want, I am just now sure if it is the most efficient way of going about it.  I think that if you are going to do it this way you might as well not use a serializer at all and simply have a view and inside the view have a small function which takes care of business (much like the example I showed you where I also did not use any serialization at all).  However, wait for Sheena to confirm this first.

backend/curriculum_tracking/serializers.py
```diff
@@ -184,15 +192,37 @@ class Meta:
...
+#         fields = ["neg_reviews"]
+
+#     neg_reviews = serializers.SerializerMethodField("get_neg_reviews") #**
+
+#     #**
+#     def get_neg_reviews(self, card):
+#         nyc_reviews = models.RecruitProjectReview.objects.filter(status='NYC', recruit_project=card.recruit_project).count()
+#         rf_reviews = models.RecruitProjectReview.objects.filter(status='R', recruit_project=card.recruit_project).count()
+#         return nyc_reviews + rf_reviews
+
```
EckardBerry: Here also, just a bit of clean up w.r.t. commented out code.

EckardBerry (changes requested):
For now I think you can start writing some test cases, those tests will probably drastically change since what is required is not yet perfectly clear.  If you need help on the test cases let me know.

PR #322, opened 2022-01-10
Log card review feedback correctly called/309

### LINKED ISSUE #309: Make sure that log_card_moved_to_review_feedback is called correctly and does what it should. Note that a card can be given feedback multiple times

EckardBerry:

## Description:

1. Added some extra tests to backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
## Screenshots/videos

Sam took over

2. Added functionality in curriculum_tracking/api_views.py for logging a card either moving into review feedback or complete depending on the status of the card3. 

3. Added/amended tests to take these extra logging events into account 4. 

## I solemnly swear that:


backend/curriculum_tracking/activity_log_entry_creators.py
```diff
@@ -64,7 +64,7 @@ def log_card_moved_to_review_feedback(card, actor_user):
 
 def log_project_competence_review_done(review):
     event_type, _ = EventType.objects.get_or_create(name=COMPETENCE_REVIEW_DONE)
-    for user in review.recruit_project.recruit_users.all():
+    for user in review.recruit_project.recruit_users.all():     # Not sure why the For loop is needed, the particular card will only have one user assigned to it
```
EckardBerry: I am not sure that the For loop is needed here, the recruit_project is going to be assigned to a specific user, even if it is assigned to many different users, by the time we get to "def debounce_create" part of the filtering is according to the effected_user which to my mind is the only recruit_project we are interested in seeing that the review was done on that person's recruit_project object?

EckardBerry: Also, from def log_project_competence_review_done(review): and from def log_topic_competence_review_done(review): we always pass debounce_seconds=False.  So then, inside def debounce_create we set debounce_seconds=120 but it seems that that doesn't matter cause when we evaluate 'if debounce_seconds' we always get False anyway so I am not sure we need this?  

What if a person does a review and then within 2 minutes that same person changes their mind and therefore the review, a new Log entry won't be created but perhaps there should be?

TheAbstract (approved):
New tests passing. 
So we have a new app `activity_log`, is there a reason why we're putting these tests for it in `curriculum_tracking`?

backend/activity_log/models.py
```diff
@@ -82,8 +82,8 @@ def debounce_create(
         event_type,
         actor_user,
         effected_user,
-        object_1,
-        object_2=None,
+        object_1,       # RecruitProjectReview object
```
sheenarbw: These comments are misleading. object_1 is not always a RecruitProjectReview. Sometimes it's a topic review. Sometimes it will be a ReviewTrust instance... The idea is that we can log events to do with any kind of object, not just this one kind.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -40,3 +52,17 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+            status=review1.status,
+            timestamp=review1.timestamp,
+            comments=review1.comments,
+            topic_progress=review1.topic_progress,
+            reviewer_user=UserFactory()
+        )
+
+        creators.log_topic_competence_review_done(review1)
+        creators.log_topic_competence_review_done(review2)
+        self.assertEqual(LogEntry.objects.count(), 2)
```
sheenarbw: Also test that if we log the same thing twice, it only gets logged once.
```
 creators.log_topic_competence_review_done(review1)
 creators.log_topic_competence_review_done(review1)
self.assertEqual(LogEntry.objects.count(), 1)
```

And another missing test is making sure that the log creators get envoked at the right time. When a new review is created then the correct log creators should be called.

sheenarbw (changes requested):
Mostly good.

There are some misleading comments and missing tests.

EckardBerry:
> New tests passing. So we have a new app `activity_log`, is there a reason why we're putting these tests for it in `curriculum_tracking`?

I just added tests to tests which were already there, but if the location of the tests should move then I will move them entirely.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
EckardBerry: I've added extra tests now, the one however that I added I am not sure if you wanted it.  It basically checks that for a project review the log_topic_competence_review_done function cannot be invoked and vice versa with a topic review.

TheAbstract (approved):

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -23,6 +23,33 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+        review1 = RecruitProjectReviewFactory()
+        creators.log_project_competence_review_done(review1)
+        creators.log_project_competence_review_done(review1)
+        self.assertEqual(LogEntry.objects.count(), 1)
+
+    def test_log_project_creator_is_invoked_for_a_project_review(self):
+        review1 = RecruitProjectReviewFactory()
+        try:
+            creators.log_topic_competence_review_done(review1)
+        except AttributeError as error:
```
sheenarbw: There should not be these errors. 

If an exception is meant to be raised then you should assert that it is raised. But if it is not meant to be raised then dont squash it. Errors need to be fixed, not silenced.

sheenarbw: Maybe the error is due to the fact that you are calling log_topic... instead of log_project...

sheenarbw (changes requested):
Some dodgy stuff happening in the tests. You seem to have a strange new habit of just accepting horrible errors. We should prove that there are no horrible errors. Tests should prve that the code works so if there are horrible errors they should be fixed.


Also, you have tested that the creators work. But we also need to be sure that they are called at the right time. When a user makes a new review by interacting with the apis (via the gui) then are the creators called correctly?

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
EckardBerry: I will assert that the correct exception is raised, I deliberately called the wrong creator to test that calling the wrong creator won't work.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -23,6 +41,32 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+        creators.log_project_competence_review_done(review1)
+        creators.log_project_competence_review_done(review1)
+        self.assertEqual(LogEntry.objects.count(), 1)
+
+    def test_log_project_creator_is_invoked_for_a_project_review_else_attribute_error_is_raised(self):
+        review1 = RecruitProjectReviewFactory()
+        with self.assertRaises(AttributeError) as exception:
+            creators.log_topic_competence_review_done(review1)
+        self.assertTrue(exception, AttributeError)
+        self.assertEqual(LogEntry.objects.count(), 0)
```
TheAbstract: Not sure I get what's going on with this test. We're intentionally logging a project as a topic, and checking that an exception was raised. Is this actually related to the creator thats being tested here or is it more general?
This class is testing a specific creator, does this mean for every single creator there needs to be a test like this?

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -40,3 +84,102 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+    def verbose_instance_factory(self):
+        project = RecruitProjectFactory()
+        card = AgileCardFactory(recruit_project=project)
+        return card
+
+    def setUp(self):
+        self.api_url = self.get_list_url()
+
+    @mock.patch.object(curriculum_tracking.activity_log_entry_creators, 'log_topic_competence_review_done')
+    @mock.patch.object(curriculum_tracking.activity_log_entry_creators, 'log_project_competence_review_done')
```
TheAbstract: The same object is being mocked under different names, is this intentional?

TheAbstract (approved):
Just some questions to make sure we're on the same page.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
EckardBerry: If we have a topic review then accidentally calling "log_project_competence_review_done" should not work cause it deals with projects and not topics.  If it does work then we have a big problem since the data we are passing to the function has somehow become corrupt.  So this is kinda like a sanity check.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
EckardBerry: It's not the same function being mocked, the one is "log_topic_competence_review_done" and the other is "log_project_competence_review_done".  I mock both to assert that the correct one does get called and therefore the other does not get called depending on the scenario.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -40,3 +84,102 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+        with self.assertRaises(AttributeError) as exception:
+            creators.log_project_competence_review_done(review1)
+        self.assertTrue(exception, AttributeError)
+        self.assertEqual(LogEntry.objects.count(), 0)
+
+
+class log_project_and_topic_competence_reviews_done_TESTS(APITestCase, APITestCaseMixin):
+    LIST_URL_NAME = "agilecard-list"
+    SUPPRESS_TEST_POST_TO_CREATE = True
+    FIELDS_THAT_CAN_BE_FALSEY = [
```
sheenarbw: You can remove `FIELDS_THAT_SHOULD BE_FALSEY` and the `verbose_instance_factory`. These should not be necessary because they are used for testing the GET endpoint. We should suppress that test here since we are already testing it elsewhere

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -40,3 +84,102 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+
+    def test_log_topic_creator_is_invoked_for_a_topic_review_else_attribute_error_is_raised(self):
+        review1 = TopicReviewFactory()
+        with self.assertRaises(AttributeError) as exception:
+            creators.log_project_competence_review_done(review1)
+        self.assertTrue(exception, AttributeError)
+        self.assertEqual(LogEntry.objects.count(), 0)
+
+
+class log_project_and_topic_competence_reviews_done_TESTS(APITestCase, APITestCaseMixin):
```
sheenarbw: This test is in the wrong file. There is one file that is dedicated to testing that the log reators get envoked correctly. And there is another file that is all about testing that they work.

Please put this test in the `envoked` file.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
sheenarbw: These are still inappropriate calls since we never instantiate `activity_log_entry_creators` instanes ourselves. We do not treat that as an object.

Rather:

```
@mock.patch(curriculum_tracking.activity_log_entry_creators.log_topic_competence_review_done)
@mock.patch(curriculum_tracking.activity_log_entry_creators.log_project_competence_review_done)
```

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -40,3 +84,102 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+        card = AgileCardFactory(recruit_project=project)
+        return card
+
+    def setUp(self):
+        self.api_url = self.get_list_url()
+
+    @mock.patch.object(curriculum_tracking.activity_log_entry_creators, 'log_topic_competence_review_done')
+    @mock.patch.object(curriculum_tracking.activity_log_entry_creators, 'log_project_competence_review_done')
+    def test_log_project_competence_review_invoked_from_api_endpoint_for_project_review(
+            self, proj_comp_review, topic_comp_review
```
sheenarbw: Why have you replaced the meaningful names with strange ones?

Wouldn't this code make more sense if it said:

```
def test_log_project_competence_review_invoked_from_api_endpoint_for_project_review(
            self, log_project_competence_review_done, log_topic_competence_review_done
            ):
```

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -40,3 +84,102 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+        super_user = UserFactory(is_superuser=True)
+        card = AgileCardFactory(content_item=ProjectContentItemFactory())
+        self.login(super_user)
+        response = self.client.post(
+                path=f'{self.api_url}{card.id}/add_review/',
+                data={"status": "NYC", "comments": "dammit"}
+            )
+
+        self.assertTrue(response.status_code, 200)
+        proj_comp_review.assert_called()
```
sheenarbw: what arguments was it called with? Was it called correctly?

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -40,3 +84,102 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+        topic_comp_review.assert_not_called()
+
+    @mock.patch.object(curriculum_tracking.activity_log_entry_creators, 'log_topic_competence_review_done')
+    @mock.patch.object(curriculum_tracking.activity_log_entry_creators, 'log_project_competence_review_done')
+    def test_log_topic_competence_review_invoked_from_api_endpoint_for_topic_review(
+            self, proj_comp_review, topic_comp_review
```
sheenarbw: names are also inconsistent here. 

When you have a choice between being consistent or being inconsistent then its worth trying to be consistent.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -40,3 +84,102 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+    def test_log_project_competence_review_invoked_from_api_endpoint_for_project_review(
+            self, proj_comp_review, topic_comp_review
+    ):
+
+        super_user = UserFactory(is_superuser=True)
+        card = AgileCardFactory(content_item=ProjectContentItemFactory())
+        self.login(super_user)
+        response = self.client.post(
+                path=f'{self.api_url}{card.id}/add_review/',
+                data={"status": "NYC", "comments": "dammit"}
```
sheenarbw:
```                
data={"status": NOT_YET_COMPETENT, "comments": "dammit"}
```

If a constant is defined somewhere it's better to import it and use it. It makes your test more explicit.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -40,3 +84,102 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+    @mock.patch.object(curriculum_tracking.activity_log_entry_creators, 'log_project_competence_review_done')
+    def test_log_project_competence_review_invoked_from_api_endpoint_for_project_review(
+            self, proj_comp_review, topic_comp_review
+    ):
+
+        super_user = UserFactory(is_superuser=True)
+        card = AgileCardFactory(content_item=ProjectContentItemFactory())
+        self.login(super_user)
+        response = self.client.post(
+                path=f'{self.api_url}{card.id}/add_review/',
```
sheenarbw:
```
path=f'{self.get_instance_url()}add_review/',
```

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -40,3 +84,102 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+        review1 = TopicReviewFactory()
+        creators.log_topic_competence_review_done(review1)
+        creators.log_topic_competence_review_done(review1)
+        self.assertEqual(LogEntry.objects.count(), 1)
+
+    def test_log_topic_creator_is_invoked_for_a_topic_review_else_attribute_error_is_raised(self):
+        review1 = TopicReviewFactory()
+        with self.assertRaises(AttributeError) as exception:
+            creators.log_project_competence_review_done(review1)
+        self.assertTrue(exception, AttributeError)
```
sheenarbw: You have asserted that the exception is an AttributeError twice. Once would be enough.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -23,6 +41,32 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+        review1 = RecruitProjectReviewFactory()
+        creators.log_project_competence_review_done(review1)
+        creators.log_project_competence_review_done(review1)
+        self.assertEqual(LogEntry.objects.count(), 1)
+
+    def test_log_project_creator_is_invoked_for_a_project_review_else_attribute_error_is_raised(self):
+        review1 = RecruitProjectReviewFactory()
+        with self.assertRaises(AttributeError) as exception:
+            creators.log_topic_competence_review_done(review1)
+        self.assertTrue(exception, AttributeError)
```
sheenarbw: You have asserted that the exception is an AttributeError twice. Once would be enough.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -40,3 +84,102 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+        card.topic_progress = TopicProgressFactory()
+        card.save()
+        self.login(super_user)
+        response = self.client.post(
+            path=f'{self.api_url}{card.id}/add_review/',
+            data={"status": "NYC", "comments": "dammit"}
+        )
+
+        self.assertTrue(response.status_code, 200)
+        topic_comp_review.assert_called()
```
sheenarbw: What was it called with?

Also, What should the logentry look like? There is no test here that checks that the attributes are corect. Eg, what should object_1 be equal to? Who is the effected user? Etc? Very important that we make sure the entry is properly created.

sheenarbw (changes requested):
The tests have a few rough edges still. And some stuff is missing

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
EckardBerry: If I remove FIELDS_THAT_SHOULD_BE_FALSEY I get this error:
```
AssertionError: False is not true : code_review_competent_since_last_review_request = 0. The value must be truthy
```

And if I remove the verbose instance factory I get this error:
```
NotImplementedError: use factory_boy (or other) factories to create an instance of the model under test
```

TheAbstract (approved):

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -1,14 +1,25 @@
...
+    TopicReview
+)
+from curriculum_tracking.constants import NOT_YET_COMPETENT
 from activity_log.models import LogEntry
 from curriculum_tracking import activity_log_entry_creators as creators
 from curriculum_tracking.constants import COMPETENT
+from test_mixins import APITestCase, APITestCaseMixin
+from .factories import (
+    UserFactory,
+    AgileCardFactory,
```
sheenarbw, line 17: we also have `from . imprt factories`
Do we need this import as well?

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -106,3 +117,107 @@ def test_add_review(self):
             start_url, data={"status": COMPETENT, "comments": "blah more stuff"}
         )
         self.assertEqual(LogEntry.objects.count(), 2)
+
+
+class log_project_and_topic_competence_reviews_done_TESTS(APITestCase, APITestCaseMixin):
+    LIST_URL_NAME = "agilecard-list"
+    SUPPRESS_TEST_POST_TO_CREATE = True
+    FIELDS_THAT_CAN_BE_FALSEY = [
+        "code_review_competent_since_last_review_request",
```
sheenarbw: We do test the agile card viewset elsewhere. Is there anything else that you should be suppressing here?  

A lot of this code can be removed if you set the right flags.

sheenarbw (changes requested):
Getting there. There are still some things that are pretty long compared to what they should be.

There are also lessons that you implemented in this PR that were not implemented in the other one. I'm not sure why.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -23,6 +27,31 @@ def test_that_only_one_entry_created_for_one_review(self):
         creators.log_project_competence_review_done(review2)
         self.assertEqual(LogEntry.objects.count(), 2)
 
+    def test_two_entries_created_for_two_reviews_with_second_done_within_debounce_period_by_another_user(self):
```
sheenarbw: Can we remove this test?  It looks like duplication.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -23,6 +27,31 @@ def test_that_only_one_entry_created_for_one_review(self):
...
+            timestamp=review1.timestamp,
+            comments=review1.comments,
+            recruit_project=review1.recruit_project,
+            reviewer_user=UserFactory()
+        )
+        creators.log_project_competence_review_done(review1)
+        creators.log_project_competence_review_done(review2)
+        self.assertEqual(LogEntry.objects.count(), 2)
+
+    def test_logging_the_same_entry_twice_only_produces_one_log_entry(self):
```
sheenarbw: This one too

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -1,14 +1,29 @@
...
+from curriculum_tracking.constants import NOT_YET_COMPETENT
 from activity_log.models import LogEntry
 from curriculum_tracking import activity_log_entry_creators as creators
 from curriculum_tracking.constants import COMPETENT
+from test_mixins import APITestCase, APITestCaseMixin
+from .factories import (
+    UserFactory,
+    AgileCardFactory,
+    RecruitProjectFactory,
+)
```
TheAbstract, line 19: These imports are redundant since we've already imported `factories` in line 4.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -84,28 +99,130 @@ def test_add_review(self):
...
-            start_url, data={"status": COMPETENT, "comments": "weee"}
+            add_review_url, data={"status": COMPETENT, "comments": "weee"}
         )
         self.assertEqual(response.status_code, 200)
 
         card.refresh_from_db()
 
-        self.assertEqual(LogEntry.objects.count(), 1)
-        entry = LogEntry.objects.first()
+        self.assertEqual(LogEntry.objects.count(), 2)
```
TheAbstract, line 109: Why would there be two log entries since only one review has been done by this point in the test?

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -146,3 +263,67 @@ def test_review_requested(self):
...
+        add_review_url = f"{self.get_instance_url(card.id)}add_review/"
+        response = self.client.post(
+            add_review_url, data={"status": NOT_YET_COMPETENT, "comments": "so sad"}
+        )
+        self.assertEqual(response.status_code, 200)
+
+        card.refresh_from_db()
+
+        log_card_moved_to_review_feedback.assert_called_with(card, actor_user)
+        log_card_moved_to_complete.assert_not_called()
```
TheAbstract: I think it would be useful to also indicate that the card is actually in the right place. Having an explicit check right before we check that the right logs are called, also makes the test easier to follow.
```suggestion
        self.assertEqual(response.status_code, 200)

        card.refresh_from_db()
        self.assertEqual(card.status, AgileCard.REVIEW_FEEDBACK)
        log_card_moved_to_review_feedback.assert_called_with(card, actor_user)
        log_card_moved_to_complete.assert_not_called()
```

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -146,3 +263,67 @@ def test_review_requested(self):
...
+        add_review_url = f"{self.get_instance_url(card.id)}add_review/"
+        response = self.client.post(
+            add_review_url, data={"status": COMPETENT, "comments": "woohoo"}
+        )
+        self.assertEqual(response.status_code, 200)
+
+        card.refresh_from_db()
+        project_review = RecruitProjectReview.objects.first()
+        log_card_moved_to_complete.assert_called_with(card, actor_user)
+        log_card_moved_to_review_feedback.assert_not_called()
```
TheAbstract: Same reasoning as above. Also, `project_review` isn't being used anywhere.
```suggestion
        card.refresh_from_db()
        self.assertEqual(card.status, AgileCard.COMPLETE)
        log_card_moved_to_complete.assert_called_with(card, actor_user)
        log_card_moved_to_review_feedback.assert_not_called()
```

TheAbstract (changes requested):

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
Samantha-Hampton: @TheAbstract when a project is reviewed there are 3 scenarios in terms of logging

1)
- competence review gets logged
- if negative review logging that the card has been moved to review feedback 

2)
-  competence review gets logged
- if positive review from trusted/superuser logging that the car has been moved to complete 

3)
-  competence review gets logged
- if positive review from non trusted/non superuser, nothing else will get logged

In the tests here I used a super user so there will always be two log entries ( 1) and 2) )
But I see now that I need to also test for case 3)

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
Samantha-Hampton: ah, yes, some bad pattern matching ;)

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
Samantha-Hampton: cool :)

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
Samantha-Hampton: actually I don't think I need extra tests, since that is really testing trusted reviewer permissions, which is not what we are testing here

TheAbstract (approved):

Andy-Nkumane (approved):

TheAbstract (approved):

Andy-Nkumane (approved):

sheenarbw (approved):

PR #323, opened 2022-01-11
tests mostly work. Thows a weird exception at the end

sheenarbw:
A lot of the changes were due to the black autoformatter. Which is odd.

The main things you want to look at here are:

Now if you put a breakpoint inside the recruit_project_setup_repository actor it will actually work.

There is still something funny going on around database connections so this is not perfect. But the test actually tests the right thing now

### DIFF
--- a/backend/curriculum_tracking/tests/test_api_views.py
+++ b/backend/curriculum_tracking/tests/test_api_views.py
@@ -1,23 +1,25 @@
-import long_running_request_actors
 from git_real.tests.factories import PullRequestFactory
 from git_real.constants import GITHUB_DATETIME_FORMAT
 from social_auth.tests.factories import SocialProfileFactory
-from rest_framework.test import APITestCase
+from rest_framework.test import APITestCase, APITransactionTestCase
 from test_mixins import APITestCaseMixin
 from core.tests.factories import UserFactory, TeamFactory
 from django.urls import reverse
 from . import factories
 from core.tests import factories as core_factories
 from datetime import timedelta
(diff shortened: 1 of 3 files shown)


backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -66,7 +68,7 @@ def verbose_instance_factory(self):
         return topic_progress
 
 
-class AgileCardViewsetTests(APITestCase, APITestCaseMixin):
+class AgileCardViewsetTests(APITransactionTestCase, APITestCaseMixin):
```
sheenarbw, line 71: This line is important

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -275,8 +291,15 @@ def get_repo_mock(
         url = f"{self.get_list_url()}{card.id}/setup_project_repo/"
         self.login(super_user)
         response = self.client.post(path=url, data={"card_id": card.id})
+
+        rabbitmq_broker.join(
+            recruit_project_setup_repository.queue_name, fail_fast=True
```
sheenarbw, line 296: This tells the fake worker to run. So it's in the same process.

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -275,8 +291,15 @@ def get_repo_mock(
...
+        )
+        worker.join()
+
         self.assertEqual(response.status_code, 200)
-        self.assertTrue(project.setup_repository.assert_called)
+        project.setup_repository.assert_called()
```
sheenarbw, line 301: Better assertion

PR #326, opened 2022-01-12
Card movement in out of complete/310

### LINKED ISSUE #310: Make sure that log_card_moved_to_complete gets called correctly and does what it should. Note that a card can be moved in and out of complete multiple times

EckardBerry:

## Description:

1. Wrote some test cases in backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py

## I solemnly swear that:


EckardBerry:
- Wrote test cases to ensure that one LogEntry with event_type "CARD_MOVED_TO_COMPLETE" is created for a card with a completed status.
- Wrote a test case to ensure that two LogEntry instances are created if the creation of the 2nd instance happens more than 120 seconds after the creation of the 1st.
- Trying to mock time.sleep or datetime.date or datetime.time or timezone.now() proved unfruitful.  Keep running into the same error that although you can mock Python built-in functions, some functions like time cannot be overwritten.

EckardBerry:
So I've showed through the tests that only one LogEntry instance with event_type CARD_MOVED_TO_COMPLETE is created for a single call to debounce_create and that two LogEntry instances with event_type CARD_MOVED_TO_COMPLETE are created
for two calls to debounce_create if more than 120 seconds has elapsed from the time of the 1st call to debounce_create.

What I don't know is what would be acceptable criteria for when we want to create and store LogEntry instances with event_type CARD_MOVED_TO_COMPLETE.  There are many scenarios where we would want to always store such instances and also many other scenarios where we would not want to do so?

TheAbstract (changes requested):
We just have this error
```cmd
FAIL: test_card_to_complete_creates_one_log_entry (curriculum_tracking.tests.test_activity_log_entry_creators_envoked.log_card_moved_to_complete_Tests)
  File "/Users/umuzi/Tilde/backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py", line 84, in test_card_to_complete_creates_one_log_entry
    self.assertEqual(LogEntry.objects.first().object_1_id, card.id)
AssertionError: 185 != 120

Ran 225 tests in 82.764s

FAILED (failures=1)
```

sheenarbw (changes requested):
Tests still dont look complete. 

We need to do 2 things:
1. Test that the creator works. This you have done
2. test that the creater gets called when it is meant to be called. This has not been attempted. 

We dont just need the function to exist and work in isolation. We need to know that Tilde calls the function cporrectly. An unused function is useless. We need to make sure it is used.


Also, I think you wrote your tests in a funny way.

test_log_creators_envoked should test that they are envoked. As in make sure that they are called. You did not test this at all.
test_activity_log_entry_creators should test that the functions themselves work. Your code tests that the functions work and should be there.


We need all the tests to exist and be in the correct files.

EckardBerry:
Please see the test "test_log_project_competence_review_invoked_from_api_endpoint_for_project_review (line 210)" which shows that the function "log_card_moved_to_complete" is not called from the api-endpoint, should it be called from the api end point?  If so, why should there be a separate test when further up (line 166) it was manually called?

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -1,14 +1,21 @@
...
+from django.test import TestCase
+import curriculum_tracking
 from test_mixins import APITestCaseMixin
 from . import factories
 from core.tests.factories import UserFactory
+from .factories import AgileCardFactory, RecruitProjectFactory, ProjectContentItemFactory
 from curriculum_tracking.models import AgileCard, ContentItem
 from activity_log.models import LogEntry
-from curriculum_tracking import activity_log_entry_creators as creators
+from curriculum_tracking import activity_log_entry_creators as creators, activity_log_entry_creators
```
TheAbstract: What's going on with this import? We are importing `activity_log_entry_creators` as `creators` and then importing the same thing again with the original name. Why?

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -59,8 +66,162 @@ def test_start_project(self):
...
+        )
+        card.status = AgileCard.COMPLETE
+        card.save()
+
+        mock_now.return_value = timezone.now() + timezone.timedelta(seconds=121)
+        creators.log_card_moved_to_complete(card, actor_user)
+        self.assertTrue(card.status == AgileCard.COMPLETE)
+
+        log_entries = [entry.event_type.name for entry in LogEntry.objects.all()].count('CARD_MOVED_TO_COMPLETE')
+        self.assertTrue(log_entries, 3)
```
TheAbstract, line 171: This card has only two COMPLETE reviews, but at the end the assert is for 3 logs of `CARD_MOVED_TO_COMPLETE` (and it passes 😕 ). Either `log_entries` isn't getting the write count or the log isn't working properly. This is meant to only have two COMPLETE entries logged, right?

TheAbstract (changes requested):

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
EckardBerry: When I mock activity_log_entry_creators I cannot do so using the keyword 'creators', I have to make use of activity_log_entry_creators and therefore I have to import it.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
EckardBerry: There should be 3 entries because the test starts with a review to complete status then review feedback status and then complete status again.  The 2nd complete status occurs after the debounce period of 120 seconds and therefore 3 entries should be created.

TheAbstract (approved):

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -1,14 +1,22 @@
...
+from django.test import TestCase
+import curriculum_tracking
 from test_mixins import APITestCaseMixin
 from . import factories
 from core.tests.factories import UserFactory
+from .factories import AgileCardFactory, RecruitProjectFactory, ProjectContentItemFactory
 from curriculum_tracking.models import AgileCard, ContentItem
 from activity_log.models import LogEntry
 from curriculum_tracking import activity_log_entry_creators as creators
+from curriculum_tracking import activity_log_entry_creators
```
sheenarbw, line 13: why are we importing these twice? look at line 12. Remember all the trouble that imports caused before? 

If something is already available then you probably dont need to import it again

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -1,14 +1,22 @@
 """for each of the log entry creators in curriculum_tracking.activity_log_entry_creators, make sure it is called when it should be and creates the correct log entries
 """
 from rest_framework.test import APITestCase
-
+from django.test import TestCase
+import curriculum_tracking
 from test_mixins import APITestCaseMixin
 from . import factories
 from core.tests.factories import UserFactory
+from .factories import AgileCardFactory, RecruitProjectFactory, ProjectContentItemFactory
```
sheenarbw, line 9: Why are you importing these when they are already available? You can call `factories.AgileCardFactory` whenever you want

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -1,14 +1,22 @@
...
 from curriculum_tracking.models import AgileCard, ContentItem
 from activity_log.models import LogEntry
 from curriculum_tracking import activity_log_entry_creators as creators
+from curriculum_tracking import activity_log_entry_creators
 from curriculum_tracking.constants import COMPETENT
+from django.utils import timezone
+from mock import patch
```
sheenarbw, line 16: usually we just say `@mock.patch`. It's a buit more explicit than just `patch`

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -59,8 +67,162 @@ def test_start_project(self):
...
+        actor_user = UserFactory(is_superuser=True)
+        card = factories.AgileCardFactory(
+            status=AgileCard.COMPLETE,
+            content_item=factories.ProjectContentItemFactory(
+                project_submission_type=ContentItem.LINK, template_repo=None
+            ),
+        )
+        card.save()
+
+        creators.log_card_moved_to_complete(card, actor_user)
```
sheenarbw, line 83: This test is in the wrong file.

You just envoked the function. You called it.

This file is all about making sure things are called when they are supposed to be called. This file tests that things are envoked when they need to be. 

Please move the code.

Or please choose a new and meaningful standard and then make all the existing action creator tests fit that standard.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -59,8 +67,162 @@ def test_start_project(self):
...
+
+        card.save()
+        creators.log_card_moved_to_complete(card, actor_user)
+        self.assertTrue(card.status == AgileCard.COMPLETE)
+        self.assertEqual(LogEntry.objects.count(), 1)
+
+        card.status = AgileCard.REVIEW_FEEDBACK
+        card.save()
+        creators.log_card_moved_to_review_feedback(card, actor_user)
+        self.assertTrue(card.status == AgileCard.REVIEW_FEEDBACK)
```
sheenarbw, line 108: There is a lot of stuff going on here. Can't you just do this:

```
actor_user = make_the_user
card = make_the_card

creators.log_card_moved_to_review_feedback(card, actor_user)
creators.log_card_moved_to_review_feedback(card, actor_user)

self.assertTrue(log_entries, 1)
```

Why should the creator care about any of that extra stuff? If we added a new state to cards (eg in_trusted_review)  then this test should still be totally relevent and correct. 

Always think about least privilage. With code things are always on a need to know basis. Every piece of code should know as little as possible about the surrounding system.  That's called loose coupling.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -59,8 +67,162 @@ def test_start_project(self):
...
+
+        nyc_review = factories.RecruitProjectReviewFactory(
+            recruit_project=card.recruit_project,
+            timestamp=initial_complete_review.timestamp + timezone.timedelta(seconds=20),
+            reviewer_user=actor_user
+        )
+        card.status = AgileCard.REVIEW_FEEDBACK
+        card.save()
+        creators.log_card_moved_to_review_feedback(card, actor_user)
+
```
sheenarbw, line 157: Again, there is a lot going on here. What can we strip away?

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -59,8 +67,162 @@ def test_start_project(self):
...
+        self.assertTrue(card.status == AgileCard.COMPLETE)
+
+        log_entries = [entry.event_type.name for entry in LogEntry.objects.all()].count('CARD_MOVED_TO_COMPLETE')
+        self.assertTrue(log_entries, 3)
+
+
+class log_card_moved_to_complete_api_view_Tests(APITestCase, APITestCaseMixin):
+    LIST_URL_NAME = "agilecard-list"
+    SUPPRESS_TEST_POST_TO_CREATE = True
+    FIELDS_THAT_CAN_BE_FALSEY = [
```
sheenarbw, line 177: Why do we need this? Aren't we testing the rest of this viewset elsewhate? Should we not be suppressing something else?

Less is more.

sheenarbw (changes requested):
There is a lot of extra stuff going on in here.

PR #344, opened 2022-01-26
Log card review request called/312

### LINKED ISSUE #312: Make sure that log_card_review_requested gets called correctly and does the right thing. Note that a review can be requested on a single card multiple times.

Samantha-Hampton:

## Description:

1) Wrote tests  to check if the activity logging on requesting a review on a card worked as it should, and amended the request_review function (in curriculum_tracking, AgileCardViewset) as needed.

2) From writing a test in test_activity_log_entry_creators_envoked.py that failed I then saw that in api_views.py (curriculum tracking) the request_review function (in the class AgileCardViewset) did not call log_creators (like the add_review function does).

3) Moved tests related to permissions to test_api_views and amended tests in test_activity_log_entry_creators_envoked to only the tests needed for the logging

4) Created new test class for tests related to review request permissions in test_api_views, and wrote a setUp function to avoid code repetition in the tests.

## I solemnly swear that:

### DIFF
--- a/backend/curriculum_tracking/api_views.py
+++ b/backend/curriculum_tracking/api_views.py
@@ -356,8 +356,7 @@ class AgileCardViewset(viewsets.ModelViewSet):
             card.finish_topic()
         else:
             raise Http404
-        log_creators.log_card_review_requested(card, card.assignees) #**** what to use as actor_user here? 
-        #card.assignee is not right
+        log_creators.log_card_review_requested(card=card, actor_user=request.user) 
         # comes from here: def log_card_review_requested(card, actor_user) in activity_log_entry_creators.py
         card.refresh_from_db()
         assert (
--- a/backend/curriculum_tracking/api_views.py
+++ b/backend/curriculum_tracking/api_views.py
@@ -356,6 +356,9 @@ class AgileCardViewset(viewsets.ModelViewSet):
             card.finish_topic()
         else:
             raise Http404
+        log_creators.log_card_review_requested(card, card.assignees) #**** what to use as actor_user here? 
+        #card.assignee is not right
+        # comes from here: def log_card_review_requested(card, actor_user) in activity_log_entry_creators.py
         card.refresh_from_db()
         assert (
             card.status == models.AgileCard.IN_REVIEW
--- a/backend/curriculum_tracking/api_views.py
+++ b/backend/curriculum_tracking/api_views.py
@@ -350,14 +350,16 @@ class AgileCardViewset(viewsets.ModelViewSet):
     )
     def request_review(self, request, pk=None):
         card: models.AgileCard = self.get_object()
+
         if card.recruit_project:
             card.recruit_project.request_review(force_timestamp=timezone.now())
         elif card.topic_progress:
             card.finish_topic()
         else:
             raise Http404
-        log_creators.log_card_review_requested(card=card, actor_user=request.user) 
-        # comes from here: def log_card_review_requested(card, actor_user) in activity_log_entry_creators.py
+
+        log_creators.log_card_review_requested(card=card, actor_user=request.user)
+
         card.refresh_from_db()
         assert (
             card.status == models.AgileCard.IN_REVIEW
--- a/backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
+++ b/backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
@@ -113,7 +113,7 @@ class log_card_review_requested_Tests(APITestCase, APITestCaseMixin):
     SUPPRESS_TEST_POST_TO_CREATE = True
     SUPPRESS_TEST_GET_LIST = True
 
-    def test_review_requested_authorised_user(self):
+    def test_review_requested(self):
         actor_user = UserFactory(is_superuser=True, is_staff=True)
         content_item = factories.ProjectContentItemFactory(
             project_submission_type=ContentItem.LINK, template_repo=None
@@ -146,54 +146,3 @@ class log_card_review_requested_Tests(APITestCase, APITestCaseMixin):
         self.assertEqual(entry.object_1, card.recruit_project)
         self.assertEqual(entry.object_2, None)
         self.assertEqual(entry.event_type.name, creators.CARD_REVIEW_REQUESTED)
-
-    def test_review_requested_assignee(self):
-
-        content_item = factories.ProjectContentItemFactory(
-            project_submission_type=ContentItem.LINK, template_repo=None
-        )
-        card = factories.AgileCardFactory(
-            status=AgileCard.READY,
-            recruit_project=factories.RecruitProjectFactory(content_item=content_item),
-            content_item=content_item,
-        )
-        self.login(card.assignees.first())
-        card.start_project()
-        self.assertEqual(card.status, AgileCard.IN_PROGRESS)
-
-        request_review_url = f"{self.get_instance_url(card.id)}request_review/"
-        response = self.client.post(request_review_url)
-
-        self.assertEqual(response.status_code, 200)
-
-        card.refresh_from_db()
-        self.assertEqual(card.status, AgileCard.IN_REVIEW)
-
-        # sanity check
-        self.assertEqual(card.assignees.count(), 1)
-
-        self.assertEqual(LogEntry.objects.count(), 1)
-        entry = LogEntry.objects.first()
-
-        self.assertEqual(entry.actor_user, card.assignees.first())
-        self.assertEqual(entry.effected_user, card.assignees.first())
-        self.assertEqual(entry.object_1, card.recruit_project)
-        self.assertEqual(entry.object_2, None)
-        self.assertEqual(entry.event_type.name, creators.CARD_REVIEW_REQUESTED)
-
-    def test_review_requested_non_assignee(self):
-        actor_user = UserFactory()
-        content_item = factories.ProjectContentItemFactory(
-            project_submission_type=ContentItem.LINK, template_repo=None
-        )
-        card = factories.AgileCardFactory(
-            status=AgileCard.IN_PROGRESS,
-            recruit_project=factories.RecruitProjectFactory(content_item=content_item),
-            content_item=content_item,
-        )
-        self.login(actor_user)
-        self.assertEqual(card.status, AgileCard.IN_PROGRESS)
-
-        request_review_url = f"{self.get_instance_url(card.id)}request_review/"
-        response = self.client.post(request_review_url)
-        self.assertEqual(response.status_code, 403)
--- a/backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
+++ b/backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
@@ -181,8 +181,8 @@ class log_card_review_requested_Tests(APITestCase, APITestCaseMixin):
         self.assertEqual(entry.object_2, None)
         self.assertEqual(entry.event_type.name, creators.CARD_REVIEW_REQUESTED)
 
-    def test_review_requested_non_authorised_user(self):
-        actor_user = UserFactory(is_superuser=False)
+    def test_review_requested_non_assignee(self):
+        actor_user = UserFactory()
         content_item = factories.ProjectContentItemFactory(
             project_submission_type=ContentItem.LINK, template_repo=None
         )
--- a/backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
+++ b/backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
@@ -119,14 +119,15 @@ class log_card_review_requested_Tests(APITestCase, APITestCaseMixin):
                 project_submission_type=ContentItem.LINK, template_repo=None
             )
         card = factories.AgileCardFactory(
-            status=AgileCard.READY,
+            status=AgileCard.READY, 
+
             
             recruit_project=factories.RecruitProjectFactory(content_item=content_item),
             content_item=content_item
             
         )
-        card.start_project()
-        self.login(actor_user)
+        self.login(actor_user) #do we not login first before starting project?
+        card.start_project() #means card is now IP
 
         request_review_url = f"{self.get_instance_url(card.id)}request_review/"
         response = self.client.post(request_review_url)
--- a/backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
+++ b/backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
@@ -30,7 +30,6 @@ class log_card_started_Tests(APITestCase, APITestCaseMixin):
         start_url = f"{self.get_instance_url(card.id)}start_project/"
         response = self.client.post(start_url)
         self.assertEqual(response.status_code, 200)
-        
 
         card.refresh_from_db()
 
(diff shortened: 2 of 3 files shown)


backend/curriculum_tracking/api_views.py
```diff
@@ -356,6 +356,8 @@ def request_review(self, request, pk=None):
             card.finish_topic()
         else:
             raise Http404
+        log_creators.log_card_review_requested(card=card, actor_user=request.user) 
+        # comes from here: def log_card_review_requested(card, actor_user) in activity_log_entry_creators.py
```
EckardBerry: You will need to not commit and push on notes that you leave yourself.  I get what you are doing, I also do that but it does actually confuse others which need to read and understand your code.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -106,3 +106,45 @@ def test_add_review(self):
...
+            recruit_project=factories.RecruitProjectFactory(content_item=content_item),
+            content_item=content_item
+            
+        )
+        self.login(actor_user) #do we not login first before starting project?
+        card.start_project() #means card is now IP
+
+        request_review_url = f"{self.get_instance_url(card.id)}request_review/"
+        response = self.client.post(request_review_url)
+        assert response.status_code == 200, response.data
```
EckardBerry: Because we want to use proper unit testing conventions please rather use self.assertEqual and not just assert, here is an example:
   ```
   ```

Also, there is some other tests here (not yours) which also makes use of just assert and not self.assert, can you please fix those and make sure the tests still pass.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -106,3 +106,45 @@ def test_add_review(self):
             start_url, data={"status": COMPETENT, "comments": "blah more stuff"}
         )
         self.assertEqual(LogEntry.objects.count(), 2)
+
+
+class log_card_review_requested_Tests(APITestCase, APITestCaseMixin):
+    LIST_URL_NAME = "agilecard-list"  #based on class of api-viewset
```
EckardBerry: Try not to commit comments like these, comments like that is for yourself so that you can understand and keep track of things, they should not form part of the codebase.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -106,3 +106,45 @@ def test_add_review(self):
...
+    SUPPRESS_TEST_POST_TO_CREATE = True
+    SUPPRESS_TEST_GET_LIST = True
+
+    def test_review_requested(self):
```
EckardBerry, line 116: Your test names will need to be much more descriptive than this.  We not only want to know if the functionality of the api-view works we also need to know what are we supposed to get back.  For instance:

```
test_review_request_from_api_returns_card_in_review_status
```

I deliberately chose a long name to make the intentions of the test as clear as I can.  Although this name is perhaps not the best and can be improved upon, it does serve as an example of making the test name such so that the intentions of the test are evidently clear.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -106,3 +106,45 @@ def test_add_review(self):
...
+            )
+        card = factories.AgileCardFactory(
+            status=AgileCard.READY, 
+
+            
+            recruit_project=factories.RecruitProjectFactory(content_item=content_item),
+            content_item=content_item
+            
+        )
+        self.login(actor_user) #do we not login first before starting project?
```
EckardBerry: The inline comments should not form part of the code base.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -106,3 +106,45 @@ def test_add_review(self):
...
+        card.start_project() #means card is now IP
+
+        request_review_url = f"{self.get_instance_url(card.id)}request_review/"
+        response = self.client.post(request_review_url)
```
EckardBerry, line 131: It is good that you assert that you get a 200 response code, but you should also assert that the cards status is now 'IN_REVIEW'.  I know that the view already does that but you need to do that in your test as well.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -106,3 +106,45 @@ def test_add_review(self):
...
+        assert response.status_code == 200, response.data
+
+        card.refresh_from_db()
+
+        # sanity check
+        assert card.assignees.count() == 1
```
EckardBerry: I am not sure why you check how many assignees the card has, is it relevant to check that in this test?

EckardBerry (changes requested):
Hello

Very good effort on this, I remember the first time I worked on Django I think Sheena sent it back with requested changes like 30 or 40 times, but this is a lot better than my 1st attempts.

Here are some remarks other than the inline comments I've made:

1. When you write a description for your ticket, it should be more about what
   we are trying to achieve (The big picture) so that whoever reviews the ticket
   has a global idea of what is supposed to happen.
   
2. You have to also please remember to tick those boxes underneath the part that
   reads "I solemnly swear that:"
   
3. Also be sure to fill in the part about related issues, this ticket has several
   related issues.
   
4. Because we want to use proper unit testing conventions please rather use 
   self.assertEqual and not just assert, here is an example:
   ```

5. You will need some more tests as well, notice permission classes of the view, it shows that only a person assigned to a card should be able to request a review on it.  So have a test which shows that for a non-assignee you get a 403 response.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
sheenarbw: This shows a misunderstanding.  Logging in is only needed when interacting with the api. `card.start_project` doesn't touch apis. 

In terms of comments - if they are just for you then they should be removed. If they are likely to be useful to other people then they should be in the code.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -106,3 +106,45 @@ def test_add_review(self):
...
+    SUPPRESS_TEST_GET_LIST = True
+
+    def test_review_requested(self):
+        actor_user = UserFactory(is_superuser=True, is_staff=True)
+        content_item=factories.ProjectContentItemFactory(
+                project_submission_type=ContentItem.LINK, template_repo=None
+            )
+        card = factories.AgileCardFactory(
+            status=AgileCard.READY, 
+
```
sheenarbw: There is a lot of blank space. Are you using `black`? I think it would clean this up automatically.

backend/curriculum_tracking/api_views.py
sheenarbw: Also you can usually use vscode to just tell you where things are defined. Proper use of IDEs makes these kinds of comments unnecessary. You can just choose "go to definition"

Samantha-Hampton: Noted - the excitement of getting a new test to pass got the better of me, will scrutinise more closely in future ;)

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
Samantha-Hampton: I can see how it's good to have a very clear name of test intention. I had followed the convention of the first test written for another logging test class already there (i.e from the log_card_start_Tests class). So now I'm confused as to whether to follow the convention or amend the name of the test.....

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
Samantha-Hampton: haha, again this was me pattern matching from a previous test. Perhaps the sanity check was not meant to be kept in the code base? but I guess just a double check that all is well? I actually think I agree, that it is odd. Is there ever an occasion that a card would have more than one assignee....

Samantha-Hampton:
> Hello
> 
> Very good effort on this, I remember the first time I worked on Django I think Sheena sent it back with requested changes like 30 or 40 times, but this is a lot better than my 1st attempts.
> 
> Here are some remarks other than the inline comments I've made:
> 
> 1. When you write a description for your ticket, it should be more about what
>    we are trying to achieve (The big picture) so that whoever reviews the ticket
>    has a global idea of what is supposed to happen.
> 2. You have to also please remember to tick those boxes underneath the part that
>    reads "I solemnly swear that:"
> 3. Also be sure to fill in the part about related issues, this ticket has several
>    related issues.
> 4. Because we want to use proper unit testing conventions please rather use
>    self.assertEqual and not just assert, here is an example:
>    ```
>    self.assertEqual(response.status_code, 200) 
>    ```
> 5. You will need some more tests as well, notice permission classes of the view, it shows that only a person assigned to a card should be able to request a review on it.  So have a test which shows that for a non-assignee you get a 403 response.


Very helpful (and kind!) feedback thanks Eckard - working on the code now - will pop you a message if I have any questions :)

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
Samantha-Hampton: done :)

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
Samantha-Hampton: noted, will check before pushing next time :)

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
Samantha-Hampton: have removed

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
Samantha-Hampton: have added this check

Samantha-Hampton:
Cleaned up code and added additional tests which now pass

EckardBerry (approved):

sheenarbw (changes requested):
just a few remarks

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
Samantha-Hampton: Ah, I see what happened, I pushed the changes to the testing file I made but not to the api_views file (which weren't so much changes as removing comments and breakpoints made for my own reference) 
Will push those changes.
Will check into using black too. 

In terms of this:
"This shows a misunderstanding. Logging in is only needed when interacting with the api. card.start_project doesn't touch apis."
Was from pattern matching on other tests - but I see now that I can set the card to being IP already when it's created, so will do that
Thanks for the feedback :)

Samantha-Hampton: @sheenarbw 
I tried writing the test without card.start_project() (by creating a card that was already in progress)
but got an error according to an assert statement in the model file, i.e:

File "/Users/sam/Documents/Umuzi/Products/Tilde/backend/curriculum_tracking/models.py", line 709, in request_review
    assert (
AssertionError: cannot request a review if the project isn't started - P: something awesome part 173 - foo.681@example.com

So leaving as it was for now so that the tests don't fail, but keen to see how else it could be done?

Andy-Nkumane (approved):

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -106,3 +106,94 @@ def test_add_review(self):
...
+            project_submission_type=ContentItem.LINK, template_repo=None
+        )
+        card = factories.AgileCardFactory(
+            status=AgileCard.READY,
+            recruit_project=factories.RecruitProjectFactory(content_item=content_item),
+            content_item=content_item,
+        )
+        self.login(actor_user)
+        card.start_project()
+        self.assertEqual(card.status, AgileCard.IN_PROGRESS)
```
TheAbstract, line 128: @Samantha-Hampton The only reason the login is need is because of the post you're making later in line 131, the assert for the `card.status` isn't affected by this and will pass even without the login.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -106,3 +106,94 @@ def test_add_review(self):
...
+        self.assertEqual(LogEntry.objects.count(), 1)
+        entry = LogEntry.objects.first()
+
+        self.assertEqual(entry.actor_user, card.assignees.first())
+        self.assertEqual(entry.effected_user, card.assignees.first())
+        self.assertEqual(entry.object_1, card.recruit_project)
+        self.assertEqual(entry.object_2, None)
+        self.assertEqual(entry.event_type.name, creators.CARD_REVIEW_REQUESTED)
+
+    def test_review_requested_non_authorised_user(self):
```
TheAbstract: Looking at this test my first comment was this;
> Not sure I get the point of this test, are we saying that only review requests from superusers will be logged? Looking at 
```
def log_project_competence_review_done(review):
``` 
> I don't think this is the case.

Then I realised you're testing that for someone who isn't an assignee nothing should be logged since they shouldn't be able to request reviews on other peoples stuff. Maybe this could be made more explicit, since we have `test_review_requested_assignee` right before, this makes it clear what the test is for: 
```suggestion
    def test_review_requested_by_non_assignee(self):
```

Also `is_superiser` is `False` by default for `UserFactory(is_superuser=False)` so these are equivalent
```
actor_user = UserFactory(is_superuser=False)
actor_user = UserFactory()
```
This `UserFactory()` is any regular user, not a super user and not staff.

Samantha-Hampton: @TheAbstract thank you for the feedback :)

Samantha-Hampton:
@TheAbstract and @sheenarbw 
I suppose I'm keen to understand if the three tests created are enough?
The only other idea I had was somehow to test that additional log entries are created when review is requested additional times after NYCs issued - or not necessary?

TheAbstract (approved):
Re this: https://github.com/Umuzi-org/Tilde/pull/344#issuecomment-1034838670

Minimally I usually think of valid and invalid use-cases; two questions in mind should be,
- What should happen when this is used the way we intended?
- What should happen when this is not used the way we intended?

Sometimes there are also other tests for how the thing should be used or interact with other parts of Tilde. Sheena generally provides guidance about that.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -106,3 +106,94 @@ def test_add_review(self):
...
+        self.assertEqual(LogEntry.objects.count(), 1)
+        entry = LogEntry.objects.first()
+
+        self.assertEqual(entry.actor_user, card.assignees.first())
+        self.assertEqual(entry.effected_user, card.assignees.first())
+        self.assertEqual(entry.object_1, card.recruit_project)
+        self.assertEqual(entry.object_2, None)
+        self.assertEqual(entry.event_type.name, creators.CARD_REVIEW_REQUESTED)
+
+    def test_review_requested_non_assignee(self):
```
sheenarbw: What are these tests for? I think permissions were tested elsewhere already, weren't they? 

And if this test is needed, then surely the permission tests should not live in the log creator test files?

sheenarbw (changes requested):
Looks like stuff works, but some tests are ain a funny place. 

If you are testing the permissions on the endpoints, then it would be good to test those how all the other permissions get tested.  Look for other tests that have this
```
    LIST_URL_NAME = "agilecard-list"
```

Look at how other permissions are tested.

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
sheenarbw: You can check the code that raises the exception. See what conditions it checks?

TheAbstract (approved):

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -156,6 +162,95 @@ def _test_set_due_time_permissions(self, card, get_progress):
...
+        request_review_url = f"{self.get_instance_url(card.id)}request_review/"
+        response = self.client.post(request_review_url)
+
+        self.assertEqual(response.status_code, 403)
+        self.assertEqual(response.data["detail"].code, "permission_denied")
+
+        card.refresh_from_db()
+        self.assertEqual(card.status, AgileCard.IN_PROGRESS)
+
+        # test using not the cards assignee
```
sheenarbw: It looks like you are testing 4 different situations here. Should they all be in one function? Or should it be 4 functions? 
It also looks like ther eis a lot of repeated functionality. Maybe a lot of the setup could go in a setUp method?

sheenarbw (changes requested):
Getting there. There is some unclean stuff happening in the tests though

TheAbstract (approved):

sheenarbw (approved):

PR #357, opened 2022-02-01
Set due dates in bulk updated/106

### LINKED ISSUE #106: Create api endpoint for setting due dates in bulk
Put it inside curriculum_tracking/api_views.py in ManagmentCommands unless you can think of a better place. 

Inputs:
- team
- content_item_id
- flavors

make sure you select the right cards. flavours must match. look at the FlavourMixin class.

Make use of AgileCard.set_due_time 

Permissions: People with PERMISSION_MANAGE_CARDS on the team can do this. Deny anyone else 

Tests can go in core/tests/test_api_views.py

EckardBerry:
Related issues: #106 

## Description:

1. Added some code to the ManagmentActions view, see backend/curriculum_tracking/api_views.py
2. Added some code to the SetDueDate serializer, see backend/curriculum_tracking/serializers.py 
3. Added some tests to test_api_views, see backend/core/tests/test_api_views.py 

## I solemnly swear that:


backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -467,3 +472,142 @@ def test_team1_users_cannot_view_team2_burndown_snapshot_objects(self):
...
+        self.login(self.super_user)
+        url = f'{self.get_instance_url(pk=self.blue_team.id)}bulk_set_due_dates/'
+        response = self.client.post(path=url, format='json', data={
+            'due_time': '2021-12-03T14:17', 'content_item': card.content_item.id, 'team': str(self.blue_team.id),
+            'flavours': [card.flavours.first().id, card.flavours.last().id]
+        })
+
+        self.assertEqual(response.status_code, 200)
+        card.refresh_from_db()
+        self.assertTrue(card.due_time, '2021-12-03T14:17')
```
TheAbstract: Shouldn't this be `assertEqual`?

TheAbstract (changes requested):
Everything looks fine, and those other tests aren't an issue anymore, nick work 👍🏾.

There's just one thing that seems odd, in the tests there's something like this;
```
card_time = ?
expected_time = x
assertTrue(card_time, expected_time)
```
shouldn't this be like this
```
assertEqual(card_time, expected_time)
```
Is the issue that `expected_time` is just a string and `card_time` is a date time object? If so couldn't we just convert `expected_time` into a date time object then assert that they're equal? 

Strings are "falsey" so in any of the tests, all of these pass
```py
self.assertTrue(card.due_time, '2021-12-03T14:17') # This is the correct time
self.assertTrue(card.due_time, '2021-10-03T14:18')
self.assertTrue(card.due_time, '007')
```
I checked this by add the these to `test_super_user_can_bulk_set_due_dates_for_a_team`

backend/curriculum_tracking/tests/test_api_views.py
EckardBerry: You're right, I fixed the assertion.

TheAbstract: This still has the `assertTrue()` which will pass regardless of what the second argument is. For example
```py
self.assertTrue(card.due_time, 'not even a date')
```

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -467,3 +472,142 @@ def test_team1_users_cannot_view_team2_burndown_snapshot_objects(self):
...
+        self.assertEqual(response.status_code, 200)
+
+        team_cards = get_team_cards(self.red_team, card_1.content_item)
+        self.assertIn(card_1, team_cards)
+        self.assertIn(card_2, team_cards)
+        self.assertNotIn(card_3, team_cards)
+
+        for card in team_cards:
+            card.refresh_from_db()
+            self.assertTrue(card.due_time, '2022-01-20T12:17')
```
TheAbstract: Same thing here, this will pass
```py
self.assertTrue(card.due_time, 'not even a date')
```

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -467,3 +472,142 @@ def test_team1_users_cannot_view_team2_burndown_snapshot_objects(self):
...
+        response = self.client.post(path=url, format='json', data={
+            'due_time': '2022-01-20T12:17', 'content_item': card_1.content_item.id, 'team': str(self.red_team.id),
+            'flavours': [card_1.flavours.first().id]
+        })
+        self.assertEqual(response.status_code, 200)
+
+        card_1.refresh_from_db()
+        card_2.refresh_from_db()
+        card_3.refresh_from_db()
+        self.assertTrue(card_1.due_time, '2022-01-20T12:17')
```
TheAbstract: Same here, this will pass
```py
self.assertTrue(card_1.due_time, 'not even a date')
```

TheAbstract (changes requested):
See https://github.com/Umuzi-org/Tilde/pull/357#pullrequestreview-871847264 for what I mean about the comments below, and why those lines are not actually checking the dates.

TheAbstract (approved):

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -136,28 +137,31 @@ def _test_set_due_time_permissions(self, card, get_progress):
 
         self.login(recruit)
 
-        response = self.client.post(url, data={"due_time": due_time_1})
+        response = self.client.post(
+            url, data={"due_time": due_time_1, "content_item": card.content_item.id}
+        )
+
         self.assertEqual(response.status_code, 200)
```
Samantha-Hampton, line 245: I'm probably misunderstanding something, so apologies if so, but looking at the flow of this test it seems like a recruit is logging in and then setting a due time on one of their own content items - and it being acceptable thing to do.
I didn't know that a recruit was meant to be able to do this?

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -411,59 +408,258 @@ def test_users_can_view_their_own_reviewer_trusts(self):
...
+        card = AgileCardFactory(recruit_project=project)
+        card.reviewers.add(self.user_one_blue)
+        card.assignees.add(self.user_one_red)
+        self.login(self.user_one_blue)
+        url = f"{self.get_instance_url(pk=self.red_team.id)}bulk_set_due_dates/"
+        response = self.client.post(
+            path=url,
+            format="json",
+            data={
+                "due_time": "2021-12-03T14:17",
```
Samantha-Hampton, line 628: tested that the "due_time" in this test could also literally be set to any string, but my assumption is that this doesn't matter here because we are testing that the due dates CAN'T in fact be set at all.

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -411,59 +408,258 @@ def test_users_can_view_their_own_reviewer_trusts(self):
...
+                "due_time": due_date,
+                "content_item": card.content_item.id,
+                "team": str(self.blue_team.id),
+                "flavours": [card.flavours.first().id, card.flavours.last().id],
+            },
+        )
+
+        self.assertEqual(response.status_code, 200)
+        card.refresh_from_db()
+        date_expected = datetime.strptime(due_date, "%Y-%m-%dT%H:%M").replace(
```
Samantha-Hampton: I did some online searching on the Django timezone utils and found the following that could be a different/more efficient way of doing:

if you add this to the imports:

from django.utils.dateparse import parse_datetime

and then replace this line with:

date_expected = parse_datetime(due_date).replace(tzinfo=timezone.utc)

The tests still pass.
(resource here -- https://docs.djangoproject.com/en/4.0/topics/i18n/timezones/)

Samantha-Hampton (changes requested):
Since it still works the same, I'm not sure if you _need_ to change the way you deal with the date string to use the Django timezone utils thingy I found. So if you don't think it's necessary I'll change my review. I'm not certain if its one those things that makes it more efficient, but it is a bit neater :)

backend/curriculum_tracking/tests/test_api_views.py
TheAbstract: This is actually in `develop` and is only used in `AgileCardViewsetTests`, so it doesn't affect `TestBulkSetDueDatesApi` but we'll need some clarity from Sheena about `_test_set_due_time_permissions`. For topic cards I don't think it's an issue.

backend/curriculum_tracking/tests/test_api_views.py
TheAbstract: Yes here `due_time` doesn't matter, whatever it is, it shouldn't be able to go through.

backend/curriculum_tracking/tests/test_api_views.py
TheAbstract: @Samantha-Hampton 🙆🏾‍♂️ I like this much better.

Samantha-Hampton (approved):

Andy-Nkumane (approved):

backend/core/serializers.py
```diff
@@ -130,3 +129,19 @@ class Meta:
...
     message = serializers.CharField()
+
+
+class BulkSetDueTimeSerializer(serializers.Serializer):
+    class Meta:
+        fields = [
+            "due_time"
+            "flavours",
+            "content_item",
+            "team"
```
sheenarbw: since the team comes from the url, why does the serializer also need a team field?

backend/core/views.py
```diff
@@ -144,6 +145,26 @@ def summary_stats(self, request, pk=None):
...
 
+    @action(
+        detail=True,
+        methods=["POST"],
+        serializer_class=serializers.BulkSetDueTimeSerializer,
+        permission_classes=[HasObjectPermission(permissions=Team.PERMISSION_MANAGE_CARDS)]
+    )
+    def bulk_set_due_dates(self, request, pk=None):
+
+        team: models.Team = models.Team.objects.get(pk=pk)
```
sheenarbw: I think you can just say `self.get_object()`

backend/curriculum_tracking/serializers.py
```diff
@@ -221,8 +222,12 @@ class Meta:
             "open_pr_count",
             "oldest_open_pr_updated_time",
             "repo_url",
+            "flavour_names"
         ]
 
+    due_time = serializers.DateTimeField()
```
sheenarbw: Why are you changing this?

backend/curriculum_tracking/serializers.py
```diff
@@ -244,16 +249,16 @@ class Meta:
         fields = ["timestamp"]
 
 
-class AddReviewerUserSerializer(serializers.ModelSerializer):
+class SetDueTimeSerializer(serializers.ModelSerializer):
     class Meta:
-        model = models.AgileCard
-        fields = ["reviewers"]
+        model = models.RecruitProject
+        fields = ["due_time", "content_item", "flavours"]
```
sheenarbw: Why are you changing this?

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -116,13 +118,13 @@ def test_set_topic_card_due_time_permissions(self):
...
     def _test_set_due_time_permissions(self, card, get_progress):
-        recruit = UserFactory()
+        recruit = UserFactory(first_name="recruit")
 
         staff_member = UserFactory(is_staff=True)
 
         card.assignees.add(recruit)
 
-        due_time_1 = timezone.now() + timedelta(days=7)
+        due_time_1 = timezone.now() + timedelta(days=8)
```
sheenarbw: why are you changing this?

sheenarbw (changes requested):
This looks like it does the right thing. But also a whole lot of things that it shouldn't be doing.

We can wither set the due time of 1 card at a time or many cards at a time.

Setting the due time of 1 card at a time is already implemented. That code should not change.

Setting the due time of many cards in bulk is the thing this code needs to allow. That means making one @action and one serializer and then testing that one action.

backend/curriculum_tracking/serializers.py
TheAbstract: This change was made before I took this over.

backend/curriculum_tracking/serializers.py
TheAbstract: This change was made before I took this over.

backend/core/serializers.py
TheAbstract: These were the required fields on the issue.

Andy-Nkumane (approved):

Samantha-Hampton (approved):
Everything looks fine to me.
I tried to get this working on my local Tilde set up by creating a team and trying to set the bulk due dates, but kept failing - I'm probably not doing something right...but were you able to do this?

TheAbstract:
@Samantha-Hampton I assume you mean through the admin panel, I haven't done that manually.

backend/curriculum_tracking/serializers.py
```diff
@@ -221,6 +222,7 @@ class Meta:
             "open_pr_count",
             "oldest_open_pr_updated_time",
             "repo_url",
+            "flavour_names"
```
sheenarbw: What is this for? Can flavour_names be removedf?

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -35,6 +31,7 @@ class CardSummaryViewsetTests(APITestCase, APITestCaseMixin):
         "start_time",
         "open_pr_count",
         "oldest_open_pr_updated_time",
+        "flavour_names",
```
sheenarbw: What is this for?

backend/core/views.py
```diff
@@ -144,6 +145,26 @@ def summary_stats(self, request, pk=None):
...
+    )
+    def bulk_set_due_dates(self, request, pk=None):
+
+        team: models.Team = self.get_object()
+        serializer = self.get_serializer(data=request.data)
+
+        if serializer.is_valid():
+
+            team_cards = get_team_cards(team, serializer.validated_data.get('content_item'))
+            [
```
sheenarbw, line 162: This would have been better as a for loop.

backend/curriculum_tracking/tests/test_api_views.py
sheenarbw: We do self driven learning so the rules around setting due times is that:
- if there is no due time on a card, the assignee can set the due time or a manager can set a due time 
- if there is a due time already then only a manager can change it 

So if someone makes a commitment they cant just change it willy nilly.

sheenarbw (changes requested):
Looking good. Needs about 2 minutes worth of polish then we are good :)

PR #361, opened 2022-02-06
Teams dashboard

### LINKED ISSUE #300: On the teams dashboard, make sure that every graph has the same date range
You'll need to edit the getActivityLogCountsByDayForUsers function.

Find the minimum and maximum date and then make sure that all dates from the min to the max are represented in all the user output.

babalwa-mbolekwa:
Related issues: [#300 ]

## Description:

## Screenshots/videos
lineChart

## I solemnly swear that:

### DIFF
--- a/frontend/src/components/pages/TeamDashboard/Presentation.js
+++ b/frontend/src/components/pages/TeamDashboard/Presentation.js
@@ -55,7 +55,13 @@ const ReviewLineChart = ({
       }}
     >
       <CartesianGrid strokeDasharray="3 3" />
-      <XAxis dataKey="date" domain={[minimumDate, maximumDate]} />
+      <XAxis
+        dataKey="date"
+        domain={[
+          Math.floor(new Date(minimumDate).getTime() / 1000),
+          Math.floor(new Date(maximumDate).getTime() / 1000),
+        ]}
+      />
       <YAxis domain={[minimumCount, maximumCount]} />
       <Tooltip />
       <Line
--- a/frontend/src/components/pages/TeamDashboard/Presentation.js
+++ b/frontend/src/components/pages/TeamDashboard/Presentation.js
@@ -41,7 +41,6 @@ const ReviewLineChart = ({
   maximumDate,
 }) => {
   // TODO calculate maximum and minimum so that all the graphs have the same proportions
-
   return (
     // <ResponsiveContainer width="100%" height="100%">
     <LineChart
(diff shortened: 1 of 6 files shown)


babalwa-mbolekwa:
Please kindly review but not yet officially. The LineCharts only display the correct minimum and maximum dates for the team members whose IDs are not included in the `activityLogDayCounts` object in `src/stories/TeamDashboard.stories.js`

kaleem99 (changes requested):
- looks good but the date range for the first 3 users is different than the date range for the rest of the users, also looks like the tests can be made drier by removing one of the activityLogDayCounts objects.
- also in the utils.js file the getMinAndMaxDate and getMinimumAndMaximumValue seems to be doing similar things it would be better to use one function for iterating through activityLogDayCounts object and then call the two functions and check for the minMax date and minMax Value.

babalwa-mbolekwa:
I have thus far managed to make the Line charts have the same dates. I am currently stuck on sorting the dates to be in ascending order on the Line Chart
LineChart

babalwa-mbolekwa:
lineChart

kaleem99 (approved):

Sbonelo01 (approved):

frontend/src/components/pages/TeamDashboard/Presentation.js
```diff
@@ -25,6 +25,8 @@ import {
 } from "recharts";
 import { getMinimumAndMaximumValue } from "./utils";
```
sheenarbw: You are importing 2 things from utils. Wouldn't it make sense to have 1 import statement instead of 2?

Vuyisanani (approved):

frontend/src/components/pages/TeamDashboard/utils.test.js
```diff
@@ -1,79 +1,211 @@
...
+      {
+        date: "2021-07-16",
+        COMPETENCE_REVIEW_DONE: 0,
+        PR_REVIEWED: 0,
+      },
+      {
+        date: "2021-07-17",
+        COMPETENCE_REVIEW_DONE: 0,
+        PR_REVIEWED: 0,
+      },
```
sheenarbw, line 66: should the 18th be in here too? Just thinking that it might look like folks are reviewing every day when they are not.

sheenarbw (changes requested):
This is good. There is just a bit of polish. It would also be good to make sure that there aren't gaps in the data.

babalwa-mbolekwa:
dateRangeXAxis

Sbonelo01 (approved):

kaleem99 (approved):

sheenarbw (approved):

PR #383, opened 2022-03-01
Remove review access/369

### LINKED ISSUE #369: Sometimes the add review button is visible when it shouldn't be
This bug was pointed out by Sam

Sometimes the AddReview button will show up when looking at the card details page, but when viewing the same card on a person's board then the button doesn't come up. 

This is likely an error to do with permissions. It's probably the card details page that has the problem.

If someone has view access to a person, but not review access, then they should not see any "Add review" buttons.

FaithMo:

## Description:

## Screenshots/videos
When the user has the frontend: view only access, the Add Review button should not appear on the card.

image

## I solemnly swear that:


kaleem99 (approved):

babalwa-mbolekwa (approved):

frontend/src/components/regions/AgileBoard/AgileCard/AgileCardActions/utils.js
```diff
@@ -35,13 +36,16 @@ export function getTeamPermissions({ authUser, viewedUser }) {
 
 function getShowAddReviewButton({ card, permissions, isReviewer }) {
   const REVIEW_STATUSES = [IN_REVIEW, COMPLETE, REVIEW_FEEDBACK];
-  if (isReviewer && REVIEW_STATUSES.indexOf(card.status) !== -1) return true;
+  if(isReviewer && REVIEW_STATUSES.indexOf(card.status) !== -1){
+    if(permissions[VIEW_ONLY]) return false;
```
sheenarbw: This logis isn't correct. Sometimes a person can be given multiple permissions. They dont cancel each other out, they stack up.

If you have permission to use the blue pen and I give you permission to use the red pen as well, you can still use the blue pen. 

`getShowAddReviewButtons` looks like it works fine. 

Take a look at `src/components/regions/CardDetails/index.js`.   

There's a piece of code in there that is wrong:
```
  const showAddReviewButton =
    (isReviewer || isStaff) &&
    [IN_REVIEW, COMPLETE, REVIEW_FEEDBACK].indexOf(projectCardStatus) !== -1;
```
Can you fix this code.

frontend/src/constants.js
```diff
@@ -58,13 +58,15 @@ export const VIEW_ALL = "VIEW_ALL";
...
 export const TRUSTED_REVIEWER = "TRUSTED_REVIEWER";
+export const VIEW_ONLY = "VIEW_ONLY";
 
 export const TEAM_PERMISSIONS = [
   MANAGE_CARDS,
   VIEW_ALL,
   ASSIGN_REVIEWERS,
   REVIEW_CARDS,
   TRUSTED_REVIEWER,
+  VIEW_ONLY
```
sheenarbw: Permissions need to be implemented on the backend. VIEW_ONLY doesn't exist anywhere so this definately wont work.

You might as well have added a permission called `CAN_HAZ_SANDWICH`, it would have had the same effect. The frontend is a representation of what the backend knows. If you write logic around things that dont mean anything then the logic is meaningless.

sheenarbw (changes requested):
You seem to be editing the wrong file and also inventing new permissions that dont exist. 

You'll need to start again. The code you wrote wont have any effect as far as I can see. So I dont think you've introduced new bugs,. but the bug you were trying to fix has not been fixed.

kaleem99 (approved):

babalwa-mbolekwa (approved):

frontend/src/components/regions/CardDetails/index.js
```diff
@@ -108,12 +108,10 @@ function CardDetailsUnconnected({
   const isAssignee =
     ((project || {}).recruitUsers || []).indexOf(authUser.userId) !== -1;
 
-  const isStaff = authUser.isStaff === 1;
-
   const projectCardStatus = project && project.agileCardStatus;
 
   const showAddReviewButton =
-    (isReviewer || isStaff) &&
+    (isReviewer) &&
```
sheenarbw: We have a utility function that does this. Should we not call that function? I think it's called `getShowAddReviewButton`

sheenarbw (changes requested):

kaleem99 (approved):

babalwa-mbolekwa (approved):

frontend/src/components/regions/CardDetails/Presentation.js
```diff
@@ -168,7 +168,7 @@ export default ({
             showUpdateProjectLinkForm={showUpdateProjectLinkForm}
             linkSubmission={linkSubmission}
             formErrors={formErrors}
-            showAddReviewButton={showAddReviewButton}
+            getShowAddReviewButton={getShowAddReviewButton}
```
sheenarbw: This is a bug.

The ProjectDetails component says
```
{showAddReviewButton && (
        <CardButton
          label="Add Review"
          startIcon={<RateReviewRoundedIcon />}
          onClick={handleClickAddReview}
        />
      )}
```

It expects that `showAddReviewButton` is handed in as a prop. But now that prob will always be undefined. That means that no more review buttons will show up for anyone ever.  

Function arguments matter. If you dont pass in the right props you wont get the right behavior.  

Please do not edit the `ProjectDetails` component. Please just pass it the information that it needs. I would recommend calling the `getShowAddReviewButton` function somewhere in this code.

frontend/src/components/regions/CardDetails/index.js
```diff
@@ -145,7 +135,7 @@ function CardDetailsUnconnected({
     projectReviews: currentProjectReviews,
 
     handleClickAddReview,
-    showAddReviewButton,
+    getShowAddReviewButton,
```
sheenarbw: Dont pass thew function in. Call the function and pass the result into the component.

frontend/src/components/regions/CardDetails/Presentation.js
```diff
@@ -148,7 +148,7 @@ export default ({
   handleClickAddReview,
   handleClickUpdateProjectLink,
   showUpdateProjectLinkForm,
-  showAddReviewButton,
```
sheenarbw: This should not change. We need a boolean here, not a fdunction.

sheenarbw (changes requested):
This has a major bug. 

You  did a lot of work to make the `getShowAddReviewButton` function available to different components. Have you considered calling the function?

frontend/src/utils/cardButtons.js
```diff
@@ -0,0 +1,160 @@
...
+    const REVIEW_STATUSES = [IN_REVIEW, COMPLETE, REVIEW_FEEDBACK];
+    if (isReviewer && REVIEW_STATUSES.indexOf(card.status) !== -1) return true;
+    if (
+      (permissions[REVIEW_CARDS] || permissions[TRUSTED_REVIEWER]) &&
+      REVIEW_STATUSES.indexOf(card.status) !== -1
+    )
+      return true;
+  
+    return false;
+  }
```
babalwa-mbolekwa: Hey Faith, please make the code dryer. There's repetition with:
```
REVIEW_STATUSES.indexOf(card.status) !== -1
```
How about:
```
return (
    (ifYour'reReviwer or
      haveReviewPermission or
      areTrustedReviewer) &&
    REVIEW_STATUSES.indexOf(card.status) !== -1
  )
```
then you can also remove the `return true` and `return false` because they'll be automatically returned from the statement

babalwa-mbolekwa (changes requested):

kaleem99 (approved):

babalwa-mbolekwa (approved):

Vuyisanani (changes requested):
Hi @FaithMo 

Please fix your merge conflicts

Vuyisanani (changes requested):
Great work @FaithMo 

Was there a specific reason to delete `frontend/src/components/regions/AgileBoard/AgileCard/AgileCardActions/utils.js`?

Was the file used only by one file?

FaithMo:
Hey @Vuyisanani yes, it was intentional. The code is refactored and moved to frontend/src/utils/cardButtons.js so that we can be able to re-use the functions in other components later

frontend/src/utils/cardButtons.js
```diff
@@ -0,0 +1,158 @@
...
+    IN_PROGRESS,
+    IN_REVIEW,
+    TEAM_PERMISSIONS,
+    MANAGE_CARDS,
+    REVIEW_CARDS,
+    TRUSTED_REVIEWER,
+  } from "../constants";
+  
+  export function getTeamPermissions({ authUser, viewedUser }) {
+    let result = {};
```
Vuyisanani: Is it one results, or results that you store in this object?

Should be plural, not singular

frontend/src/utils/cardButtons.js
```diff
@@ -0,0 +1,158 @@
...
+    let showButtonCancelReviewRequest = false;
+  
+    const isAssignee = card.assignees.indexOf(authUser.userId) !== -1;
+    const isReviewer = card.reviewers.indexOf(authUser.userId) !== -1;
+    const permissions = getTeamPermissions({ authUser, viewedUser });
+  
+    let reviewRequestButtons;
+    if (card.contentTypeNice === "project") {
+      // PROJECT CARDS
+
```
Vuyisanani: you dont need this line (empty line)

frontend/src/utils/cardButtons.js
```diff
@@ -0,0 +1,158 @@
...
+      reviewRequestButtons = getReviewRequestButtons({
+        card,
+        permissions,
+        isAssignee,
+      });
+      showButtonRequestReview = reviewRequestButtons.showButtonRequestReview;
+      showButtonCancelReviewRequest =
+        reviewRequestButtons.showButtonCancelReviewRequest;
+    } else if (card.contentTypeNice === "workshop") {
+      // WORKSHOP CARDS
```
Vuyisanani: this comment is flowtiing alone, like the one in line 80

You can either remove or move it on top of workshop cards

Vuyisanani (changes requested):
This is absolutely looking great, except the few comments, with all the other comments I did not comment on

Vuyisanani (approved):

sheenarbw (changes requested):
This seems to have a bug.

When I run this frontend against real data then the following happens:
1. When I look at abdul.jaggers@umuzi.org 's board I see the "add review" button on his level zero katas card
2. When I look at the details of this card I do not see the "Add review" button 

I should have a consistent experience. If I can review the cards I should see the button in both places.

My first thought is that the data isn't flowing properly. I assume that you just aren't calling the function properly inside the card details page. Please double check your props are correct.

Another option would be to just use the CardActions component inside the CardDetails page. We can just show all the relevant buttons.

Sbonelo01 (approved):

Vuyisanani (approved):

sheenarbw (approved):
Looks good.

PR #395, opened 2022-03-07
Feat/read only api for activity log.models.event type

Rokhuda:

## Description:

## I solemnly swear that:

### DIFF
--- a/backend/activity_log/serializers.py
+++ b/backend/activity_log/serializers.py
@@ -26,26 +26,19 @@ class ActivityLogDayCountSerializer(serializers.Serializer):
 class EventTypeSerializer(serializers.ModelSerializer):
     class Meta:
         model = models.EventType
+        fields = ["name"]
 
 
 class LogEntrySerializer(serializers.ModelSerializer):
     class Meta:
         model = models.LogEntry
+        fields = ["id", "timestamp", "event_type", "actor_user", "effected_user"]
 
 
 class ActivityLogEventTypeSerializer(serializers.Serializer):
     class Meta:
-        fields = ["id"]
+        fields = ["event_type", "actor_user", "effected_user"]
 
-    id = serializers.SerializerMethodField("get_id")
+    timestamp = serializers.DateTimeField(auto_now_add=True)
     event_type = EventTypeSerializer(read_only=True, many=True)
-    activity_log = LogEntrySerializer(read_only=True, many=True)
-
-    def get_id(self, instance):
-        result = f"timestamp={str(instance['timestamp'])}"
-        if instance["filters"]:
-            return f"{result}&{instance['filters']}"
-        return result
-
-    def get_timestamp(self, instance):
-        return str(instance["timestamp"])
+    log_entry = LogEntrySerializer(many=True)
--- a/backend/activity_log/serializers.py
+++ b/backend/activity_log/serializers.py
@@ -1,3 +1,4 @@
+from . import models
 from rest_framework import serializers
 
 
--- a/backend/activity_log/serializers.py
+++ b/backend/activity_log/serializers.py
@@ -24,3 +24,32 @@ class ActivityLogDayCountSerializer(serializers.Serializer):
 
     def get_total(self, instance):
         return instance["total"]
+
+
+class EventTypeSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = EventType
+
+
+class LogEntrySerializer(serializers.ModelSerializer):
+    class Meta:
+        model = LogEntry
+
+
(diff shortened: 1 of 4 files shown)


backend/activity_log/tests/test_api_views.py
```diff
@@ -58,3 +58,40 @@ def test_list_api_filter_by_event_type(self):
...
+class TestActivityLogEventTypeViewSet(APITestCase, APITestCaseMixin):
+    LIST_URL_NAME = "activitylogeventtype-list"
+    SUPPRESS_TEST_GET_LIST = True
+    SUPPRESS_TEST_POST_TO_CREATE = True
+
+    def setUp(self):
+        self.today = timezone.now()
+
+        self.entry1 = factories.EventTypeFactory()
+        self.entry1 = factories.LogEntryFactory()
```
Samantha-Hampton: It's highly possible that there is something I'm not aware of, but does the second self.entry1 line not simply overwrite the first self.entry1 line?

backend/activity_log/tests/test_api_views.py
```diff
@@ -58,3 +58,40 @@ def test_list_api_filter_by_event_type(self):
...
+        self.entry1.timestamp = self.today
+        self.entry1.save()
+
+        self.entry2 = factories.EventTypeFactory()
+        self.entry2 = factories.LogEntryFactory()
```
Samantha-Hampton: similar concern here

backend/activity_log/serializers.py
```diff
@@ -24,3 +21,23 @@ def get_date(self, instance):
...
+        fields = ["name"]
+
+
+class LogEntrySerializer(serializers.ModelSerializer):
+    class Meta:
+        model = models.LogEntry
+        fields = ["id"]
+
+
+class ActivityLogEventTypeSerializer(serializers.Serializer):
```
Samantha-Hampton: If this is the serialiser you are using why did you also create EventTypeSerializer and LogEntrySerilizer?

backend/activity_log/tests/test_api_views.py
```diff
@@ -58,3 +58,40 @@ def test_list_api_filter_by_event_type(self):
...
+        self.entry2.timestamp = self.today
+        self.entry2.save()
+
```
Samantha-Hampton: do we not log in a generic user here to see that anyone logged in can access? or is assumed that this will be the case?

backend/activity_log/tests/test_api_views.py
```diff
@@ -58,3 +58,40 @@ def test_list_api_filter_by_event_type(self):
...
+        url = f"{self.get_list_url()}?effected_user={self.entry1.effected_user.id}"
+        response = self.client.get(url)
+        self.assertEqual(len(response.data), 1)
+        self.assertEqual(self.entry1.timestamp, self.today)
+
+    def test_list_api_filter_by_actor_user(self):
+        url = f"{self.get_list_url()}?actor_user={self.entry1.actor_user.id}"
+        response = self.client.get(url)
+        self.assertEqual(len(response.data), 1)
+        self.assertEqual(self.entry1.timestamp, self.today)
```
Samantha-Hampton: in your setUp you create two entries but in your tests you only use one of them - is there a reason for this?

Samantha-Hampton (changes requested):
I'm still figuring a bunch of these things out so not sure how valid all my points are - so apologies if they turn out to be incorrect - but it is what I noticed so far :)

backend/activity_log/tests/test_api_views.py
```diff
@@ -68,15 +68,12 @@ class TestActivityLogEventTypeViewSet(APITestCase, APITestCaseMixin):
     def setUp(self):
         self.today = timezone.now()
 
-        self.entry1 = factories.EventTypeFactory()
```
Samantha-Hampton: I suppose I'm a bit confused by which of these you deleted....since the ticket is on creating the api for an event_type, would you not test it by creating one of those instead of a log entry - but I'm not sure how much difference this makes. Maybe we can ask Sheena later....

Samantha-Hampton (commented):
I think it would be good to chat to Sheena in the session later, since I'm a bit mystified by what we actually want this API to do. The ticket suggests it's just related to the Event type. In which case since an event_type only has a name and a description are we only interested in exposing the different event_types that exist? Or is that so we can search by event_type and then further refine - so that we need to see which log_entry the event_type is associated with?

Samantha-Hampton (approved):

Andy-Nkumane (approved):

sheenarbw (approved):

PR #406, opened 2022-03-09
Log card review request cancelled called/311

### LINKED ISSUE #311: Make sure that log_card_review_request_cancelled gets called correctly and does the right thing. Note that a review can be cancelled multiple times for a single card

Samantha-Hampton:

## Description:

1) Added a log_creator to the cancel_review_request function in the api_views under curriculum tracking

2) Amended the test class in test_activity_log_entry_creators_envoked such that with setUp it could work for both testing logging of review_request as well as cancel_review_request

3) Amended the test class in test_api_views, under curriculum_tracking, to include setUp such that it could work for the existing permissions tests for review_request as well as for the new permission tests written for cancel_review_request

4) Removed assertion for card status in 'cancel_review_request' function in api_views.py in curriculum tracking and added test instead to check status of card after review request cancelled when a card has had review feedback and has requested review then to make sure it goes back to review feedback.

## I solemnly swear that:


backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
```diff
@@ -108,41 +111,74 @@ def test_add_review(self):
         self.assertEqual(LogEntry.objects.count(), 2)
 
 
-class log_card_review_requested_Tests(APITestCase, APITestCaseMixin):
+class log_card_review_requested_and_cancelled_Tests(APITestCase, APITestCaseMixin):
```
TheAbstract, line 227: This looks like you're replacing one test with another, instead of adding a new test. Is this intentional?

Samantha-Hampton: Yup, the idea is to leverage the same setUp for both sets of tests

Andy-Nkumane (approved):

backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py
Samantha-Hampton: @TheAbstract any other changes to request on this PR?

TheAbstract (approved):

backend/curriculum_tracking/api_views.py
```diff
@@ -387,8 +387,14 @@ def cancel_review_request(self, request, pk=None):
...
 
-        # card.status = models.AgileCard.IN_PROGRESS
-        # card.save()
+        log_creators.log_card_review_request_cancelled(
+            card=card, actor_user=request.user
+        )
+        card.refresh_from_db()
+
+        assert (
+            card.status == models.AgileCard.IN_PROGRESS
```
sheenarbw: Are you sure? 
If a card is in review feedback then a review is requested, then the request is cancelled, then where should the card be? 
If it ends up in progress then I'm not sure that this is the correct behavior. We shouldn't make it seem like it's the intended behavior.

sheenarbw (changes requested):
This looks good, it has one potential bug. I'm not even sure if it's a bug, but it might cause problems.

backend/curriculum_tracking/api_views.py
Samantha-Hampton: what is funny, is that I didn't write that particular bit of code lol

Samantha-Hampton: I mean I wrote this part, but not the rest:

log_creators.log_card_review_request_cancelled(
             card=card, actor_user=request.user
         )

Samantha-Hampton: oh goodness, maybe I did write it all, pattern matching from the function above. The sleep deprivation is getting to me. Trying to think how to address though, because either it should be in progress or back in review feedback, but maybe here is not the place to try assert this and instead we should write a test about this behaviour....

Samantha-Hampton: ok, have added a new test - but possibly this is tested elsewhere already.....

TheAbstract: It looks like there two scenarios:
1. Card is in Progress, I think its ready for review; request a review, notice a bug, and cancel my review request. Card goes back to In Progress.
2. I've already gotten some feedback so my car is in Review Feedback, I think its ready for review; request a review, notice a bug, and cancel my review request. Card goes back to Review Feedback.  

@Samantha-Hampton I believe this is the current behaviour. So what we would need is a test that checks these scenarios to make sure that we're logging the right thing.

TheAbstract (approved):

Andy-Nkumane (approved):

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -116,8 +116,16 @@ def setUp(self):
...
 
-    def test_request_review_permissions_non_superuser_staff(self):
+        self.login_as_superuser()
+        request_review_url = f"{self.get_instance_url(self.card_2.id)}request_review/"
+        response = self.client.post(request_review_url)
+
+        self.assertEqual(response.status_code, 200)
 
+        self.card_2.refresh_from_db()
+        self.assertEqual(self.card_2.status, AgileCard.IN_REVIEW)
```
sheenarbw: It looks like the setup for this test is taking on too much responsibility.  
Ideally a setup should only have sanity checking assersions (if any assersians at all)

A setup is not a test. Test functions start with the word `test_`

Can this be refactored so that the setup has less responsibility? 

The code works, but this is a bad practice and it shouldn't turn into a bad habit.

sheenarbw: Sanity checking asserts are almost like documentation. They should pass unless the world has gone totally mad.

sheenarbw (changes requested):
There is one odd habit that seems to be developing. Let's clean that up. Then we are good.

Andy-Nkumane (changes requested):
with the latest changes, there's a lot of repeated code - can you make it DRY

Andy-Nkumane:
please fix conflicts

TheAbstract (changes requested):
Getting some errors when running the tests
```
ERROR: test_cancel_review_requested (curriculum_tracking.tests.test_activity_log_entry_creators_envoked.log_card_review_requested_and_cancelled_Tests)
  File "/Users/umuzi/Tilde/backend/curriculum_tracking/tests/test_activity_log_entry_creators_envoked.py", line 233, in setUp
    self.content_item = factories.ProjectContentItemFactory(
NameError: name 'factories' is not defined
```
The issue is this; `ProjectContentItemFactory` is being imported directly in line 21 but it's being used as if we've just imported `factories`. So `factories.ProjectContentItemFactory()` vs `ProjectContentItemFactory()`.

TheAbstract (approved):

TheAbstract:
@Samantha-Hampton there are some `frontend` files here that have been changed, is this intentional?

TheAbstract (approved):

Andy-Nkumane (approved):

sheenarbw (approved):

PR #338, opened 2022-01-20
Expand review feature

Sbonelo01:
Related issues: [#263]

#### How the solution works is:
If we find a long review, we shorten it, and then we display the crop-free icon next to the review. If the review is already short, we don't display the icon next to the review.
clicking the icon opens a modal, the modal displays the full review

## Screenshots/videos
Web capture_20-1-2022_195437_localhost

The Modal:
Web capture_20-1-2022_20530_localhost

## I solemnly swear that:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #429, opened 2022-03-23
Close button to modal widget

### LINKED ISSUE #402: Add a close button to the Modal widget
src/components/widgets/Modal.js 

Add an X to the top right of the widget with `onClick={onClose}` 

Make sure that the whole file matches the styleguide

babalwa-mbolekwa:
Related issues: [#402 ]

## Description:

## Screenshots/videos
closeButton

## I solemnly swear that:


Sbonelo01 (changes requested):
This looks very good

can we please work on making it responsive to mobile 

Screenshot_2022-03-23_12-39-21(Babalwa-review)

Sbonelo01 (changes requested):
This looks good and is responsive.

Please have a look at the style guide regarding file extensions and how to name functional components: https://github.com/Umuzi-org/Tilde/blob/develop/frontend/STYLE_GUIDE.md

kaleem99 (approved):

Sbonelo01 (approved):

KhalidUmuzi (approved):

KhalidUmuzi (changes requested):
So with this change it doesn't affect any other modals on Tilde but only this one.

I believe that for consistency, all modals on tilde should have the same behavior so if we apply this change here, we apply it everywhere.

in the filepath frontend/src/widgets there is a file called Modal.js

That is used in a number of places and the change should be made to this modal and filter through to all other modals on Tilde.

And example of this is https://github.com/Umuzi-org/Tilde/pull/338. According to a comment made on this issue, this change is made to address that modal as well. It does not.

babalwa-mbolekwa:
> So with this change it doesn't affect any other modals on Tilde but only this one.
> 
> I believe that for consistency, all modals on tilde should have the same behavior so if we apply this change here, we apply it everywhere.
> 
> in the filepath frontend/src/widgets there is a file called Modal.js
> 
> That is used in a number of places and the change should be made to this modal and filter through to all other modals on Tilde.
> 
> And example of this is #338. According to a comment made on this issue, this change is made to address that modal as well. It does not.

@KhalidUmuzi The styling doesn't work out when I try to style the addReviewModal or the Modal itself. The width of the addReview modal remains the same. I think there's a number of things that need to be changed in the background in addition to what we changed

Vuyisanani (approved):

sheenarbw (approved):
weeee

PR #485, opened 2022-05-11
Fix failing unit test

### LINKED ISSUE #476: URGENT: Fix failing unit test on frontend

KhalidUmuzi:

## Description:

## I solemnly swear that:


FaithMo (approved):
looks good

Sbonelo01 (approved):

frontend/src/components/pages/TeamDashboard/Presentation.jsx
```diff
@@ -72,9 +72,11 @@ export default function ({ team, activityLogDayCounts }) {
     activityLogDayCounts,
   });
 
+  const dateFilter = new Date();
```
sheenarbw: If I ask you to explain to me in English what the date filter is, you will say it is the "end date" or the "current date". It's used for filtering, but what is it? It's definitely not a filter. A filter is a function.

frontend/src/components/pages/TeamDashboard/utils.js
```diff
@@ -1,4 +1,4 @@
-export function updateActivityLogDayCounts({ activityLogDayCounts }) {
+export function updateActivityLogDayCounts({ activityLogDayCounts }, dateFilter) {
```
sheenarbw: Why use positional arguments? Isn't destructuring generally much safer? And it's good to be consistent. 

`export function updateActivityLogDayCounts({ activityLogDayCounts, dateFilter }) `

sheenarbw (changes requested):
This works but the code has some strangeness to it.  

If we introduce a new convention to the code then it's good to know that the juniors will follow that convention. The main thing here is that we should aim to use restructuring and avoid using positional arguments. And when in doubt be consistent. 

The name is a bit of a bunny one - if I just see how the function is used then it is not at all clear how the date is used in the filter, is it a start date or an end date? If I want to understand a function call then I need to go sniff around in the function definition. Ideally if I look at a line of code I should understand it's intention without going on any treasure hunts. 

If I put this code in I would probably end up cleaning it up or getting someone else to.

Sbonelo01 (approved):

FaithMo (approved):

sheenarbw (approved):

PR #547, opened 2022-07-14
Endpoit for accessing which tilde curriculums are associated with each tilde user

### LINKED ISSUE #410: Create API endpoint for accessing which Tilde curriculums are associated with each Tilde user

Rokhuda:

## Description:

## I solemnly swear that:


backend/core/serializers.py
```diff
@@ -203,6 +203,26 @@ class Meta:
...
 
+class StreamRegistrationSerialiser(serializers.ModelSerializer):
+    class Meta:
+        model = models.StreamRegistration
+        fields = [
+            'id',
+            'active',
+            'user',
+            'stream',
+            'active'
```
Samantha-Hampton: Is there meant to be two 'active' fields?

backend/core/serializers.py
```diff
@@ -203,6 +203,26 @@ class Meta:
...
+            ]
+
+    user = serializers.SerializerMethodField("get_user_name")
+    stream = serializers.SerializerMethodField("get_stream_name")
+
+    def get_user_name(self, instance):
+        return instance.user.email
+
+    def get_stream_name(self, instance):
+        return instance.stream.name
```
Samantha-Hampton: Can someone be registered for more than one stream? Is this possible? Should it be?

Samantha-Hampton (commented):
I'm going to look in more detail later/tomorrow, just two questions for now....

backend/core/serializers.py
Rokhuda: @ry-oc

Samantha-Hampton (changes requested):
Hey Rokhuda :), not sure if you heard the part of the discussion in standup this morning about the stream thing? (and how it should be ignored, lol). So I think we want to focus here on returning all the curriculums associated with a user. A user will have a number of curriculums depending on their "stream" and so we actually don't care about the stream but rather what are all the curriculums they have linked to them (ie bootcamp, intro to tilde, etc). Does this make sense?

Andy-Nkumane (changes requested):
getting there:
* there are 2 CourseRegistrationSerialiser one in core and the other in curriculum tracking

Rokhuda:
@Andy-Nkumane can you check again, https://github.com/Umuzi-org/Tilde/pull/547/commits/723b78cc509fa4a260cca2cbbbb895a04fa70fba

Andy-Nkumane (changes requested):
backend/core/management/serializers.py seems to be a copy of backend/curriculum_tracking/serializers.py

Samantha-Hampton:
Seems this branch has some conflicts which need to be resolved

Samantha-Hampton (changes requested):
I think this is shaping up nicely! :) Would be good to also get curriculum_id and user_id, so that when we call the api we can filter per user or curriculum. 

This is an example of yours when I run the dev server:
Screenshot 2022-08-01 at 16 26 28

And here is an example of the agile card info - it includes the agile card id, but also the content item id and the assignee id. 

Screenshot 2022-08-01 at 16 26 55

Let me know if this makes sense?

backend/backend/urls.py
```diff
@@ -163,6 +163,12 @@
 #     "user-recrtuitprojects",
 # )
 
+router.register(
+    r"course_registration",
```
sheenarbw: I know this might be news but 'course_registration' should actually be pleural. Our URLs are not as consistent as they should be.

Reason:

If we `GET` `/api/course_registrations` then the response will be a list of course registrations

backend/curriculum_tracking/api_views.py
```diff
@@ -145,7 +145,7 @@ class CardSummaryViewset(viewsets.ModelViewSet):
         | core_permissions.HasObjectPermission(
             permissions=Team.PERMISSION_VIEW,
             get_objects=core_permissions.get_teams_from_user_filter("assignees"),
-        ),
+        )
```
sheenarbw: What autoformatted are you using? It looks like your setup is disagreeing with mine.

This project uses `black`, just the default settings. 

It's important that we use the same setup.

backend/curriculum_tracking/api_views.py
```diff
@@ -1045,3 +1033,10 @@ class ContentItemAgileWeightViewSet(viewsets.ModelViewSet):
     serializer_class = serializers.ContentItemAgileWeightSerializer
     filter_backends = [DjangoFilterBackend]
     filterset_fields = ["content_item", "weight"]
+
+
+class CourseRegistrationViewset(viewsets.ModelViewSet):
+    queryset = models.CourseRegistration.objects.all().order_by("user")
+    serializer_class = serializers.CourseRegistrationSerialiser
+    filter_backends = [DjangoFilterBackend]
+    permission_classes = [core_permissions.IsReadOnly & permissions.IsAuthenticated]
```
sheenarbw: These permissions are a bit strange. Basically, any user can check what any other user is registered for so long as they are logged in.

backend/curriculum_tracking/serializers.py
```diff
@@ -231,10 +231,7 @@ class Meta:
 class NewReviewSerializer(serializers.ModelSerializer):
     class Meta:
         model = models.RecruitProjectReview
-        fields = [
-            "status",
-            "comments",
-        ]
+        fields = ["status", "comments"]
```
sheenarbw, line 234: again, the autoformatter doesn't seem to be correct.

backend/curriculum_tracking/serializers.py
```diff
@@ -707,3 +686,18 @@ def save(self,**kwargs):
         instance.set_flavours(flavour_names)
         return instance
 
+
+
+class CourseRegistrationSerialiser(serializers.ModelSerializer):
+    class Meta:
+        model = models.CourseRegistration
+        fields = ["id", "user", "curriculum"]
```
sheenarbw: `user` always exposes the user id, that is the standard behavior. If you want to expose the user email then the field should be named `user_email` and the method should be called `get_user_email`. 

If there is a sensible default behavior then it's best not to override it.  You can add extra stuff. But arbitrarily changing the definition of a word is probably going to confuse someone. 

Same for the `curriculum` field.

backend/curriculum_tracking/api_views.py
```diff
@@ -1045,3 +1033,10 @@ class ContentItemAgileWeightViewSet(viewsets.ModelViewSet):
     serializer_class = serializers.ContentItemAgileWeightSerializer
     filter_backends = [DjangoFilterBackend]
     filterset_fields = ["content_item", "weight"]
+
+
+class CourseRegistrationViewset(viewsets.ModelViewSet):
+    queryset = models.CourseRegistration.objects.all().order_by("user")
```
sheenarbw, line 1039: You can add filters here.

```
filterset_fields = ['user','curriculum']
```
This will let you filter by those ids. See my comments on the serializer, it's important that these are ids and not strings.

backend/curriculum_tracking/serializers.py
sheenarbw: If Sam needs you to also expose the email address and curriculum name then that's fine, but those should be separate fields that are explicitly named.

sheenarbw (changes requested):
On the right track but there is some work to do:

- some inconsistent naming of fields
- your auto formatter is fighting with my autoformatter. You should be using `black`
- I added some stuff about filters, I think that's the help you were looking for?
- list urls should be pleural (explained inline)

backend/curriculum_tracking/serializers.py
```diff
@@ -679,31 +662,54 @@ class Meta:
...
-        instance = super(ContentItemAgileWeightSerializer,self).save()
+                raise ValidationError(
+                    f"flavour '{flavour}' not allowed. Choose from {available_flavours}"
+                )
+        instance = super(ContentItemAgileWeightSerializer, self).save()
         instance.set_flavours(flavour_names)
         return instance
 
+
+class CourseRegistrationSerialiser(serializers.ModelSerializer):
```
sheenarbw, line 690: This is a bit verbose. You are implementing stuff that DRF gives us for free

backend/curriculum_tracking/serializers.py
```diff
@@ -679,31 +662,54 @@ class Meta:
...
+
+    user_name = serializers.CharField(required=False)
+    curriculum_name = serializers.CharField(required=False)
+
+    class Meta:
+        model = models.CourseRegistration
+        fields = ["id", "user", "curriculum", "user_name", "curriculum_name"]
+
+    user = serializers.SerializerMethodField("get_user")
+    user_name = serializers.SerializerMethodField("get_user_email")
```
sheenarbw: Is it a user name or a user email? 

Can you please name stuff according to what it is?

backend/curriculum_tracking/serializers.py
```diff
@@ -679,31 +662,54 @@ class Meta:
...
+class CourseRegistrationSerialiser(serializers.ModelSerializer):
+
+    user_name = serializers.CharField(required=False)
+    curriculum_name = serializers.CharField(required=False)
+
+    class Meta:
+        model = models.CourseRegistration
+        fields = ["id", "user", "curriculum", "user_name", "curriculum_name"]
+
+    user = serializers.SerializerMethodField("get_user")
```
sheenarbw: This is not needed, we get it for free:

```
user = serializers.SerializerMethodField("get_user")

...

    def get_user(self, instance):
        return instance.user.id
```

If you put "user" inside the Meta.fields array then DRF will automatically check if CourseReg has a user attribute. Since it's a foreign key DRF automatically exposes the id. 

The same goes for curriculum.

sheenarbw (changes requested):
It mostly looks good. There's some funny business in the serializer

backend/curriculum_tracking/serializers.py
sheenarbw: Same comment as before. 

You are getting a user's email and calling it a name. Please call an email an email. 

Call a spade a spade. Call an email an email.

You are saying: `user_name = email things`
It would be better to say `user_email = email things`

backend/curriculum_tracking/serializers.py
```diff
@@ -679,31 +662,46 @@ class Meta:
...
+                )
+        instance = super(ContentItemAgileWeightSerializer, self).save()
         instance.set_flavours(flavour_names)
         return instance
 
+
+class CourseRegistrationSerialiser(serializers.ModelSerializer):
+
+    user_name = serializers.CharField(required=False)
+    curriculum_name = serializers.CharField(required=False)
```
sheenarbw: Can you make these readonly?

sheenarbw (changes requested):

sheenarbw (approved):

PR #548, opened 2022-07-15
Widget for buttons

### LINKED ISSUE #469: Create a story that contains our button components
Create a widget named Button that takes the material UI button and styles it our way.

Make sure we always import this new widget and use it instead of using the normal mui button.

Later on we will get some input from the dpd team and we'll need to apply styling. When we re-style our buttons we should just edit one file instead of many. 

image

FaithMo:

## Description:

Here are pages that I used the widget on 
## Screenshots/videos
image

image

image

image

image

image

image

## I solemnly swear that:


frontend/src/components/widgets/Button.jsx
```diff
@@ -0,0 +1,20 @@
+import React from "react";
+import Button from "@material-ui/core/Button";
+import { makeStyles } from "@material-ui/core/styles";
+
+const useStyles = makeStyles((theme) => ({
+  button: {
+    border: "1px solid grey",
```
Sbonelo01, line 8: Can you also please style the button to not always turn its children to upper case

Sbonelo01 (changes requested):
This button looks good.

The second step in this issue says that we now need to :
```
"Make sure we always import this new widget and use it instead of using the normal mui button."
```

Can you please also implement this aswell

I'm not sure if this should be in an unprotected branch since we will have to get feedback from dpd first.

Sbonelo01 (approved):

Vuyisanani (approved):

sheenarbw (approved):

PR #587, opened 2022-08-25
Burndown create should save latest data

Andy-Nkumane:

## Description:

## I solemnly swear that:


backend/curriculum_tracking/models.py
```diff
@@ -1477,9 +1477,9 @@ class BurndownSnapshot(models.Model):
     def create_snapshot(Cls, user):
         match = Cls.objects.filter(
             user=user,
-            timestamp__gte=timezone.now()
+            timestamp__gte=timezone.now()  
             - timedelta(hours=Cls.MIN_HOURS_BETWEEN_SNAPSHOTS),
-        ).first()
+        ).last()
```
Samantha-Hampton: So the tests pass on my side whether this is set to last() or first() so I'm not sure what this change is doing?
Or if maybe the tests aren't capturing what a board regen would look like? To be honest I'm not sure how to test a board regen so I can't actually say if this isn't the right way, lol

Samantha-Hampton (changes requested):
So after relooking what I think might work is to create a few snapshots in quick succession so that only one instance will be stored (i.e don't use the MIN_HOURS_BETWEEN_SNAPSHOTS + 1) and between each of them add a card with status complete, then at the end, the number of cards in the complete column should be the total number of cards you added that are complete if the last() version was saved for the snapshots. (If that test passes then you can change last() to first() in the model and satisfy yourself that the test fails in the older version of the code). Does this make sense?

backend/curriculum_tracking/models.py
```diff
@@ -1477,20 +1477,32 @@ class BurndownSnapshot(models.Model):
...
+            timestamp__gte=timezone.now()  
             - timedelta(hours=Cls.MIN_HOURS_BETWEEN_SNAPSHOTS),
         ).first()
 
         if match:
             # there was a snapshot taken recently, no need to store another one
-            return match
-
+            # print(f'Match: {match}')
+            # match_1 = Cls.objects.filter(
```
sheenarbw: update the match if it exists

backend/curriculum_tracking/models.py
```diff
@@ -1477,20 +1477,32 @@ class BurndownSnapshot(models.Model):
     def create_snapshot(Cls, user):
         match = Cls.objects.filter(
             user=user,
-            timestamp__gte=timezone.now()
+            timestamp__gte=timezone.now()
```
sheenarbw, line 1576: remove this filter

Andy-Nkumane: the filter continues on the next line

backend/curriculum_tracking/tests/test_BurndownSnapshot.py
```diff
@@ -64,3 +64,71 @@ def test_calculations_correct(self):
...
+
+        factories.AgileCardFactory(
+            content_item=factories.ContentItemFactory(content_type=ContentItem.TOPIC),
+            assignees=[self.user],
+            status=AgileCard.COMPLETE,
+        )
+
+        BurndownSnapshot.create_snapshot(user=self.user)
+        snapshot = BurndownSnapshot.objects.first()
+        snapshot.save()
```
Samantha-Hampton: this version of the snapshot variable doesn't get used before you create the next one so you can probably delete

Samantha-Hampton (approved):
this seems like it's doing what it should

Samantha-Hampton (changes requested):
haha, sorry, I meant to say, except for the minor unnecessary variable I mentioned, this seems to be doing what it should

Samantha-Hampton (approved):

backend/curriculum_tracking/models.py
```diff
@@ -1477,21 +1477,23 @@ class BurndownSnapshot(models.Model):
...
+            update_create = Cls.objects.create
+    
         cards = AgileCard.objects.filter(assignees=user)
         project_cards = cards.filter(content_item__content_type=ContentItem.PROJECT)
 
         complete_cards = cards.filter(status=AgileCard.COMPLETE)
         complete_project_cards = project_cards.filter(status=AgileCard.COMPLETE)
 
-        return Cls.objects.create(
+        return update_create(
```
sheenarbw: Sometimes this will return an instance and sometimes it will return None.

Can we rather have consistent behaviour? Always returning the instance seems like the best bet.

backend/curriculum_tracking/tests/test_BurndownSnapshot.py
```diff
@@ -64,3 +64,69 @@ def test_calculations_correct(self):
...
+                content_type=ContentItem.PROJECT,
+                project_submission_type=ContentItem.LINK,
+            ),
+            assignees=[self.user],
+            status=AgileCard.COMPLETE,
+        )
+
+        BurndownSnapshot.create_snapshot(user=self.user)
+        snapshot = BurndownSnapshot.objects.first()
+        snapshot.save()
```
sheenarbw: What's all this about?

Why not just:

```
snapshot = BurndownSnapshot.create_snapshot(user=self.user)
```

backend/curriculum_tracking/tests/test_BurndownSnapshot.py
```diff
@@ -64,3 +64,69 @@ def test_calculations_correct(self):
...
+        self.assertEqual(snapshot.cards_in_complete_column_total_count, 1)
+        self.assertEqual(snapshot.project_cards_in_complete_column_total_count, 1)
+
+        factories.AgileCardFactory(
+            content_item=factories.ContentItemFactory(content_type=ContentItem.TOPIC),
+            assignees=[self.user],
+        )
+
+        BurndownSnapshot.create_snapshot(user=self.user)
+        BurndownSnapshot.create_snapshot(user=self.user)
```
sheenarbw: why so many calls to create_snapshot? 

Can't we just have one at the beginning and one at the end? 2 in total?

If there is a reason for this can you please put in a comment so that your reasoning is clear.

sheenarbw (changes requested):
It looks like it works fine, there's just a bit of clean code stuff. 

One thing is an inconsistent return, even though it's not a bug now it's likely to cause problems down the line. Inconsistencies are a major cause of bugs

backend/curriculum_tracking/models.py
Andy-Nkumane: @sheenarbw I am not sure how this will return a none - looking at lines 1484-1488, one of two will be used either `Cls.objects.update` or `Cls.objects.create` - is there something I'm missing?

backend/curriculum_tracking/models.py
```diff
@@ -1477,20 +1477,29 @@ class BurndownSnapshot(models.Model):
...
 
         complete_cards = cards.filter(status=AgileCard.COMPLETE)
         complete_project_cards = project_cards.filter(status=AgileCard.COMPLETE)
 
+        if match:
+            # there was a snapshot taken recently, it will be updated
+            return Cls.objects.filter(
+                user=user,
+                timestamp__gte=timezone.now()  
+                - timedelta(hours=Cls.MIN_HOURS_BETWEEN_SNAPSHOTS),
```
sheenarbw: This filter isn't very DRY.

How about just:

```
Cls.objects.filter(pk=match.id)
```

backend/curriculum_tracking/models.py
```diff
@@ -1477,20 +1477,29 @@ class BurndownSnapshot(models.Model):
...
+            ).update(
+                cards_total_count=cards.count(),
+                project_cards_total_count=project_cards.count(),
```
sheenarbw: these update arguments also aren't very DRY. 

How about:

```
snapshot_values = {
       "cards_total_count" : ...
        .....
}
```

Now you can use that in multiple places:

eg:

```
some_query.update(**snapshot_values)

Cls.objects.create(
     user=user,
    **snapshot_values
)
```

sheenarbw (changes requested):
The code is not DRY. There are going to be more upgrades to this snapshot mechanism in the near future, as soon as we update it then someone would need to clean up the unDRY code or just take out more technical debt.

thavha (approved):

sheenarbw (approved):

PR #593, opened 2022-08-26
reduce re-render on team dashboard

Sbonelo01:
Related issues: [RENDER tidy]

## I solemnly swear that:


TheMaja (approved):

Vuyisanani (approved):

sheenarbw (changes requested):
This code doesn't address the issue at all.

frontend/src/components/pages/TeamDashboard/index.jsx
```diff
@@ -46,6 +49,9 @@ function DashboardUnconnected({
     }
   }, [teamId, fetchTeam, team, fetchUserActivityLogDayCountsSequence]);
 
+  if (teams === undefined || !activityLogDayCounts)
```
KhalidUmuzi: Can we apply the same "truthy" logic to teams here?

KhalidUmuzi (changes requested):
This is good, But just one thing

KhalidUmuzi:
Also please merge develop as there are merge conflicts

frontend/src/components/pages/TeamDashboard/index.jsx
```diff
@@ -59,13 +64,19 @@ function DashboardUnconnected({
...
+    });
+ 
   return <Presentation {...props} />;
 }
 
 function mapStateToProps(state) {
   return {
-    teams: state.apiEntities.teams || {},
+    teams: state.apiEntities.teams,
     authUser: state.App.authUser || {},
```
KhalidUmuzi: Does this also need to be changed?

KhalidUmuzi (changes requested):
Check out below

frontend/src/components/pages/TeamDashboard/index.jsx
Sbonelo01: This props does not change - it does not seem to be a cause for any re-renders

Sbonelo01: `authUser` does not seem to be a cause of any re-renders

KhalidUmuzi (approved):

frontend/src/components/pages/TeamDashboard/index.jsx
sheenarbw: Neither would this:

```
    authUser: state.App.authUser || {thisObjectMakesTheCodeMessy: true},
    // unnecessary comments
   bubbles: null
```

If you remove the unnecessary code then the code will be cleaner. If you leave junk lying around just because it doesn't effect execution then it's still junk.

sheenarbw (changes requested):
some unclean code.

KhalidUmuzi (approved):

frontend/src/components/pages/TeamDashboard/index.jsx
```diff
@@ -59,14 +64,20 @@ function DashboardUnconnected({
...
 
 function mapStateToProps(state) {
   return {
-    teams: state.apiEntities.teams || {},
-    authUser: state.App.authUser || {},
+    teams: state.apiEntities.teams,
+    authUser: state.App.authUser,
 
     activityLogDayCounts: Object.values(
       state.apiEntities.activityLogDayCounts || {}
```
sheenarbw:
```
> o = {}
{}
> values = Object.values(o)
[]
> values2 = Object.values(o)
[]
> values === values2
false
```

Can you please always remove things that could recalculate props 

This code would be more appropriate:
```
activityLogDayCounts: state.apiEntities.activityLogDayCounts
```

frontend/src/components/pages/TeamDashboard/index.jsx
```diff
@@ -59,14 +64,20 @@ function DashboardUnconnected({
       userIds: team ? team.members.map((member) => member.userId) : [],
     }),
   };
+  if (activityLogDayCounts.length === 0)
+    getActivityLogCountsByDayForUsers({
+      activityLogDayCounts,
+      eventTypes,
+      userIds: team ? team.members.map((member) => member.userId) : [],
+    });
```
sheenarbw: I don't think this code does anything...

getActivityLogCountsByDayForUsers returns an object. It doesn't change anything. You are computing a value and then doing nothing with it.  What is your intention here?

sheenarbw (changes requested):
There is more work to do. And one confusing thing.

frontend/src/components/pages/TeamDashboard/index.jsx
```diff
@@ -65,12 +71,9 @@ function DashboardUnconnected({
...
   return {
-    teams: state.apiEntities.teams || {},
-    authUser: state.App.authUser || {},
-
-    activityLogDayCounts: Object.values(
-      state.apiEntities.activityLogDayCounts || {}
-    ),
+    teams: state.apiEntities.teams,
+    authUser: state.App.authUser,
+    activityLogDayCounts: state.apiEntities.activityLogDayCounts,
```
KhalidUmuzi, line 76: Surely this breaks the activitylogDayCounts somewhere? We were expecting a list containing the values but we are now passing an object through. this would mean that wherever it is used, is not getting the data in the format it was before.

My recommendation here is to make any change on line 26. Ensure that it retains the format we are expecting else, stuff is going to break

KhalidUmuzi (changes requested):

frontend/src/components/pages/TeamDashboard/index.jsx
```diff
@@ -20,8 +22,10 @@ function DashboardUnconnected({
   let urlParams = useParams() || {};
 
   const teamId = parseInt(urlParams.teamId, 10);
-  const team = teams[teamId];
-  React.useEffect(() => {
+  const team = teams && teams[teamId];
+  activityLogDayCounts = [];
```
KhalidUmuzi: So the way this was before was here

`activityLogDayCounts: Object.values(
      state.apiEntities.activityLogDayCounts || {}
    ),,`
    
Now it is just an empty array here. We need to retain the format it was inside mapStateToProps. There should be logic here to make sure we get the Object.values just like we did before.

KhalidUmuzi (changes requested):

KhalidUmuzi:
@sheenarbw On this particular page, the activityLogDayCounts is being used, but the page that it is being used on is actually not even displaying data any longer.

However in terms of just this dev task specifically, the re-renders should be ok. Am I correct in saying this or should we fix the bug and make it work as well?

TheMaja (approved):

KhalidUmuzi (approved):

sheenarbw (changes requested):
When I run this code then I see an old bug that should have been fixed by now.

Please make sure branches are up to date with the latest develop. Maybe you re-introduced the bug, who knows? I see a bug I'm not expecting to see so I wont merge this.

Best practice has always been to make sure your code is compatible with develop by making sure it is up to date with develop.

Sbonelo01:
render-team-dashboard

Ngoakor12 (approved):
re-renders reduced.

FaithMo (approved):
fixed!
image

sheenarbw (approved):

PR #621, opened 2022-09-25
burn chart skipping date

TheMaja:
- initial commit
- currentUserBurndownStats filter userId using auth
- date duplicates removed from burn chart
- timestamp used to filter date duplicates
- duplicate dates removed from the burn down chat
- util file with gap function created
- dates added on the x-axis
- dates sorted on chart and demo data is displayed
- array split into sub-arrays
- working uncleaned version
- date duplicates removed and dates filled between gap dates

## Description:

## Screenshots/videos
**Chart with activity Dots**
chart with dots
**Current version with activity Dots removed looks clean**
dots removed

## I solemnly swear that:


frontend/src/components/regions/UserActions/UserBurndownStats.jsx
```diff
@@ -23,21 +24,35 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 export default ({ burnDownSnapshots }) => {
+  burnDownSnapshots = burnDownSnapshots.reduce((unique, o) => {
```
Sbonelo01: For readability, Can you please better name the arguments taken by the reduce method? The method takes the current value and adds it to the accumulating value: `reduce(acc, curr)`

`unique` is a good name, but `o` isn't. Please just fix that.

frontend/src/components/regions/UserActions/utils.js
```diff
@@ -0,0 +1,52 @@
...
+  const allDatesArr = [];
+
+  while (firstDate <= lastDate) {
+    allDatesArr.push(new Date(firstDate).toISOString().split("T")[0]);
+    firstDate.setDate(firstDate.getDate() + 1);
+  }
+
+  const getDaysArray = function (start, end) {
+    for (
+      var gapDates = [], firstGapDate = new Date(start);
```
Sbonelo01: Is there a specific reason for using `var`

Sbonelo01 (changes requested):
Well done Themba! This seems to solve the problem - no dates seem to be getting skipped

- The `updateBurnDownSnapShots` function could be broken down further.

KhalidUmuzi (approved):

frontend/src/components/regions/UserActions/utils.js
```diff
@@ -0,0 +1,61 @@
+export function updateBurnDownSnapshots({ burnDownSnapshots }) {
+  const datesArray = [];
+  burnDownSnapshots.forEach((burnDownSnapshot) => {
+    datesArray.push(burnDownSnapshot.timestamp);
+  });
```
sheenarbw, line 4: This is good. I just want to guide you to writing code more efficiently. It'll save you time in the long run

Try using a `map`  to get the datesArray. Where else can map be useful?

frontend/src/components/regions/UserActions/utils.js
```diff
@@ -0,0 +1,61 @@
...
+
+  const uniqueDatesArr = Array.from(new Set(datesArray)).sort(
+    (a, b) => new Date(a) - new Date(b)
+  );
+
+  const firstDate = new Date(new Date(uniqueDatesArr[0]).getTime());
```
sheenarbw: I wonder if this code will error with an empty snapshot array. If the api call that fetches the data hasn't finished then the array might be empty.

Can you add another test that just makes sure nothing breaks when the array is empty?

frontend/src/components/regions/UserActions/UserBurndownStats.jsx
```diff
@@ -49,24 +58,32 @@ export default ({ burnDownSnapshots }) => {
             name="Total Cards"
             stroke={orange[400]}
             activeDot={{ r: 8 }}
+            strokeWidth={1}
+            dot={false}
```
sheenarbw, line 62: can you please include a screenshot in the pr description? I'm not sure what this will look like.

frontend/src/components/regions/UserActions/utils.js
```diff
@@ -0,0 +1,61 @@
+export function updateBurnDownSnapshots({ burnDownSnapshots }) {
```
sheenarbw: I wonder if there is a better name for this function. Always remember that good code makes its intentions clear.

Is this meant to fill in the gaps? Maybe `fillInSnapshotDateGaps` or something like that? 

And then if you are struggling to write self-documenting code, comments can be very helpful

sheenarbw (changes requested):
This is good. There are a few little things worth looking at, mostly I thing there are a few lessons worth learning

frontend/src/components/regions/UserActions/utils.js
KhalidUmuzi: Have you tried to implement sheena's suggestion here? its not in the current changes if so

frontend/src/components/regions/UserActions/utils.js
KhalidUmuzi: Has this been addressed and checked? I see no new tests or changes to ensure this doesnt error out.

KhalidUmuzi (changes requested):

TheMaja:
@KhalidUmuzi  tried using your method by using burnDownSnapshots by removing the datesArray but it breaks everything apparently the datesArray is very important because it keeps the original burnDownSnapshots un-modified and thats very usefull down on the function.

KhalidUmuzi (approved):

sheenarbw (approved):

PR #622, opened 2022-09-27
Update various api serializers to expose total number a card moves to review feedback

Andy-Nkumane:

## Description:

## I solemnly swear that:


backend/curriculum_tracking/serializers.py
```diff
@@ -354,8 +365,18 @@ class Meta:
             "open_pr_count",
             "oldest_open_pr_updated_time",
             "repo_url",
+            "total_number_of_negative_reviews",
```
sheenarbw: maybe a better name would be:

number_of_times_card_moved_to_review_feedback.

If this was really the total number of negative reviews then the call would be very different.

backend/curriculum_tracking/serializers.py
```diff
@@ -326,6 +330,13 @@ def get_users_that_reviewed_open_prs(self, instance):
     def get_users_that_reviewed_open_prs_emails(self, instance):
         return [o.email for o in instance.get_users_that_reviewed_open_prs()]
 
+    def get_total_number_of_negative_reviews(self, instance):
+        total_number_of_negative_reviews = len(
+            LogEntry.objects.filter(effected_user=instance.assignees.first(),
```
sheenarbw: This query gets the total number of times any card belonging to the user was bounced back.

You need to filter by the actual card we are looking at. 

If I have 2 cards:

card_1 bounced 2 times and card_2 bounced 3 times then the total bounces for card_1 is 2 and not 5.

sheenarbw: please write a test

backend/curriculum_tracking/serializers.py
```diff
@@ -932,3 +956,11 @@ def get_reviewer_user_emails(self, instance):
     def get_status(self, instance):
         if instance.agile_card:
             return instance.agile_card.status
+
+    def get_total_number_of_negative_reviews(self, instance):
+        total_number_of_negative_reviews = len(
+            LogEntry.objects.filter(effected_user=instance.recruit_users.first(),
```
sheenarbw: not DRY.  This function has the same problem as the others

sheenarbw (changes requested):
The logic is incorrect. Also it is not DRY. 

Please test your work.  I also think you'll need to update the existing tests so that they pass. The api test mixin will end up failing a bunch of stuff if you add extra fields that retuyrn falsy data by default.  If that sentence doesn't make sense, please run the tests

backend/curriculum_tracking/serializers.py
```diff
@@ -327,6 +331,16 @@ def get_users_that_reviewed_open_prs(self, instance):
     def get_users_that_reviewed_open_prs_emails(self, instance):
         return [o.email for o in instance.get_users_that_reviewed_open_prs()]
 
+    def get_number_of_times_card_moved_to_review_feedback(self, instance):
+        if instance.recruit_project:
```
Samantha-Hampton: maybe we aren't interested in topics for this but there are some topics that require review, so then if they did get a negative review, it wouldn't come up here. Maybe we just need to check to make sure we aren't filtering out data we may want?

Samantha-Hampton (changes requested):
I think it mostly looks ok, but maybe good to write that extra test Sheena was suggesting that checks that the correct values are coming out, eg:

If I have 2 cards:

card_1 bounced 2 times and card_2 bounced 3 times then the total bounces for card_1 is 2 and not 5.

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -447,6 +457,113 @@ def test_set_project_link(self):
...
+        self.assertEqual(response.status_code, 200)
+        card1.refresh_from_db()
+
+        card_moved_to_review_feedback_id = EventType.objects.filter(name=creators.CARD_MOVED_TO_REVIEW_FEEDBACK).first().id
+        c = LogEntry.objects.filter(
+            event_type=card_moved_to_review_feedback_id, 
+            effected_user=card1.assignees.first(),
+            object_1_id=card1.content_item.id,
+        ).last()
+        c.timestamp = c.timestamp - timedelta(days=2)
```
Samantha-Hampton: I'm getting an error when I run the tests:

```
ERROR: test_get_number_of_times_card_moved_to_review_feedback (curriculum_tracking.tests.test_api_views.AgileCardViewsetTests)
  File "/Users/sam/Documents/Umuzi/Products/Tilde/backend/curriculum_tracking/tests/test_api_views.py", line 500, in test_get_number_of_times_card_moved_to_review_feedback
    c = LogEntry.objects.filter(
AttributeError: 'NoneType' object has no attribute 'timestamp'

Ran 70 tests in 116.148s

FAILED (errors=1)
```

So it's not registering a matching log entry object, so there is no timestamp to update

Samantha-Hampton (changes requested):

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -447,6 +457,98 @@ def test_set_project_link(self):
         project.refresh_from_db()
         self.assertEqual(project.link_submission, link_2)
 
+    def test_get_number_of_times_card_moved_to_review_feedback(self):
+        recruit = UserFactory()
```
Samantha-Hampton: extremely minor, but this recruit item does not appear to be used, so can be deleted

Samantha-Hampton (changes requested):

Rokhuda (approved):

backend/curriculum_tracking/serializers.py
```diff
@@ -355,8 +367,21 @@ class Meta:
             "open_pr_count",
             "oldest_open_pr_updated_time",
             "repo_url",
+            "number_of_times_card_moved_to_review_feedback",
         ]
 
+    number_of_times_card_moved_to_review_feedback = serializers.SerializerMethodField("get_number_of_times_card_moved_to_review_feedback")
+
+    def get_number_of_times_card_moved_to_review_feedback(self, instance):
```
sheenarbw, line 375: This really doesn't look DRY to me. Please avoid copy-pasting whole functions.

backend/curriculum_tracking/serializers.py
```diff
@@ -355,8 +367,21 @@ class Meta:
...
+        if instance.recruit_project:
+            number_of_times_card_moved_to_review_feedback = len(
+                LogEntry.objects.filter(effected_user=instance.assignees.first(), 
+                object_1_id=instance.recruit_project.pk,
```
sheenarbw, line 379: This query should be more explicit. 

We can have a recruit project with the same id as a repo. We can have a recruit project with the same id as a review as well.

Please also add object_1's content type to the query. Explicit is better than implicit.

backend/curriculum_tracking/serializers.py
```diff
@@ -928,3 +956,12 @@ def get_reviewer_user_emails(self, instance):
     def get_status(self, instance):
         if instance.agile_card:
             return instance.agile_card.status
+
+    def get_number_of_times_card_moved_to_review_feedback(self, instance):
+        number_of_times_card_moved_to_review_feedback = len(
+            LogEntry.objects.filter(effected_user=instance.recruit_users.first(),
```
sheenarbw, line 965: please avoid copy paste. DRY code is better.

You can add a function to the RecruitProject model. Then just call the function as needed.

There is no reason to introduce this technical debt now.

backend/curriculum_tracking/tests/test_activity_log_entry_creators.py
```diff
@@ -122,6 +122,7 @@ def test_start_project(self):
                 project_submission_type=ContentItem.LINK, template_repo=None
             ),
         )
+        creators.EventType.objects.get_or_create(name=creators.CARD_MOVED_TO_REVIEW_FEEDBACK)
```
sheenarbw: These tests shouldn't need to be updated.

If you make your sql queries more efficient then these changes wont be required.

backend/curriculum_tracking/serializers.py
```diff
@@ -928,3 +956,12 @@ def get_reviewer_user_emails(self, instance):
...
+            object_1_id=instance.pk, 
+            event_type=EventType.objects.get(name='CARD_MOVED_TO_REVIEW_FEEDBACK')),
```
sheenarbw: This query results in multiple SQL queries being fired off. We could mke this more efficient by doing something like:

```
filter(event_type__name=CARD_MOVED_TO_REVIEW_FEEDBACK)
```

backend/curriculum_tracking/tests/test_api_views.py
```diff
@@ -53,6 +55,7 @@ def verbose_instance_factory(self):
         card = factories.AgileCardFactory(recruit_project=project)
         card.reviewers.add(core_factories.UserFactory())
         card.assignees.add(core_factories.UserFactory())
+        creators.EventType.objects.get_or_create(name=creators.CARD_MOVED_TO_REVIEW_FEEDBACK)
```
sheenarbw: These changes shouldn't be needed

sheenarbw (changes requested):
A few things can be made more efficient. And the code is still not DRY

PR #624, opened 2022-09-28
Storybook multi user burndown chart

KhalidUmuzi:

## Description:

Created a new storybook entirely for this one so as to avoid changing the current one. Without doing the integration, the current Burn down Chart will break.

Colours and other CSS is subject to change. 

## Screenshots/videos

image
image
image

## I solemnly swear that:

- [X] I have merged the develop branch into my branch and fixed any merge conflicts


FaithMo (approved):

frontend/src/components/regions/UserActions/NewUserBurndownStats.jsx
```diff
@@ -0,0 +1,197 @@
+import React, { useState } from "react";
```
sheenarbw: This file has a funny name. And it is in a funny place.

Who is this new user of which you speak?

And should this not be visible on one of the Teams tabs?

sheenarbw: It would be good to put this into a directory of it's own, it would do best with a PResentation.jsx file and an index.jsx file

frontend/src/components/regions/UserActions/NewUserBurndownStats.jsx
```diff
@@ -0,0 +1,197 @@
...
+  const authedUserIdSnapshot = burnDownSnapshots.filter(
+    (snapshot) => snapshot.user === authedUserId
+  );
+
+  // organize data from the team, and format it so the chart can use it
+  const usersOnTeam = [];
+  burnDownSnapshots.forEach((snapshot) => {
+    if (
+      !usersOnTeam.some((user) => user.user === snapshot.user) &&
+      snapshot.user !== authedUserId
```
sheenarbw: This should be in index.jsx

frontend/src/components/regions/UserActions/NewUserBurndownStats.jsx
```diff
@@ -0,0 +1,197 @@
...
+
+  burnDownSnapshots.map(
+    (burnDownSnapshot) =>
+      (burnDownSnapshot.timestamp = new Date(burnDownSnapshot.timestamp)
+        .toISOString()
+        .slice(0, 10))
+  );
+  
+  const authedUserId = 1
+  // seperate the authed user and the other users' snapshots
```
sheenarbw: Why?

frontend/src/stories/newBurnDownChart.stories.js
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+import BurnDownChart from "../components/regions/UserActions/NewUserBurndownStats";
```
sheenarbw: Best to be consistent with names when possible.

sheenarbw (changes requested):
On the right track but there are some strange things with naming.

The new storybook conventions I've been experimenting with might be useful to you here.

1. Put the stories in the same directory as the component. Eg: look at src/components/pages/UserCodeReviewPerformance
2. The storybook can draw the unconnected component. This isn't demonstrated in the code, but I think it simplifies things a lot.

eg in index.jsx you would have:

```
export function SomethingSomethingUnconnected({...}){ ....}

...

const SomethingSomething = connect(
  mapStateToProps,
  mapDispatchToProps
)(SomethingSomethingUnconnected);
export default SomethingSomething;

```

Then storybook can draw SomethingSomethingUnconnected

frontend/src/components/pages/MultiUserBurndownChart/Presentation.jsx
```diff
@@ -0,0 +1,142 @@
...
+  LineChart,
+  Line,
+  XAxis,
+  YAxis,
+  CartesianGrid,
+  Tooltip,
+  Legend,
+  ResponsiveContainer,
+} from "recharts";
+import { Typography, Button, Grid } from "@material-ui/core";
```
Ngoakor12: mui imports don't match [styleguide.md](https://github.com/Umuzi-org/Tilde/blob/develop/frontend/STYLE_GUIDE.md#importing-things)

Ngoakor12 (changes requested):
Regarding the Filter metrics buttons:
- Should we not use the one made by @FaithMo in `frontend/src/components/widgets/Button.jsx`?
- Either way I think some padding/space between the filter metric buttons would make a visual improvement

frontend/src/components/pages/MultiUserBurndownChart/Presentation.jsx
```diff
@@ -0,0 +1,147 @@
...
+    marginRight: "1em"
+  },
+  unselectedFilter: {
+    marginRight: "1em"
+  }
+}));
+
+function Presentation({ userSnapshotArray, authedUser }){
+  const [metricFilter, setMetricFilter] = useState("TotalProjects");
+  console.log(userSnapshotArray);
```
Sbonelo01: Please remove this log

Sbonelo01 (changes requested):
This looks good.

remove console.logs and unnecessary comments

frontend/src/components/pages/MultiUserBurndownChart/Presentation.jsx
FaithMo: you missed this log

FaithMo (changes requested):
otherwise, looks good

Sbonelo01:
I have moved this back to review feedback :). You forgot the console.log statement

frontend/src/components/regions/UserActions/UserBurndownStats.jsx
```diff
@@ -31,14 +31,15 @@ export default ({ burnDownSnapshots }) => {
...
   );
-
   const filterStartDate = new Date();
   filterStartDate.setDate(filterStartDate.getDate() - 21);
 
   const burnDownSnapshotsFilledDates = fillInSnapshotDateGaps({
     burnDownSnapshots,
   });
 
+  console.log(burnDownSnapshots);
```
Ngoakor12: you forgot this console log

Ngoakor12 (changes requested):
I get this error when running trying to run the story:
image

Also get other errors even after importing the function from the right directory

sheenarbw (commented):
The changes to package-lock.json are concerning. Are you sure you meant to change that?

frontend/src/components/pages/MultiUserBurndownChart/Presentation.jsx
```diff
@@ -0,0 +1,152 @@
...
+    color: "green",
+    marginRight: "1em"
+  },
+  unselectedFilter: {
+    marginRight: "1em"
+  }
+}));
+
+function Presentation({ userSnapshotArray, authedUser }){
+  const [metricFilter, setMetricFilter] = useState("TotalProjects");
```
sheenarbw: useState in a presentation file? Are you sure?

frontend/src/components/pages/MultiUserBurndownChart/Presentation.jsx
```diff
@@ -0,0 +1,152 @@
...
+  const classes = useStyles();
+  return (
+    <React.Fragment>
+      <Typography variant="h6" component="h2">
+        Burndown
+      </Typography>
+      <Typography variant="h6" component="h1">
+        Filter Metric:{" "}
+        <Button
```
sheenarbw: There will come a time when the metrics change, eg we will want to display total weights, not just total counts.

If we were to introduce one new metric then a whole lot of code would need to change.

How can you make this more DRY?

Something like:

```
const metrics = {
    cardsInCompleteColumnTotalCount: "Complete Cards",
   someOtherDataKey: "User Friendly Title",
}

...

# draw the buttons:

{
    Object.keys(metrics).map(metric => (
           <Button ....>
    ))
}
...
 
# draw the lines:


                  {userSnapshotArray.map((userOnTeam) => (
                    <Line
                      type="monotone"
                      data={userOnTeam.snapshot}
                      dataKey=${selectedDataKey}   ################## why all the if statements? 
                      name={`${userOnTeam.userEmail}`}
                      stroke={userOnTeam.user === authedUser ? green[400] : red[400]}
                      activeDot={{ r: 8 }}
                      strokeWidth={1}
                      dot={false}
                      key={userOnTeam.user}
                    />
                  ))}
                


```

frontend/src/components/pages/MultiUserBurndownChart/index.jsx
```diff
@@ -0,0 +1,76 @@
...
+  //   mapStateToProps
+  currentTeamBurndownStats,
+  userTeam,
+  // mapDispatchToProps
+
+  // storybook
+  forceUser,
+}) {
+  let urlParams = useParams() || {};
+  const authedUser = forceUser || parseInt(urlParams.userId);
```
sheenarbw: This won't work. The user id will be undefined since this will be on a team page and not a user page. 

Maybe just stop trying to highlight one specific user. I'm not sure why you want to do that anyway, it's not a requirement, it's making the code more complicated, and it wont even work in context since we are looking at a team.

This line also doesn't make sense. Are you saying that if there is a userId in the url then the current loggede in user is always that user? Is this just bad naming conventions? Or is it a misunderstanding.

If I log in and my userid is 2 then  the authenticated user has an id of 2.
If I go and look at your board and you have an id of 512, then the authenticaed user still has an id of 2.

frontend/src/components/pages/MultiUserBurndownChart/utils.js
```diff
@@ -0,0 +1,44 @@
+export function fillInSnapshotDateGaps({ currentUserBurndownStats }) {
```
sheenarbw, line 1: this looks complicated. Have you considered tests?

sheenarbw (changes requested):
This can be more DRY, and has a bug

frontend/src/components/pages/MultiUserBurndownChart/Presentation.jsx
KhalidUmuzi: Is it incorrect to use the useState hook in the Presentation file?

KhalidUmuzi:
> The changes to package-lock.json are concerning. Are you sure you meant to change that?

I'm unsure why this is being included. I've merged develop into my branch and it is still showing changes I didn't even stage or commit.

backend/curriculum_tracking/serializers.py
```diff
@@ -750,7 +750,6 @@ class Meta:
             "project_cards_in_complete_column_total_count",
         ]
 
-
 class ReviewTrustSerializer(serializers.ModelSerializer):
     class Meta:
         model = models.ReviewTrust
```
Sbonelo01: Should this file be in the PR?

Sbonelo01 (changes requested):
This works as expected, Aside from the `package-lock.json` file, there is also backend file that's also in this branch.

Ngoakor12 (changes requested):
Looks good...

- frontend tests are failing because:
  - `fillInSnapshotDateGaps` is not being called correctly in the tests
  - `removeDuplicates` either doesn't exist or is being imported from the wrong file
image

- Seeing which line belongs to which learner is a bit difficult. Is it not possible to differentiate each learner with a different colour? maybe because we don't know how many learners would be in each group?

KhalidUmuzi:
> * Seeing which line belongs to which learner is a bit difficult. Is it not possible to differentiate each learner with a different colour? maybe because we don't know how many learners would be in each group?

It was like that initially, but this is going to be viewed on the team Dashboard, so separating each user gets complicated since we dont know how many users on a team

Ngoakor12 (approved):

frontend/src/components/pages/MultiUserBurndownChart/MultiUserBurndownChart.stories.jsx
```diff
@@ -0,0 +1,29 @@
...
+import multiUserBurndownData from "./storyData/multiUserBurnDownData.json";
+import { MultiUserBurndownChartUnconnected } from ".";
+
+export default {
+  title: "Tilde/pages/MultiUserBurndownChart",
+  component: MultiUserBurndownChartUnconnected,
+};
+
+const Template = (args) => (
+  <Router>
```
sheenarbw: I'm not convinced that you need a router here... are you linking to something?

frontend/src/components/pages/MultiUserBurndownChart/index.jsx
```diff
@@ -0,0 +1,80 @@
+import React, {useState} from "react";
+import Presentation from "./Presentation";
+import { connect } from "react-redux";
+import { fillInSnapshotDateGaps } from "./utils";
+
+export function MultiUserBurndownChartUnconnected({
+  //   mapStateToProps
+  currentTeamBurndownStats,
+  team,
+  metrics,
```
sheenarbw: This is very misleading... Do you think the state will work like that?

Open up the state object in the redux dev tools and look at how teams are represented. 
And why would metrics be in there at all?  That's very random.

mapStateToProps only has stuff that comes from:
- the apiEntities. This has a very standard format that is always the same
- api logs: you aren't using that here. so it's irrelevent

frontend/src/components/pages/MultiUserBurndownChart/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+  // storybook
+}) {
+  const [metricFilter, setMetricFilter] = useState("projectCardsTotalCount");
+
+  const handleChangeMetricFilter = (newValue) => {
+    setMetricFilter(newValue);
+  };
+  
+  currentTeamBurndownStats = currentTeamBurndownStats
+    ? Object.values(currentTeamBurndownStats)
```
sheenarbw, line 26: This looks like a bug. You are saying that if it exists, then make an Array, otherwise initialize it to an empty object. The moment we connect this to real data it will break.

frontend/src/components/pages/MultiUserBurndownChart/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+  currentTeamBurndownStats.map(
+    (burnDownSnapshot) =>
+      (burnDownSnapshot.timestamp = new Date(burnDownSnapshot.timestamp)
+        .toISOString()
+        .slice(0, 10))
+  );
+  // organize data from the team, and format it so the chart can use it
+  const userSnapshotArray = [];
+  currentTeamBurndownStats.forEach((snapshot) => {
+    if (!userSnapshotArray.some((user) => user.user === snapshot.user)) {
```
sheenarbw, line 37: This code is quite involved. Do you think it might warrent some tests?

frontend/src/components/pages/MultiUserBurndownChart/storyData/multiUserBurnDownData.json
```diff
@@ -0,0 +1,112 @@
...
+      "user": 2,
+      "timestamp": "2022-03-14",
+      "cardsTotalCount": 11,
+      "projectCardsTotalCount": 3,
+      "cardsInCompleteColumnTotalCount": 0,
+      "projectCardsInCompleteColumnTotalCount": 0
+    }
+  },
+  "team": {
+    "id": 29,
```
sheenarbw, line 86: This wouldn't be in the state. But you say it'll come from mapStateToProps

frontend/src/components/pages/MultiUserBurndownChart/storyData/multiUserBurnDownData.json
```diff
@@ -0,0 +1,112 @@
...
+      },
+      {
+        "userId": 4,
+        "userEmail": "someone.cool@example.com",
+        "userActive": true
+      }
+    ]
+  },
+  "metrics": {
+    "projectCardsTotalCount": "Total Projects",
```
sheenarbw: This wouldn't be in the state. But you say it'll come from mapStateToProps

frontend/src/components/pages/MultiUserBurndownChart/Presentation.jsx
```diff
@@ -0,0 +1,99 @@
...
+                  dot={false}
+                  key={userOnTeam.user}
+                />
+              ))}
+            </LineChart>
+          </ResponsiveContainer>
+        </Grid>
+        <Grid item xs={2}>
+          <Typography variant="h6" component="h2">
+            Cohort users
```
sheenarbw: Are you sure we will only use this for looking at cohorts? Maybe just "users" would be better?

frontend/src/components/pages/MultiUserBurndownChart/index.jsx
```diff
@@ -0,0 +1,80 @@
...
+export function MultiUserBurndownChartUnconnected({
+  //   mapStateToProps
+  currentTeamBurndownStats,
+  team,
+  metrics,
+  // mapDispatchToProps
+
+  // storybook
+}) {
+  const [metricFilter, setMetricFilter] = useState("projectCardsTotalCount");
```
sheenarbw, line 14: This is an interesting choice of initial metric. Do you think it'll be particularly interesting to watch the total number of projects as they change over time? Or do you think that staff members would be more interested in a different metric as the default?

sheenarbw (changes requested):
There are some funky data things happening. Better to be consistent.

Also, the ux isn't amazing:

1. What is this actually for? There is a key at the bottom of the graph that lists all the email addresses. Why list them twice if there is no functionality here?
image

2. Everyone's graph is the same colour:

image


It would be way better if each line was a different colour so you could read the graph without having to hover over every data point. 


3. These buttons are pretty chunky:

image


I think it would be good to get rid of the word "BurnDown" and maybe even "Filter Metric". Why say "filter metric" anyway? Consider the user. What will they think those words mean? We want to pick which metric to display. 

I think MUI allows for smaller buttons. I think you can mark buttons as dense or compact or something like that to shrink them.

frontend/src/components/pages/MultiUserBurndownChart/index.jsx
KhalidUmuzi: So I was thinking of allowing us to use the metrics from the backend rather than hardcoding them in the frontend. The reason is if we want to add new metrics in the future, all we need to do is update the DB with new metrics. If we hardcode it on the frontend, then whenever we do want to add new metrics, we need to deploy a new version of the code.

@sheenarbw Is my thinking alright here or would we prefer to just hardcode the metrics on the frontend and leave it like that?

TheMaja (changes requested):
- looks good but one of ther tests is failing.
image

PR #628, opened 2022-10-06
Add review modal upgrade

Ngoakor12:
Related issues: [[Upgrade the “Add Review” modal](https://www.notion.so/Upgrade-the-Add-Review-modal-e5370ceca4d04dd187685518f250b2f3)]

## Description:

## Screenshots/videos
image
Screenshot from 2022-10-12 09-48-10
image

## I solemnly swear that:


FaithMo (approved):

FaithMo (approved):
nice!

KhalidUmuzi (approved):

frontend/src/components/regions/AddCardReviewModal/index.jsx
```diff
@@ -33,19 +43,36 @@ function AddReviewModalUnconnected({
...
+      status: false,
+    }));
+    const { name, value } = e.target;
+    setFormValues((prevFormValues) => ({ ...prevFormValues, [name]: value }));
+  };
+
   const handleSubmit = (e) => {
     e.preventDefault();
+
+    if (formValues.status === "")
```
sheenarbw: This is a strange way to do things. You are doing the validation on the frontend and then even if you know that the form is invalid, the api call gets made.

There are 2 general approaches to these sorts of things:

1. Do the validation on the frontend, and then if the frorm looks incorrect then don't make the api call. 
This is nice because it stops you from making unnecessary API calls. BUT it's also completely terrible because it means that the frontend and the backend both need to implement the same validation logic which is not at all DRY.


2. Make the api call, allow the backend to do form validation, and display any error messages that show up.
This one means that validation logic only needs to be implemented once. And if the backend's error responses are highly consistent and predictable then the frontend just needs to show the errors as they appear. We are using Django Rest Framework so the error structure is very consistent.  If there is a failed api call then we just need to display the messages that came from the server.


Let's go with 2.

sheenarbw (changes requested):
A good attempt.

frontend/src/components/regions/AddCardReviewModal/Presentation.jsx
```diff
@@ -83,7 +83,7 @@ export default function Presentation({
           </Grid>
         </Grid>
 
-        {hasFailedApiResponse && (
+        {latestApiCallStatus && latestApiCallStatus.responseOk === false && (
           <Alert severity="error" className={classes.alert}>
             An error occured while trying to submit your review
```
FaithMo: please fix the spelling of "occurred"

FaithMo (changes requested):
comment: would it not be better to remove the error after the user chooses one of the options. example, if they attempt to submit without choosing a status, but after making a choice, the error that was displayed gets removed? What do you think?
image

Ngoakor12:
> comment: would it not be better to remove the error after the user chooses one of the options. example, if they attempt to submit without choosing a status, but after making a choice, the error that was displayed gets removed? What do you think? image

Would that not be frontend validation though? I think it contradicts Sheena's point of not doing the validation on the frontend and backend.

FaithMo (approved):

KhalidUmuzi (approved):

sheenarbw (approved):

PR #631, opened 2022-10-12
Bug: team dashboard: graphs aren’t populating any more

FaithMo:
Related issues: [Bug: team dashboard: graphs aren’t populating any more](https://www.notion.so/Bug-team-dashboard-graphs-aren-t-populating-any-more-04fa47dd2e41485a83c692925b74a41a)

## Description:

## Screenshots/videos
image

## I solemnly swear that:


Ngoakor12 (changes requested):
Tilde graphs seem to be fixed :ok_hand: 
The storybook representing it on the other hand does not display graphs even though it probably should.

FaithMo:
> Tilde graphs seem to be fixed ok_hand The storybook representing it on the other hand does not display graphs even though it probably should.

the dummy data for the storybook is far back from the date that's being filtered. the filter date is 21 days ago and the dummy data for the storybook is last year. so only days on or after 21 days can be displayed

Ngoakor12:
> the dummy data for the storybook is far back from the date that's being filtered. the filter date is 21 days ago and the dummy data for the storybook is last year. so only days on or after 21 days can be displayed

Ohhh makes sense, updating the times in the storybook json files to fit in the 21 day window is not a task one wants to manually do just for the data to display :sweat_smile:

Ngoakor12 (approved):

KhalidUmuzi:
Yea agreed. We do however need to discuss a better way of viewing data that is dynamic and that changes based on time such as this one. Perhaps we should instead create a function that spits out some dummy data for us based on the current day.

What do you guys think of that?

Ngoakor12 (approved):

KhalidUmuzi (approved):

frontend/src/apiAccess/selectors/activityLogSelectors.js
```diff
@@ -37,11 +37,12 @@ export const getActivityLogCountsByDayForSingleUser = ({
   const resultByDate = {};
 
   eventTypes.forEach((eventType) => {
-    const userLogsFilteredByType = userLogs.filter(
-      (element) =>
-        element.id.match(`event_type__name=${eventType}&`) ||
-        element.id.match(`event_type__name=${eventType}$`)
-    );
+    const userLogsFilteredByType = userLogs.map((type) => {
```
sheenarbw: This code was confusing. Why did you choose to name a logEntry a type?

If we have an array of apples and we want to do a map on the apples then the code would look like this

```
apples = [apple1,apple2, apple3]
apples.map( (apple, index)=>{whatever})
//                      ^^^^
//                     every element is an apple
```

If we are looping over log entries then every element is not a type. It is a log entry.

If you were a carpenter then would you call a nail a hammer? No, it would confuse people. Call things what they are.

frontend/src/utils/randomDateBetweenCurrentDayAnd21DaysAgo.js
```diff
@@ -0,0 +1,17 @@
+export function randomDateBetweenCurrentDayAnd21DaysAgo() {
```
sheenarbw: This is defined in a ery funny place. It's only used in the story, right? Maybe it should be with the story.

That said, maybe it shouldn't exist at all. Stories should be pretty predictable. 

Consider using a generator to yield a non-random sequence of dates.

sheenarbw (changes requested):
On the right track, but names are important.

There's also some things about predictability and cohesion

Ngoakor12 (approved):

KhalidUmuzi (approved):

sheenarbw (approved):

PR #641, opened 2022-11-08
Making sure log pr opened gets called correctly

Andy-Nkumane:

## Description:

## I solemnly swear that:


Andy-Nkumane:
Currently, I don't have a fix for object 1

Samantha-Hampton:
> Currently, I don't have a fix for object 1

@Andy-Nkumane I don't understand, why can't object_1 = pull_request
and is there a reason we don't have object_2? Should object_2 not be the repo concerned? or am I missing something?

backend/git_real/activity_log_creators.py
```diff
@@ -46,7 +46,7 @@ def log_pr_opened(pull_request):
     match = LogEntry.objects.filter(
         actor_user=pull_request.user,
         effected_user=pull_request.user,
-        object_1=pull_request,
+        # object_1=pull_request,
```
Samantha-Hampton: do you also get this error if you have object_1 = pull_request?:
```
django.core.exceptions.FieldError: Field 'object_1' does not generate an automatic reverse relation and therefore cannot be used for reverse querying. If it is a GenericForeignKey, consider adding a GenericRelation.
```

Samantha-Hampton (commented):
I've been trying different things but the git_real models, etc are hurting my brain lol. Maybe we need some guidance from Sheena since I assume we do need to have object_1 = pull_request?

Also wondering why we don't have object_2 as the repo like we do for log_push_event...do you know why?

backend/git_real/activity_log_creators.py
Andy-Nkumane: yes, that is the main reason i commented it out

Andy-Nkumane: In terms of object 2, I'm not fully sure about it - sometimes it is there and sometimes it isn't

Samantha-Hampton (approved):
I've asked Sheena about the object 2 thing, but otherwise seems like it is doing what it should...

backend/git_real/tests/test_activity_log_creators.py
```diff
@@ -102,4 +104,37 @@ def test_pushing(self, has_permission):
...
+
+
+        self.assertEqual(LogEntry.objects.count(), 1)
+
+        self.assertEqual(entry.actor_user, pull_request.user)
+        self.assertEqual(entry.effected_user, pull_request.user)
+        self.assertEqual(entry.object_1, pull_request)
+        self.assertEqual(entry.object_2, project)
+        self.assertEqual(entry.event_type.name, creators.PR_OPENED)
+        self.assertEqual(pull_request.state, 'open')
```
Samantha-Hampton: So I made a breakpoint() above here since I wanted to verify if there were any other projects associated with the repo (because it would be good to make sure that the correct project is associated with object 2 in the case of more than one project per repo) but I discovered something interesting when running line by line:

`entry.actor_user == pull_request.user == None` (there is no return value so doesn't seem like there is anything associated with user)

similarly for
`entry.object_2 == project == None`
`entry.effected_user == pull_request.user == None`

So concerned the tests just passing because all those values are just null instead of actually verifying that they are the correct log entry item...

Samantha-Hampton (changes requested):

Samantha-Hampton (approved):

backend/git_real/activity_log_creators.py
```diff
@@ -42,20 +44,24 @@ def log_pr_closed(pull_request):
...
-
     match = LogEntry.objects.filter(
         actor_user=pull_request.user,
         effected_user=pull_request.user,
-        object_1=pull_request,
         event_type=event_type,
         timestamp__gte=timezone.now() - timezone.timedelta(minutes=2),
-    ).first()
-
+        object_1_content_type = ContentType.objects.get_for_model(pull_request)
```
sheenarbw: A `get` would be more appropriate here.  Would there not always be one content type for one model?

backend/git_real/activity_log_creators.py
```diff
@@ -42,20 +44,24 @@ def log_pr_closed(pull_request):
...
+        ).first()
+    
     if match == None:
         LogEntry.objects.create(
             actor_user=pull_request.user,
             effected_user=pull_request.user,
             object_1=pull_request,
+            object_2= curriculum_tracking.models.RecruitProject.objects.filter(
+                repository=pull_request.repository,
+            ).order_by("pk").last(),
```
sheenarbw: This isn't accurate. 

Eg say a learners has a part 1 and a part 2 card on their board, both of these cards will refer to the same repo. 
Now a RecruitPRoject instance is created when a card is started OR when a due date is given to a card. 

So if Jacob goes and adds a deadline to part 2, every single pr_opened event will be associated with part 2 even if the learner is working on part 1.

If would probably be better to look at timestamps on RecruitProject instances or something like that.

backend/git_real/models.py
```diff
@@ -123,8 +126,34 @@ def create_or_update_from_github_api_data(cls, repo, pull_request_data):
         pull_request, _ = cls.get_or_create_or_update(
             repository=repo, number=number, defaults=defaults, overrides=defaults
         )
+
+        if pull_request_data['base']['repo']['pushed_at'] == pull_request_data['updated_at']:
```
sheenarbw: This seems complicated.

Look at this file: git_real/tests/assets/pull_request_opened_request_body.json 

The very first attribute is "opened". This means that the pull request was just opened.   You don't need to do anything fancy, you just need to look at the right data. 

Maybe the function that calls `create_or_update_from_github_api_data`  has access to that event data? Maybe it can make a good decision about what event should be logged?

sheenarbw: This comparison also seems odd. Is this an accurate way to check if it's a push event? 

What if a person pushes code, scratches their nose, and then only opens a PR? 
Is `pushed_at` equal to the last push of any branch, or the last push on the main branch? 

It seems like there are a lot of assumptions happening here. It would be better to just look at the json passed to the webhook.

backend/git_real/tests/test_activity_log_creators.py
```diff
@@ -102,4 +106,43 @@ def test_pushing(self, has_permission):
...
+
+        self.assertEqual(PullRequest.objects.count(), 0)
+
+        body, headers = get_body_and_headers("pull_request_opened")
+        pull_request_data=body["pull_request"]
+        url = reverse(views.github_webhook)
+
+        github_name = pull_request_data["user"]["login"]
+        SocialProfileFactory(github_name=github_name)
+        user = User.objects.filter(social_profile__github_name__iexact=github_name).first()
```
sheenarbw: this query seems very complicated.

Why not:

wouldn't this work?
```
social_profile = SocualPRofileFactory(...)
user = social_profile.user
```

or even `User.objects.get(social_profile = social_profile)`

backend/git_real/tests/test_activity_log_creators.py
sheenarbw: It looks like this conversation is not resolved.

How about asserting that the user is not None?  The actor user and effected user need to both be defined.

sheenarbw (changes requested):
There's a little bit of weird logic in here. And it looks like Sam raised a concern that wasn't addressed.

backend/git_real/activity_log_creators.py
```diff
@@ -42,20 +44,34 @@ def log_pr_closed(pull_request):
...
+        recruit_projects = curriculum_tracking.models.RecruitProject.objects.filter(
+                repository=pull_request.repository,
+            )
+        recruit_project_start_time = None
+        for recruit_project in recruit_projects[::-1]:
+            recruit_project_start_time = recruit_project.start_time
+            if not recruit_project_start_time:
+                continue
+            if pull_request.created_at >= recruit_project_start_time:
+                break
```
Samantha-Hampton, line 67: This doesn't really make sense to me, I'm thinking about some of those Cloudbnb projects, where there are multiple projects on one repo and I've looked at a learners PRS where there are 2 PRs open, one for Project part 1 and one for project part 2. So this would mean that the date/time a PR is created could be after the start time of both parts of the project so how would we know we are linking the correct project to the opened PR?
(if you weren't allowed to start the next part of the project until the previous part was finished then this would work but not all projects are setup like that....)

Samantha-Hampton (changes requested):
Let me know if my other comment doesn't make sense?

Perhaps the tests should be expanded for the case of more than one project in a repo with a few PRs on each to see that we can ensure that the log_pr_open is associated with the right project....

backend/git_real/activity_log_creators.py
Andy-Nkumane: @Samantha-Hampton I agree with that but recruits tend to create new pr instead of opening the closed pr. It won't always work and also currently there shouldn't be a continuation project that allows two cards to be worked on - with the cloudbnb the learners that are doing are doing double the work as some of the code is repeated and when merged merge conflicts are created, thus doing that is not ideal. Continuation projects from what i have seen so far should not be worked on simultaneously.

Samantha-Hampton: @Andy-Nkumane yes and also there are edge cases, eg, whereby a learner may get part 1 into complete, start on part 2, and then get a re-review on part 1 and end up in the same predicament if a new PR is made for part 1. So I guess there needs to be a different way to link which part of a continuation project is associated with a PR. Maybe we need to ask Sheena if there is another way to think about it....

sheenarbw: We will need to guess a little bit here. 

```
def get_card(): 
    all_cards = find all AgileCards that could be associated with that PR (the repo->project->card)
    for status in [REVIEW_FEEDBACK, IN_PROGRESS, IN_REVIEW, COMPLETE,BACKLOG,BLOCKED]:  # this order matters
         filtered_cards = all_cards.filter(status=status)
         if len(filtered_cards):
               order the cards by time spent in the column 
               return oldest? Or maybe this is different per column. 
```

The reason for the ordering is:
- REVIEW_FEEDBACK: it's more important to address feedback than work on new code, this assumes the learners are being good
- IN_PROGRESS: this is the next column the learner should be working on. After this point it gets weird
- IN_REVIEW: maybe a learner has a conversation with a reviewer and pushes more code or something. Who knows.

The rest of the columns are included so that if anything truly weird happens at least the activity will be associated with something. The rest of the columns are ordered according to weirdness, in ascending order.

And then we will just need to keep an eye on things. This log might tell us a bit about how the learners behave when it comes to prioritising code.

A completely different alternative would be to just associate the event with the repo, we would lose information though. So it's either guess a little, or lose resolution as far as I can see.

PR #649, opened 2022-12-13
Add tests on pull request template

FaithMo:
Related issues: [URGENT: Unit test failing](https://www.notion.so/URGENT-unit-tests-failing-0e9ac4d3e32e42199f71150740d0f3bc)

## Description:

## I solemnly swear that:


Ngoakor12 (commented):
I think this is already covered in the last point in the checklist, no?

FaithMo:
> I think this is already covered in the last point in the checklist, no?

this is if you have added tests to any feature you have done, but the one I added makes emphasis on tests that either exist before your feature or that you have added and do not break something ...

Ngoakor12:
> > I think this is already covered in the last point in the checklist, no?
> 
> this is if you have added tests to any feature you have done, but the one I added makes emphasis on tests that either exist before your feature or that you have added and do not break something ...

But this means we'll now have three points on the checklist related to tests... I don't think it's bad but maybe it's redundant and can be summarised better?
- the first point about tests is for adding tests 
- the second test related point is about if both the old and new pass

Would changing the existing last point to just  "**New and existing unit tests pass locally**" be better than adding a slightly similar point? I think it covers new and old test well

FaithMo:
> > > I think this is already covered in the last point in the checklist, no?
> > 
> > 
> > this is if you have added tests to any feature you have done, but the one I added makes emphasis on tests that either exist before your feature or that you have added and do not break something ...
> 
> But this means we'll now have three points on the checklist related to tests... I don't think it's bad but maybe it's redundant and can be summarised better?
> 
> * the first point about tests is for adding tests
> * the second test related point is about if both the old and new pass
> 
> Would changing the existing last point to just "**New and existing unit tests pass locally**" be better than adding a slightly similar point? I think it covers new and old test well

makes sense. otherwise just leave it as is? cause it covered the same thing?

pull_request_procedure.md
```diff
@@ -9,8 +9,9 @@ When opening a Pull request, there is a checklist of points that appears on the
 - [ ] My code follows the style guidelines of this project
 - [ ] I have merged the develop branch into my branch and fixed any merge conflicts
 - [ ] I have performed a self-review of my own code
+- [ ] I have commented my code in hard-to-understand areas
 - [ ] My changes generate no new warnings
-- [ ] New and existing unit tests pass locally with my changes
+- [ ] I have added tests that prove my fix is effective or that my feature works
 - [ ] I have tested new or existing tests and made sure that they pass
```
Ngoakor12, line 15: using "or" on the "new or existing" part of the sentence makes it sounds like you should choose one or the other

Ngoakor12 (changes requested):

pull_request_procedure.md
FaithMo: i thought of it as "I have tested new tests" (if I created new tests) or "I have tested existing tests" (if I have not added any tests, to make sure nothing breaks)

Ngoakor12 (approved):

sheenarbw (approved):

PR #650, opened 2023-05-15
The components folder has been converted to TypeScript.

TheMaja:

## Description:

## I solemnly swear that:


frontend-mini-challenge/components/LoggedOutPage.tsx
```diff
@@ -23,8 +23,17 @@ export default function Page({ children }) {
   });
 
   return (
-    <AppShell padding="md" header={<Header height={60} p="xs"></Header>}>
+    <AppShell
+      padding="md"
+      header={
+        <Header height={60} p="xs">
+          <Container></Container>
```
sheenarbw: Why is there a container in the header now?

frontend-mini-challenge/components/LoggedInPage.tsx
```diff
@@ -21,6 +21,11 @@ export default function Page({
   serverSidePropsCorrectlyCalled,
   isLoggedIn,
   loggedInUserData,
+}: {
+  children: React.ReactNode;
+  serverSidePropsCorrectlyCalled: boolean;
+  isLoggedIn: boolean;
+  loggedInUserData: any;
```
sheenarbw: `loggedInUserData` shouldn't be `any`

frontend-mini-challenge/components/LoggedInPage.tsx
```diff
@@ -131,8 +144,14 @@ Then when you make use of this page component:
     </Page>
 
 */
-export async function getServerSidePropsForLoggedInPage({ query, req }) {
-  const whoAmIResponse = await serverSideWhoAmI({ query, req });
+export async function getServerSidePropsForLoggedInPage({
+  query,
+  req,
+}: {
+  query: any;
```
sheenarbw: These shoulnd't be `any`

frontend-mini-challenge/components/LoggedInPage.tsx
```diff
@@ -131,8 +144,14 @@ Then when you make use of this page component:
...
+  req: any;
+}) {
+  const whoAmIResponse = await serverSideWhoAmI({ query, req } as any);
```
sheenarbw: also not `any`

frontend-mini-challenge/tsconfig.json
```diff
@@ -15,6 +15,6 @@
     "jsx": "preserve",
     "incremental": true
   },
-  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx"],
+  "include": ["next-env.d.ts", "**/*.ts", "**/*.tsx", "components/LoggedOutPage.tsx"],
```
sheenarbw: Look at the existing patterns here. Do you think "components/LoggedOutPage.tsx" is actually helpful? Or do you think it is already included?

frontend-mini-challenge/components/Text.tsx
```diff
@@ -1,14 +1,14 @@
...
+export function Bold({ children }: { children: React.ReactNode }) {
   return (
     <Text fw={700} component="span">
       {children}
     </Text>
   );
 }
 
-export function Underlined({ children }) {
+export function Underlined({ children }: { children: React.ReactNode }) {
```
sheenarbw: Look at how these functions are used. Are you sure they take in any ReactNode or are they called with one specific type of node

sheenarbw (changes requested):
This is a good start but has a few rough edges:

Always try to be explicit:

1. Use of `any`. Please try not to use `any`. Rather be explicit. Otherwise when a new coder works on this stuff they will be encouraged to guess and there will be errors that come from that. 
2. Use of ReactNode when we can refer to a specific type. 

Make sure you aren't  making changes you don't meant to make. There are a few strange things in here.

I haven't actually run the code and looked for errors. Can you please check yourself for errors by running
- `npm run build`
- the storybook tests

The storybook tests use this https://storybook.js.org/docs/react/writing-tests/test-runner 

You can run them by opening two terminals:

```
# terminal 1

npm run storybook

# terminal 2

npm run test-storybook
```

frontend-mini-challenge/components/LoggedInPage.tsx
```diff
@@ -131,13 +145,17 @@ Then when you make use of this page component:
...
 
 */
-export async function getServerSidePropsForLoggedInPage({ query, req }) {
-  const whoAmIResponse = await serverSideWhoAmI({ query, req });
+
+export async function getServerSidePropsForLoggedInPage({
+  query,
+  req,
+}: GetServerSidePropsContext) {
+  const whoAmIResponse = await serverSideWhoAmI({ query, req } as GetServerSidePropsContext);
```
sheenarbw: This seems like a lot of guessing...

As far as I know,  GetServerSidePropsContext looks like this:

```
 export interface GetServerSidePropsContext {
  req: IncomingMessage
  res: ServerResponse
  params?: ParsedUrlQuery
  query: ParsedUrlQuery
  preview?: boolean
  previewData?: any
}
```

serverSideWhoAmI returns something completely different to that. Just because it is used in server side things doesn't mean that the type is this random thing.

GetServerSidePropsContext is used like this only:

```
getServerSideProps = async ({ arguments,you,want }: GetServerSidePropsContext) => {
...
```

In other words, the arguments sent to a getServerSideProps function will adhere to that interface. 


If you look at the return below, you can see these lines:

```
    isLoggedIn: whoAmIResponse.status === 200,
    loggedInUserData: whoAmIResponse.responseData,
```
This tells us that the whoAmIResponse at least has a `status` that can be a number, and `responseData`. The `responseData` would be a JSON response. So basically an object.

sheenarbw (changes requested):
You have some guesswork in here. Specifically, your use of GetServerSidePropsContext is dodgy.

**In the face of ambiguity, refuse the temptation to guess** 

If you put guesses in your code then it will be hard for me to trust your code. Every time I've put in the up-front effort to understand things properly it's worked out well for me. Remember that part of the reason you are doing this work is so that you can level up and help others to level up.

One thing I'm not sure of myself if the difference between JSX.Element, ReactElement and ReactNode. So I'm going to need to research that before reviewing everything properly. If you feel like you made any guesses there then please revise that.

frontend-mini-challenge/components/Alerts.tsx
```diff
@@ -17,12 +22,10 @@ export function ErrorAlert({ children, title }) {
   );
 }
 
-export function InfoAlert({ children, title, Icon }) {
-  Icon = Icon || InfoIcon;
```
sheenarbw, line 21: This line is here on purpose. 

If Icon is passed in then use that, otherwise use InfoIcon as the default. 

Please don't remove functionality. Feel free to refactor, but don't just delete code.

frontend-mini-challenge/components/LoggedInPage.tsx
```diff
@@ -15,12 +15,18 @@ import { useEffect } from "react";
...
 export default function Page({
   children,
   serverSidePropsCorrectlyCalled,
   isLoggedIn,
   loggedInUserData,
+}: {
+  children: React.ReactNode;
+  serverSidePropsCorrectlyCalled: boolean;
+  isLoggedIn: boolean;
+  loggedInUserData: { firstName: string; email: string }
```
sheenarbw: This gets copy-pasted all over the place now:

```
{ firstName: string; email: string }
```

What if we decide to add a lastName or a cellphoneNumber? Then we'll have to change the code in a million places. 

Make an interface in one file, and then import it as needed.

Remember DRY

frontend-mini-challenge/components/LoggedInPage.tsx
```diff
@@ -55,7 +61,15 @@ export default function Page({
...
 }
 
-export function Presentation({ handleLogout, loggedInUserData, children }) {
+export function Presentation({
+  handleLogout,
+  loggedInUserData,
+  children,
+}: {
+  handleLogout: () => void;
+  loggedInUserData: { firstName: string; email: string }
```
sheenarbw: not DRY

frontend-mini-challenge/components/LoggedInPage.tsx
```diff
@@ -131,13 +145,17 @@ Then when you make use of this page component:
     </Page>
 
 */
-export async function getServerSidePropsForLoggedInPage({ query, req }) {
-  const whoAmIResponse = await serverSideWhoAmI({ query, req });
+
+
+export const getServerSidePropsForLoggedInPage = async (context: GetServerSidePropsContext) => {
```
sheenarbw: explicit is better than implicit. 

Are we really expecting the props to look like [this](https://github.com/vercel/next.js/blob/v10.2.0/packages/next/types/index.d.ts#L136-L151)?

It looks like we don't use most of that.

Good code makes its intentions clear. So it's important to be specific about what is required.

We are using Typescript so that when other coders work on this project they don't mess up and pass the wrong thing into a function. So we should not accept the wrong thing. We should only accept exactly the correct thing.

sheenarbw (changes requested):
Still some rough edges...

frontend-mini-challenge/components/LoggedInPage.tsx
```diff
@@ -15,12 +15,26 @@ import { useEffect } from "react";
 import { useRouter } from "next/router";
 import { clearAuthToken } from "../lib/authTokenStorage";
 import { useCookies } from "react-cookie";
+import { GetServerSidePropsContext } from 'next';
+
+
+
+
+const LoggedInUserDataType = {
```
sheenarbw: Should this not be an `interface`?

frontend-mini-challenge/components/LoggedInPage.tsx
```diff
@@ -131,13 +153,15 @@ Then when you make use of this page component:
     </Page>
 
 */
-export async function getServerSidePropsForLoggedInPage({ query, req }) {
-  const whoAmIResponse = await serverSideWhoAmI({ query, req });
+
+
+export const getServerSidePropsForLoggedInPage = async ({ query, req }: GetServerSidePropsContext) => {
```
sheenarbw: I have mentioned this one multiple times now. If you do not understand a comment then please do not ignore the comment.  I really don't like repeating myself this many times.

Look at how much information is inside GetServerSideProps: 
https://github.com/vercel/next.js/blob/v10.2.0/packages/next/types/index.d.ts#L136-L151

if 90% of the interface is completely inaccurate then it's a bad choice. 

Please use something appropriate here.  GetServerSidePropsContext is inappropriate.

If you disagree then say so. Do not simply ignore this feedback.

sheenarbw (changes requested):
Some repeated comments. Please address everything.

frontend-mini-challenge/components/LoggedInPage.tsx
```diff
@@ -16,12 +16,25 @@ import { useRouter } from "next/router";
 import { clearAuthToken } from "../lib/authTokenStorage";
 import { useCookies } from "react-cookie";
 
+interface Props {
```
sheenarbw: The different components have different props. Please use different interfaces for the different things.

If the foo function has the arguments a, b and c and the bar function has d, e and f then it is incorrect to say that foo requires a,b,c,d,e, f, req and query. 

Please make use of the correct interface. It should tell the truth of what is required in each function.. Each function requires different things. We need to be exact and specific. If you add in extra things that aren't needed at all then that makes it worse.

sheenarbw: please also make a point of importing things that are defined elsewhere in the code.

frontend-mini-challenge/components/LoggedInPage.tsx
```diff
@@ -16,12 +16,25 @@ import { useRouter } from "next/router";
...
+  children: React.ReactNode;
+  serverSidePropsCorrectlyCalled?: boolean;
+  isLoggedIn?: boolean;
+  loggedInUserData: {
+    firstName: string;
+    email: string;
+  };
+  handleLogout?: () => void;
+  req?: string;
```
sheenarbw: req isn't used by any function here. Why is it in this interface at all?

frontend-mini-challenge/components/LoggedInPage.tsx
sheenarbw: and req is definitely not a string.

sheenarbw (changes requested):
I think you can do better

frontend-mini-challenge/components/Alerts.tsx
```diff
@@ -3,7 +3,13 @@ import { ErrorIcon, InfoIcon } from "../brand";
 
 const ICON_SIZE = "2rem";
 
-export function ErrorAlert({ children, title }) {
+interface Props {
+  title: string;
+  children: string;
```
sheenarbw: it would be better here to say that the children could be a string or another component. Eg if we wanted to have multiple paragraphs of text then a string wouldn't cut it. We would either use multiple Text components or a Stack with a bunch of Text components inside it.

frontend-mini-challenge/components/LoggedInPage.tsx
```diff
@@ -131,13 +152,14 @@ Then when you make use of this page component:
     </Page>
 
 */
-export async function getServerSidePropsForLoggedInPage({ query, req }) {
-  const whoAmIResponse = await serverSideWhoAmI({ query, req });
+
+export async function getServerSidePropsForLoggedInPage({ query, req }: { query: { [key: string]: string }, req: IncomingMessage }) {
```
sheenarbw: There is no reason to invent things here. Look:

```
type GetServerSidePropsContext<
  Q extends ParsedUrlQuery = ParsedUrlQuery
> = {
  req: IncomingMessage & { 
    cookies: NextApiRequestCookies
  }
  res: ServerResponse
  params?: Q
  query: ParsedUrlQuery
  preview?: boolean
  previewData?: PreviewData
  resolvedUrl: string
  locale?: string
  locales?: string[]
  defaultLocale?: string
}
```
This information comes from https://github.com/vercel/next.js/blob/v10.2.0/packages/next/types/index.d.ts#L136-L151 



What datatype is query? What datatype is req?  Please use those. We know where these props come from, we know exactly what the types are.

sheenarbw (changes requested):
There is still one weird guess in here that I can see.

I need to learn about the different React children prototypes still so there is a chance I've missed something. Please triple-check that there is nothing else that you guessed on.  I'm very keen to get this finished.

frontend-mini-challenge/components/LoggedInPage.tsx
```diff
@@ -131,8 +155,15 @@ Then when you make use of this page component:
...
 */
-export async function getServerSidePropsForLoggedInPage({ query, req }) {
-  const whoAmIResponse = await serverSideWhoAmI({ query, req });
+
+export async function getServerSidePropsForLoggedInPage({
+  query,
+  req,
+}: {
+  query: ParsedUrlQuery;
+  req: IncomingMessage;
```
sheenarbw: Please use exactly this for req's type. There is no need to be creative here, we can just copy. 

Is there a reason you want to leave out the part about the cookies?

```
req: IncomingMessage & {
    cookies: NextApiRequestCookies
  }
```

sheenarbw (changes requested):
There is still one thing that looks inaccurate as far as I can see.

Also, this branch is not up to date with the latest code. 

Please do the following:
- fix the last inaccuracy
- merge the latest develop branch into your branch and fix any merge conflicts that come up
- make sure that the storybook tests still pass
- make sure you can build the application. You can see how to run it by looking at package.json

PR #651, opened 2023-07-07
Django admin: bulk deactivate users

Ngoakor12:

## Description:

Can bulk deactivate users from:
- user list page(`/admin/core/user/`)
- team details page(`/admin/core/team/272/change/`)

## Screenshots/videos

**Users list page** via actions with filter team filters(left side) and user search(top)
image

**Team details page**
image

## I solemnly swear that:

### DIFF
--- a/backend/curriculum_tracking/admin.py
+++ b/backend/curriculum_tracking/admin.py
@@ -163,6 +163,7 @@ class UserAdmin(BaseUserAdmin):
         "groups",
         # "user_permissions",
     )
+    list_per_page = 25
 
     actions = ["bulk_deactivate_users"]
 
--- a/backend/curriculum_tracking/admin.py
+++ b/backend/curriculum_tracking/admin.py
@@ -142,7 +142,6 @@ class UserAdmin(BaseUserAdmin):
     )
     list_display = ("email", "is_superuser", "active")
     list_filter = (
-        "groups",
         "is_superuser",
         "is_staff",
         "active",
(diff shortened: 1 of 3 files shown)


backend/curriculum_tracking/admin.py
```diff
@@ -141,8 +140,9 @@ class UserAdmin(BaseUserAdmin):
         # GroupInline,
         # RecruitProjectInline,
     )
-    list_display = ("email", "is_superuser")
+    list_display = ("email", "is_superuser", "active")
     list_filter = (
+        "groups",
```
sheenarbw: There are currently 294 groups in the Tilde prod database. Will this put all 294 groups here? 

image

That would be a lot of data - it'll be slow, most of the time it wont be needed, it'll also make the other filters less accessable.  

This filter should be removed.

If there is a better way to put a filter in here that would be good. Eg if there was a textbox or something then that would be usable. But I'm not sure how hard that is to pull off.

sheenarbw (changes requested):
It looks good except for one problem

Ngoakor12:
I decided to remove the team from the filters.

I also added a limit to the number of users on initial load.

backend/curriculum_tracking/admin.py
```diff
@@ -164,6 +163,14 @@ class UserAdmin(BaseUserAdmin):
         "groups",
         # "user_permissions",
     )
+    list_per_page = 25
+
+    actions = ["bulk_deactivate_users"]
+
+    def bulk_deactivate_users(modeladmin, request, users):
```
sheenarbw: This would be better:

```
def bulk_deactivate_users(modeladmin, request, users):
    users.update(active=False)
```

sheenarbw (commented):
One little thing out of place.

sheenarbw (approved):

PR #652, opened 2023-07-11
Django admin: bulk activate users

Ngoakor12:

## Description:

## I solemnly swear that:

### DIFF
--- a/backend/curriculum_tracking/admin.py
+++ b/backend/curriculum_tracking/admin.py
@@ -3,7 +3,6 @@ from . import models
 from core import models as core_models
 from adminsortable2.admin import SortableInlineAdminMixin
 from automarker import models as automarker_models
-from .helpers import activate_users
 
 
 class ContentItemAutoMarkerConfigAdmin(admin.TabularInline):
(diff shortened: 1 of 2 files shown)


backend/curriculum_tracking/helpers.py
```diff
@@ -182,3 +182,7 @@ def agile_card_reviews_outstanding(user):
 def pull_request_reviews_outstanding(user):
     # TODO: only implement this once the github webhook is healthier
     return []
+
+
+def activate_users(users):
```
sheenarbw: This is a funny place for this function.  And since the function is a one-liner, should it even exist?

backend/curriculum_tracking/admin.py
```diff
@@ -165,6 +164,13 @@ class UserAdmin(BaseUserAdmin):
         # "user_permissions",
     )
 
+    actions = ["bulk_activate_users"]
+
+    def bulk_activate_users(modeladmin, request, users):
```
sheenarbw: I think this would be better:

```
def bulk_activate_users(modeladmin, request, users):
        users.update(active=True)
```

sheenarbw (changes requested):
Looks good, there is just one thing that looks out of place. It's a small thing though.

It's very rare that it is useful to make a separate function for a one-liner. I'd only do that if the one-liner is a complicated line and either needs to be tested or reused.

sheenarbw (approved):

PR #653, opened 2023-07-21
Bulk add learners to team in admin panel

Ngoakor12:

## Description:

## Screenshots/videos
:movie_camera:  [**Demo video**](https://www.loom.com/share/a27e013c1ed04477a28945cb8a674c2e?sid=27e459ae-9b3b-4403-a806-ae4aae5749d9)

## I solemnly swear that:


kingraphaii:
Do you have an idea about how and where these users are fetched from in bulk? Like the format they come in? Tryna see if a text input or file upload would be the best option.

Ngoakor12:
> Do you have an idea about how and where these users are fetched from in bulk? Like the format they come in? Tryna see if a text input or file upload would be the best option.

I'm not too sure, my guess was that it's all manual, they get a list of emails for via email/text then add them to a new or existing team.

Ryan (@ry-oc) can clarify/decide on the file vs text input issue. Who know maybe both can be used and it would ultimately be down to preference :man_shrugging:

kingraphaii:
Yeah, had dropped him a text on RC asking on that as well. I think it's important we consider it before shipping.

Ngoakor12:
> Yeah, had dropped him a text on RC asking on that as well. I think it's important we consider it before shipping.

Cool you'll let me know once you get a reply, otherwise we'll have to raise it in a standup.

ry-oc:
Hey Ngoako,

Apologies for the delay. Looking at the screenshots there are areas I am unsure of and i don't want to make assumptions.

Could you create a video (possibly [loom](https://www.loom.com/)) to walk me through the everything you need to do to make use of this feature from start to finish?

Ngoakor12:
> Hey Ngoako,
> 
> Apologies for the delay. Looking at the screenshots there are areas I am unsure of and i don't want to make assumptions.
> 
> Could you create a video (possibly [loom](https://www.loom.com/)) to walk me through the everything you need to do to make use of this feature from start to finish?

Sure, here's the [demo video](https://www.loom.com/share/517525dab9ab424c879c55593e1f0973?sid=4dff62df-8a94-412d-afa1-52d375829967)

ry-oc (approved):
Works for me

ry-oc:
@Ngoakor12 quick question - if an email address is incorrect would it have an error or something? Or would it simply not work?

Ngoakor12:
> @Ngoakor12 quick question - if an email address is incorrect would it have an error or something? Or would it simply not work?

It wouldn't work for that email.

It will only work for emails that currently belong to a user in the database.

Ngoakor12:
I'll see if there's a way I can add user feedback for when an operation was successful or not

Ngoakor12:
> I'll see if there's a way I can add user feedback for when an operation was successful or not

Users added feedback:

image

No users added feedback:

image

Ngoakor12:
@ry-oc [video demo for the updates](https://www.loom.com/share/a27e013c1ed04477a28945cb8a674c2e?sid=27e459ae-9b3b-4403-a806-ae4aae5749d9)

ry-oc (approved):
Perfect from my side :)

backend/core/views.py
```diff
@@ -293,3 +300,32 @@ def competence_reviews_outstanding(self, request, pk=None):
 #             result[team.id]["permissions"].append(permission)
 
 #     return Response(result)
+
+
+def bulk_add_users(request, team_id):
```
kingraphaii: We will need permission & auth checks here. Currently it looks exposed to all.

backend/core/urls.py
```diff
@@ -8,4 +8,9 @@
     path("test_logs/", views.test_logs),
     path("test_long_running_request/", views.test_long_running_request),
     path("test_kill_dramatic_worker/", views.test_kill_dramatic_worker),
+    path(
+        "admin/team/<int:team_id>/bulk_add_users/",
```
kingraphaii: More of an opinion here, urls in this file aren't namespaced. Since this url is more of a utility than part of a CRUD url group we may need to define it in this fashion. We can keep the `admin` part as an indicator that this is an admin util. Something like `path('admin/add_bulk_users_to_team/<int:team_id>', ...)`.

Let me know what you think

kingraphaii (changes requested):
Check the comment about auth, otherwise it's looking good overall!

backend/core/urls.py
Ngoakor12: The logic makes sense to me.

kingraphaii (approved):

backend/backend/urls.py
```diff
@@ -206,6 +206,7 @@
     path("api-auth/", include("rest_framework.urls", namespace="rest_framework")),
     path("api/", include(router.urls)),
     path("api/", include("core.urls")),
+    path("", include("core.urls")),
```
sheenarbw: now the core urls are in here twice. It'll work but it's a bit of a messy thing to do.

How about splitting the core urls up according to purpose? For example

```
    path("api/", include("core.api_urls")),
    path("admin/", include("core.admin_urls")),
```

backend/core/forms.py
```diff
@@ -22,3 +21,10 @@ def clean(self):
             raise ValidationError("Choose one, and only one, filter to apply")
 
         return data
+
+
+class BulkAddUsersForm(forms.Form):
+    email_addresses = forms.CharField(
+        label="Email addresses",
+        widget=forms.Textarea(attrs={"rows": 15, "style": "display: flex; width:100%"}),
```
sheenarbw, line 30: Can you add some extra info about the format of the input. How must the user know how to seperate the email addresses? 

It would also be good to tap into Django's own form validation stuff.

What should the error message be if an email address isn't found?

sheenarbw: Django's form api has a lo of functionality. You are using it just to draw a form but it should be used all the way. 

Ideally, we should be able to say `form.cleaned_data["email_addresses"]` to get the email addresses of the users. Or, better yet, `form.cleaned_data["users"]`

backend/core/views.py
```diff
@@ -293,3 +301,33 @@ def competence_reviews_outstanding(self, request, pk=None):
...
+
+
+@staff_member_required
+def bulk_add_users_to_team(request, team_id):
+    team = get_object_or_404(Team, pk=team_id)
+
+    if request.method == "POST":
+        form = BulkAddUsersForm(request.POST)
+        if form.is_valid():
+            email_addresses = request.POST["email_addresses"]
```
sheenarbw: A lot of this should be inside the form class.

backend/curriculum_tracking/helpers.py
```diff
@@ -182,3 +183,32 @@ def agile_card_reviews_outstanding(user):
...
+    return []
+
+
+def remove_leading_and_trailing_whitespace(string):
+    if string and type(string) == str:
+        return string.strip()
+    return ""
+
+
+def get_email_addresses_from_str(email_addresses_str):
```
sheenarbw: Should this not be part of the form? Why is it all the way over here? It's not cohesive here. This function is only used on the form.

Also, the code can be a bit tighter. It's not necessary to separate tiny things out into new functions. It gets a bit spaghetti-like.

```
def get_email_addresses_from_str(email_addresses_str):
    email_addresses = re.split("[ ,\n]", email_addresses_str)
    email_addresses = [email.strip() for email in email_addresses]
    return [email for email in email_addresses if email]
```

sheenarbw (changes requested):
On the right track. It would be good to learn more about how Django forms work, those things have superpowers that you are working around.

backend/core/forms.py
Ngoakor12: I couldn't figure out how to get it working with `form.cleaned_data["users"]`.

backend/backend/urls.py
```diff
@@ -205,7 +205,8 @@
     path("admin/", admin.site.urls),
     path("api-auth/", include("rest_framework.urls", namespace="rest_framework")),
     path("api/", include(router.urls)),
-    path("api/", include("core.urls")),
+    path("api/", include("core.api_urls")),
+    path("admin/", include("core.admin_urls")),
```
kingraphaii: We may need to differentiate the path here, to avoid conflicts with `django-admin` urls housed here: `path("admin/", admin.site.urls),`

kingraphaii: May be add onto it, something like: `path("admin/some-differentiator", include("core.admin_urls"))`, inspired by the zmc & dj-res-auth URLs on lines 210 & 211

backend/core/views.py
```diff
@@ -293,3 +298,31 @@ def competence_reviews_outstanding(self, request, pk=None):
...
+        form = BulkAddUsersForm(request.POST)
+        if form.is_valid():
+            email_addresses = form.cleaned_data["email_addresses"]
+            users_added_to_team = add_users_to_team(team, email_addresses)
+            if users_added_to_team:
+                messages.success(
+                    request,
+                    f"The following users were successfully added to the \"{team.name}\" team: {', '.join(users_added_to_team)}",
+                )
+                return redirect(f"/admin/core/team/{team_id}/change")
```
kingraphaii: Use `reverse` or `reverse_lazy` to the named route so that this won't break in case of any changes made to the models. Like you did in the templates:

`<a href="{% url 'admin:app_list' app_label='core' %}">Core</a>`

backend/core/helpers.py
```diff
@@ -24,3 +24,15 @@ def get_first_and_last_names_from_business_email(email):
     ).groups()
 
     return first_name.capitalize(), last_name.capitalize()
+
+
+def add_users_to_team(team, email_addresses):
```
kingraphaii: Also thinking we could probably add this to the form. And use it in the same fashion we call `form.save()` for model forms. I don't see the need for a separate "helper" function for this yet.

backend/core/views.py
```diff
@@ -293,3 +298,31 @@ def competence_reviews_outstanding(self, request, pk=None):
 #             result[team.id]["permissions"].append(permission)
 
 #     return Response(result)
+
+
+@staff_member_required
+def bulk_add_users_to_team(request, team_id):
```
kingraphaii: If you're up for it, you can refactor this to use a Class Based View instead, particularly `FormView` it could make our code cleaner and a lot less boilerplate-ish. Not a hard requirement though.

kingraphaii (changes requested):

backend/core/views.py
```diff
@@ -293,3 +300,52 @@ def competence_reviews_outstanding(self, request, pk=None):
...
+            messages.error(
+                self.request,
+                f'No users were added to the "{team.name}" team. Make sure the users and/or entered email addresses exist and try again.',
+            )
+            return redirect(
+                reverse("bulk_add_users_to_team", kwargs={"team_id": team.id})
+            )
+
+    def add_users_to_team(self, team, email_addresses):
+        if team:
```
sheenarbw: Under what conditions would the team be falsely? Surely only if something went wrong in a weird way? 
If something breaks then the best thing is to have a noisy error message because then we can fix whatever the problem is.

It would be good to remove the if statement here

backend/core/views.py
```diff
@@ -293,3 +300,52 @@ def competence_reviews_outstanding(self, request, pk=None):
...
+            users = User.objects.filter(email__in=email_addresses)
```
sheenarbw: What if there are 20 email addresses entered and one of them is invalid? Should the admin user need to read through the list of email addresses and play spot the difference? 

Proper form validation is important. There should be an error message that says which emaild aren't found.

backend/core/forms.py
```diff
@@ -22,3 +22,28 @@ def clean(self):
...
+    )
+
+    def clean_email_addresses(self):
+        email_addresses_str = self.cleaned_data["email_addresses"]
+        if not email_addresses_str:
+            raise ValidationError(
+                f"No email addresses entered.",
+            )
+        email_addresses_list = self.get_email_addresses_from_str(email_addresses_str)
+        return email_addresses_list
```
sheenarbw: I need to insist here: Form validation needs to happen all the way. 

If any of the email addresses are not associated with a User then we need to raise a ValidationError that says which email addresses are causing problems.  This means we need the form to do a database query.

We should not have to query the database again in the view. So the form instance should maintain a list of users. 

There are a few ways to do this:
- You can just use something like `self.users` (messy because it's non-standard, but it'll work)
- get `clean_email_addresses` to return the users.  (messy because names matter)
- maybe you can make a hidden field called users and then have "cleaned_data['users']" get populated from the other field. This seems like it will be the tidiest but I haven't tried this before.

sheenarbw (changes requested):
Form validation is still a bit of an issue here

backend/core/views.py
```diff
@@ -300,29 +298,33 @@ def competence_reviews_outstanding(self, request, pk=None):
...
+    form_class = BulkAddUsersToTeamForm
+    template_name = "admin/core/bulk_add_users_form.html"
+    success_url = reverse_lazy("bulk_add_users_to_team")
+
+    def get_login_url(self):
+        return reverse("admin:login")
+
+    def setup(self, request, *args, **kwargs):
+        super().setup(request, *args, **kwargs)
+        self.team = Team.objects.get(id=self.kwargs["team_id"])
```
kingraphaii: It'd be nicer if we handled a scenario where id is non existent in the db because the `.get()` method, unlike `filter()` will raise an exception if the team ID does not exist.

kingraphaii (commented):
Looks good, left a little note on something we can improve on.

sheenarbw (approved):
Looks good!

PR #657, opened 2023-08-07
Added bulk card regeneration action in Team admin

kingraphaii:
> Added confirmation template (WIP)

Related issues: N/A

## Description:

This new feature allows admins to regenerate cards for Teams.

In the Django admin, users can select the teams to do the regen. for, then choose 'Bulk regenerate cards for members' in the *Actions* dropdown.

Task runs in background, so if working on large datasets, changes may take time to reflect.

## Screenshots/videos
image

Users will need to confirm this action as it deletes any existing assignments to other cards:
image

## I solemnly swear that:

- [X] I have merged the develop branch into my branch and fixed any merge conflicts


backend/core/admin.py
```diff
@@ -58,9 +59,29 @@ def deactivate_team_members(self, request, queryset: object):
...
+    def deactivate_team_members(self, request, queryset: object):
+        for team in queryset:
+            for team_member in team.active_users:
+                team_member.active = False
+                team_member.save()
 
+    def bulk_regenerate_cards_for_members(self, request, queryset: object):
+        if request.POST.get('post', None):
+            for team in queryset:
+                bulk_regenerate_cards_for_team(team)
```
sheenarbw: This would be a very long running request. It should be handled by the dramatiq worker. Take a look at how the current delete and recreate cards api works. 

I would append one task for each user who needs to be regenerated.

backend/curriculum_tracking/card_generation_helpers.py
```diff
@@ -266,7 +266,14 @@ def update_cards_accorrding_to_progress(user):
 
 
 def generate_all_content_cards_for_team(team):
-    todo
+    # todo
```
sheenarbw: This function was meant to fail loudly instead of fail silently.
I don't think it's being used anywhere though.

If you don't like the `todo` you can explicitly raise a notimplemented exception. It amounts to the same thing. `todo` is just faster to type.

backend/core/admin.py
sheenarbw: So something like:

```
from long_running_request_actors import (
                delete_and_recreate_user_cards as actor,
            )

for user_id in active_memer_ids:
    actor.send_with_options(kwargs={"user_id": user.id})

```

sheenarbw (changes requested):
Largely on the right track :) You're missing one important detail.

I also don't work with the admin panel enough to visualize the confirmation form you added there without rinning the code, but I'm sure it's fine.

backend/core/admin.py
kingraphaii: What would you recommend, spawning `delete_and_recreate_user_cards` actor for each user in a team in a new actor, say `bulk_regenerate_cards_for_members`, or running that separate actor, doing all the regeneration for all users at once

kingraphaii: Something like:

```python
@actor(max_retries=3)
def bulk_regenerate_cards_for_team(team):
    for user in team.users.filter(active=True):
        delete_and_recreate_user_cards.send_with_options(kwargs={"user_id": user.id})
```

sheenarbw: Better to do it the way I suggested. It would be faster because regens can happen in parallel. Also if there are any issues with regenning one user's board then that wont totally derail the process for all subsequent users.

backend/curriculum_tracking/card_generation_helpers.py
```diff
@@ -266,7 +266,14 @@ def update_cards_accorrding_to_progress(user):
...
+    pass
+
+
+def bulk_regenerate_cards_for_team(team):
```
sheenarbw: Then this wont be needed any more

Ngoakor12 (changes requested):
- Maybe add a proper title for action confirmation page(before "|")
image


Otherwise I don't see anything bad

Ngoakor12 (approved):

backend/core/tests/test_model_admin_actions_exist.py
```diff
@@ -0,0 +1,12 @@
+from django.test import TestCase
+
+from core.admin import TeamAdmin
+
+
+class TestModelAdminActionsExist(TestCase):
+     def setUp(self) -> None:
```
sheenarbw, line 7: If you don't make a setUp method then the super class's setup will get called by default. Did you mean to do something here?  This code doesn't do anything.

sheenarbw (approved):

PR #662, opened 2023-09-15
Python manage.py check_all_configuration

Ngoakor12:

## Description:

## I solemnly swear that:


automarker_2/automarker_app/management/commands/check_all_configuration.py
```diff
@@ -0,0 +1,12 @@
+from django.core.management.base import BaseCommand
+from automarker_app.lib.marker import get_all_marker_configs
+from automarker_app.lib.constants import CONFIG_STATUS_DEBUG, CONFIG_STATUS_PRODUCTION
+
+
+class Command(BaseCommand):
```
kingraphaii: Isn't this supposed to like actually run the configs like what "check_project_configuration" does?

[Reference](https://github.com/Umuzi-org/Tilde/blob/8b2540b663f25e3cfbe7f0c84dcb9c5bd655209f/automarker_2/automarker_app/management/commands/check_project_configuration.py#L11)

Ngoakor12: I agree when looking at the script names but the task description made me think that this should look more like the print_configuration script :shrug:

sheenarbw: It needs to run the self-check for each config.

kingraphaii:
`__file__` wasn't being updated on some iterations, so it would try to use an adapter from a different project's config on another which was causing the failure. Had to make it run `mark_project` function in a separate process as it wasn't throwing those errors if run individually.

automarker_2/automarker_app/management/commands/check_all_configuration.py
```diff
@@ -0,0 +1,96 @@
...
+    def add_arguments(self, parser: CommandParser) -> None:
+        parser.add_argument("--skip", type=int, nargs="+")
+
+    def handle(self, *args, **options):
+        good_configs = []
+        bad_configs = []
+
+        skip_ids = options.get("skip", [])
```
kingraphaii: You can cleanup the cmd, had added some arg to skip certain IDs for faster testing. Not sure if it will still be needed?

kingraphaii (commented):
The output needs some work, need to make it simple like the default output we get from marking projects, but yet clear.

automarker_2/automarker_app/management/commands/check_all_configurations.py
```diff
@@ -0,0 +1,79 @@
...
+            skip_ids = options.get("skip", [])
+
+        for config in get_all_marker_configs():
+            if (
+                config.status
+                in [
+                    CONFIG_STATUS_DEBUG,
+                    CONFIG_STATUS_PRODUCTION,
+                ]
+                and config.content_item_id not in skip_ids
```
kingraphaii, line 39: This line raises an error when we don't pass any ids to skip

kingraphaii (changes requested):

kingraphaii:
We also need to catch a timeout error, just realizes futures.ProcessPoolExecutor() will raise `subprocess.TimeoutExpired` if it passes the default, so maybe put 2 minutes? then just notify via terminal that the config check timed out.

```
# You can use this to increase it to 2 mins.
steps = _future.result(timeout=120)
```

automarker_2/automarker_app/management/commands/check_all_configurations.py
```diff
@@ -0,0 +1,84 @@
...
+                        config.content_item_id,
+                        config.flavours[0],
+                        self_test=True,
+                        fail_fast=True,
+                    )
+
+                steps = []
+                try:
+                    steps = _future.result(timeout=120)
+                except TimeoutError:
```
kingraphaii: This is the exception that gets raised:

`subprocess.TimeoutExpired`

kingraphaii (changes requested):

kingraphaii (approved):

sheenarbw (approved):
Nice work!

PR #667, opened 2023-10-24
Give new staff write access on all learner repos

kingraphaii:

## Description:
Functionality should be in the Admin panel

We need to give a user access to all repos they have view access of the projects.

Second potential functionality give a staff member access to all repos that belong to:

- a specific user
- all users in a team

## Screenshots/videos

image

image

This makes use of the already existing `invite_github_collaborators_to_repo` method inside the `RecruitProject` model. Here is a snippet of it:

```python
def invite_github_collaborators_to_repo(self):
  from social_auth.github_api import Api
  from social_auth.models import SocialProfile
  from git_real.helpers import add_collaborator, list_collaborators

  api = Api(GIT_REAL_BOT_USERNAME)
  repo = self.repository
  existing_collaborators = list_collaborators(api, repo.full_name)

  collaborator_users = {
      "reviewer users": list(self.reviewer_users.filter(active=True)),
      "assigned users": list(self.recruit_users.filter(active=True)),
      "users with explicit permission": self.get_users_with_permission(
          Team.PERMISSION_REPO_COLLABORATER_AUTO_ADD
      ),
  }
```

specifically, the `users with explicit permission` part. `Team.PERMISSION_REPO_COLLABORATER_AUTO_ADD` is a group of permissions as defined here:

```python
class Team(AuthGroup, Mixins):
  ... 
  PERMISSION_REPO_COLLABORATER_AUTO_ADD = [
      # when a repo is created then folks with these permissions get added to the repo as collaborators
      PERMISSION_MANAGE_CARDS,
      PERMISSION_VIEW_ALL,
      PERMISSION_ASSIGN_REVIEWERS,
      PERMISSION_REVIEW_CARDS,
  ]
  ...
```

This new action basically "refreshes" github collaborators based on these permissions. So if new staff "john" and "sarah" have the same perms to a certain team, doing it for "john" would be enough to give both staff members access to their repos.

## I solemnly swear that:


backend/curriculum_tracking/admin.py
```diff
@@ -164,7 +168,7 @@ class UserAdmin(BaseUserAdmin):
         # "user_permissions",
     )
 
-    actions = ["bulk_activate_users", "bulk_deactivate_users"]
+    actions = ["bulk_activate_users", "bulk_deactivate_users", "invite_to_github_repos"]
```
sheenarbw: Hmm, UX needs to be thought out. I do think it's important not to just jump in when there are UX things that need to be considered. 

Ryan needs this now, Ryan wont be the only person to ever make use of this. 

Please rope me in before building frontend stuff.  Dont do more frontend stuff here until I've taken a look at the UX please

backend/curriculum_tracking/admin.py
```diff
@@ -176,6 +180,49 @@ def bulk_deactivate_users(self, request, users):
...
+                    "action_checkbox_name": ACTION_CHECKBOX_NAME,
+                },
+            )
+
+    def _get_permitted_teams_for_users(self, users) -> list:
+        teams = []
+        for user in users:
+            for team in get_objects_for_user(
+                user,
+                core_models.Team.PERMISSION_REPO_COLLABORATER_AUTO_ADD,
```
sheenarbw: This is strange. Why is there a whole new permission here?  It really does not seem necessary

backend/long_running_request_actors.py
```diff
@@ -84,6 +84,16 @@ def recruit_project_invite_github_collaborators_to_repo(project_id):
...
 
 
+@actor(max_retries=3)
+def invite_collaborators_for_team_projects(team_name):
+    from curriculum_tracking.models import RecruitProject
+
+    projects = RecruitProject.objects.filter(recruit_users__groups__name=team_name)
+
+    for project in projects:
+        recruit_project_invite_github_collaborators_to_repo.send(project.pk)
```
sheenarbw, line 100: This can get VERY heavy. There might be a lot of repos here. Is there a way to make the inclusion of repos on complete projects optional? The default behavior should be to not include repos on projects that have completed.

sheenarbw (changes requested):
On the right track but there are some problems here:

1. Please dont make new permissions unless we are actually permitting something new
2. The UX was not discussed. It's very important to not jump the gun with user-facing stuff

backend/curriculum_tracking/admin.py
kingraphaii: Not a new perm, [ref](https://github.com/Umuzi-org/Tilde/blob/10e376e5828766b19387270b2f86f8029ddbfa9c/backend/core/models.py#L253).

snippet:
```python
class Team(AuthGroup, Mixins):

    PERMISSION_MANAGE_CARDS = PERMISSION_MANAGE_CARDS  # start cards, set due date,
    PERMISSION_VIEW_ALL = PERMISSION_VIEW_ALL  # look at anything
    PERMISSION_ASSIGN_REVIEWERS = PERMISSION_ASSIGN_REVIEWERS
    PERMISSION_REVIEW_CARDS = PERMISSION_REVIEW_CARDS
    PERMISSION_TRUSTED_REVIEWER = PERMISSION_TRUSTED_REVIEWER

    PERMISSION_VIEW = [
        # anyone with these permissions get to navigate to a specific team on the frontend and see the status of all the cards.
        PERMISSION_MANAGE_CARDS,
        PERMISSION_VIEW_ALL,
        PERMISSION_ASSIGN_REVIEWERS,
        PERMISSION_REVIEW_CARDS,
        PERMISSION_TRUSTED_REVIEWER,
    ]

    PERMISSION_REPO_COLLABORATER_AUTO_ADD = [
        # when a repo is created then folks with these permissions get added to the repo as collaborators
        PERMISSION_MANAGE_CARDS,
        PERMISSION_VIEW_ALL,
        PERMISSION_ASSIGN_REVIEWERS,
        PERMISSION_REVIEW_CARDS,
    ]
```

backend/core/templates/admin/core/confirm_add_github_collaborator.html
```diff
@@ -0,0 +1,44 @@
...
+        <a href="{% url 'admin:core_user_change' user.id %}">{{ user }}</a> &rsaquo;
+        Add as GitHub Collaborator
+    </div>
+{% endblock %}
+
+<div class="content">
+    {% block content %}
+        <h1>Adding <a href="{% url 'admin:core_user_change' user.id %}">{{ user.get_short_name }}</a> as collaborator on permissioned projects</h1>
+        {% if teams %}
+            <p>{{ user.get_short_name}} will be added as a collaborator to the following teams's projects:</p>
```
Ngoakor12: I think the correct grammar here is is `teams'` instead of `teams's`

backend/core/forms.py
```diff
@@ -66,3 +67,22 @@ def get_email_addresses_from_str(self, email_addresses_str):
...
         email_addresses = [email.strip() for email in email_addresses]
         return [email for email in email_addresses if email]
+
+
+class AddGithubCollaboratorForm(forms.Form):
+    include_complete_projects = forms.BooleanField(
+        label="Would you like to include complete projects?",
+        required=False,
+        initial=False,
+    )
```
Ngoakor12: image

UX wise: I think this should not be a question and the checkmark should be before the text. so maybe:
```
[  ] include complete projects
```

I'm thinking more close to forms where you have to tick to accept terms and conditions or you want to be sent promotional emails.

I just don't think it should be a question since the only answer option is to say yes, maybe if it was a toggle or it had to radio inputs

Ngoakor12 (changes requested):
- I'm not too certain how the permission teams are updated and if they are enough to add the collaborator to the teams, I think a demo or steps by step guide would've helped.
- Can you also add a cancel button next to the `add as collaborator` button. I can't find the words to explain it but it's similar to the "include complete projects" comment below. You can leave this if you're not convinced

Ngoakor12 (approved):

sheenarbw (approved):

PR #670, opened 2023-11-15
windows configuration

minat-hub:
Related issues: The automarker wasnt running well on windows because the initial configuration was for unix(mac and linux) os

## Description:

I  updated the steps.py so that if the os that automarker is running on is Unix it uses the existing command and if the os is windows it uses command that is compatible for windows

## I solemnly swear that:

### DIFF
--- a/automarker_2/automarker_app/lib/steps.py
+++ b/automarker_2/automarker_app/lib/steps.py
@@ -20,12 +20,17 @@ from .constants import (
     STEP_STATUS_RED_FLAG,
 )
 import json
+import platform
 
 
 def get_all_file_paths(directory):
     for path, _, filenames in os.walk(directory):
         for filename in filenames:
             yield os.path.join(path, filename)
+
+
+def check_if_current_os_windows():
+    return platform.system() == "Windows"
 
 
 class Step:
@@ -626,7 +631,7 @@ class PythonCreateVirtualEnv(Step):
         if len(stderr):
             self.set_outcome(STEP_STATUS_ERROR, message=stderr)
         else:
-            if os.name == "nt":
+            if check_if_current_os_windows():
                 python_executable_path = 'Scripts/python'
             else:
                 python_executable_path =  'bin/pip'
@@ -642,7 +647,7 @@ class PythonDoRequirementsTxtInstall(Step):
     name = "pip install requirements.txt"
 
     def run(self, project_uri, clone_dir_path, self_test, config, fail_fast):
-        if os.name == "nt":
+        if check_if_current_os_windows():
             pip_executable_path =  'Scripts/pip'
         else:
             pip_executable_path = 'bin/pip'
@@ -681,7 +686,7 @@ class PythonRunPytests(Step):
     name = "run learner pytests"
 
     def run(self, project_uri, clone_dir_path, self_test, config, fail_fast):
-        if os.name == "nt":
+        if check_if_current_os_windows():
             python_executable_path = f"{clone_dir_path}/automarker_venv/Scripts/python"
         else:
             python_executable_path = "automarker_venv/bin/python"
--- a/automarker_2/automarker_app/lib/steps.py
+++ b/automarker_2/automarker_app/lib/steps.py
@@ -627,13 +627,11 @@ class PythonCreateVirtualEnv(Step):
             self.set_outcome(STEP_STATUS_ERROR, message=stderr)
         else:
             if os.name == "nt":
-                command = (
-                    f"{clone_dir_path/'automarker_venv'/'Scripts'/'python'} -m pip install --upgrade pip"
-                )
+                python_executable_path = clone_dir_path / 'automarker_venv' / 'Scripts' / 'python'
             else:
-                command = (
-                    f"{clone_dir_path/'automarker_venv'/'bin'/'pip'} install --upgrade pip"
-                )
+                python_executable_path = clone_dir_path / 'automarker_venv' / 'bin' / 'pip'
+
+            command = f"{python_executable_path} -m pip install --upgrade pip"
             stdout, stderr = subprocess_run(command)
             if stderr:
                 breakpoint()
@@ -645,9 +643,11 @@ class PythonDoRequirementsTxtInstall(Step):
 
     def run(self, project_uri, clone_dir_path, self_test, config, fail_fast):
         if os.name == "nt":
-            command = f"{clone_dir_path/'automarker_venv'/'Scripts'/'pip'} install -r {clone_dir_path/'requirements.txt'}"
+            python_executable_path = clone_dir_path / 'automarker_venv' / 'Scripts' / 'pip'
         else:
-            command = f"{clone_dir_path/'automarker_venv'/'bin'/'pip'} install -r {clone_dir_path/'requirements.txt'}"
+            python_executable_path = clone_dir_path / 'automarker_venv' / 'bin' / 'pip'
+
+        command = f"{python_executable_path} install -r {clone_dir_path/'requirements.txt'}"
         stdout, stderr = subprocess_run(command)
         if len(stderr):
             if stderr.startswith("ERROR: Could not open requirements file"):
@@ -682,13 +682,11 @@ class PythonRunPytests(Step):
 
     def run(self, project_uri, clone_dir_path, self_test, config, fail_fast):
         if os.name == "nt":
-            command = (
-                f"cd {clone_dir_path} && {clone_dir_path}/automarker_venv/Scripts/python -m pytest --tb=line"
-            )
+            python_executable_path = f"{clone_dir_path} && {clone_dir_path}/automarker_venv/Scripts/python"
         else:
-            command = (
-                f"cd {clone_dir_path} && automarker_venv/bin/python -m pytest --tb=line"
-            )
+            python_executable_path = f"{clone_dir_path} && automarker_venv/bin/python"
+
+        command = f"cd {python_executable_path} -m pytest --tb=line"
         stdout, stderr = subprocess_run(command)
         if re.search("=== no tests ran in .* ===", stdout):
             self.set_outcome(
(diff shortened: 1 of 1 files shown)


kingraphaii (approved):

sheenarbw (changes requested):
Good code makes it's intensions clear. 

Code that works can still be bad code.

automarker_2/automarker_app/lib/steps.py
```diff
@@ -673,11 +681,13 @@ class PythonRunPytests(Step):
     name = "run learner pytests"
 
     def run(self, project_uri, clone_dir_path, self_test, config, fail_fast):
-        command = (
-            f"cd {clone_dir_path} && automarker_venv/bin/python -m pytest --tb=line"
-        )
-        stdout, stderr = subprocess_run(command)
+        if os.name == "nt":
+            python_executable_path = f"{clone_dir_path} && {clone_dir_path}/automarker_venv/Scripts/python"
```
kingraphaii: Is the && part if the python_executable_path?

automarker_2/automarker_app/lib/steps.py
```diff
@@ -639,7 +642,12 @@ class PythonDoRequirementsTxtInstall(Step):
     name = "pip install requirements.txt"
 
     def run(self, project_uri, clone_dir_path, self_test, config, fail_fast):
-        command = f"{clone_dir_path/'automarker_venv'/'bin'/'pip'} install -r {clone_dir_path/'requirements.txt'}"
+        if os.name == "nt":
+            python_executable_path = clone_dir_path / 'automarker_venv' / 'Scripts' / 'pip'
```
kingraphaii: Is this a python executable path or pip?

kingraphaii: if we want to stick to using python executable then we have to adapt it to "executable -m pip"

kingraphaii (changes requested):
There were some issues I noticed while I was trying to help Mpho set up his environment where some commands like "cp" and "mv" or "rm" weren't working. Do we have that covered here or there was something I did wrong

minat-hub:
> There were some issues I noticed while I was trying to help Mpho set up his environment where some commands like "cp" and "mv" or "rm" weren't working. Do we have that covered here or there was something I did wrong

Hello Percival, thank you for pointing that out. These issues were resolved by either using git bash terminal or ubuntu windows subsystem for linux terminal instead of powershell . It was because the automaker was being run on windows powershell. I was thinking along the lines of adding in the Readme that windows users should use the git bash or windows subsystem to run the automarker. Should i do this or include the equivalent commands for windows instead?

kingraphaii:
> Should i do this or include the equivalent commands for windows instead?

If we can make it work with command prompt or powershell, I think it'll be a proper windows implementation, we can resort to git bash/cygwin if the cmd/powershell approach does not work

minat-hub:
> If we can make it work with command prompt or powershell, I think it'll be a proper windows implementation, we can resort to git bash/cygwin if the cmd/powershell approach does not work

Hello Percival, I updated all the commands for automarker for powershell terminal but some of them doesnt work for command prompt so that means some of these commands had to be updated for it too. This wasn't a very efficient approach as it made the code bulky. Some of the linux command when converted to PowerShell also behaved slightly differently. I opened a pr to inform windows users to use git bash/cygwin. What do you think?

kingraphaii:
> What do you think?

I think that is OK if it works! Everyone who needs to use it should have git anyway so it's highly likely they have git-bash

kingraphaii (approved):

automarker_2/automarker_app/lib/steps.py
```diff
@@ -639,7 +642,12 @@ class PythonDoRequirementsTxtInstall(Step):
     name = "pip install requirements.txt"
 
     def run(self, project_uri, clone_dir_path, self_test, config, fail_fast):
-        command = f"{clone_dir_path/'automarker_venv'/'bin'/'pip'} install -r {clone_dir_path/'requirements.txt'}"
+        if os.name == "nt":
+            pip_executable_path =  'Scripts/pip'
```
sheenarbw: This doesn't seem DRY or clear. 

Can't we write this down once rather than over and over?  We check this 3 times `if os.name == "nt":`.
And it's really not clear what nt means, if I want to understand this code then I need to google it.

sheenarbw (changes requested):
This code can be more DRY and more clear.

automarker_2/automarker_app/lib/steps.py
```diff
@@ -626,9 +631,12 @@ def run(self, project_uri, clone_dir_path, self_test, config, fail_fast):
         if len(stderr):
             self.set_outcome(STEP_STATUS_ERROR, message=stderr)
         else:
-            command = (
-                f"{clone_dir_path/'automarker_venv'/'bin'/'pip'} install --upgrade pip"
-            )
+            if check_if_current_os_windows():
+                python_executable_path = 'Scripts/python'
+            else:
+                python_executable_path =  'bin/pip'
```
kingraphaii: Did you mean to put `pip` here? Also the naming can be better, or as a whole, why not have one function handle this, like:

```python
def get_python_executable_path():
    return 'Scripts/python' if is_current_os_windows() else 'bin/python'
```

then use that where we need to

automarker_2/automarker_app/lib/steps.py
```diff
@@ -639,7 +647,12 @@ class PythonDoRequirementsTxtInstall(Step):
     name = "pip install requirements.txt"
 
     def run(self, project_uri, clone_dir_path, self_test, config, fail_fast):
-        command = f"{clone_dir_path/'automarker_venv'/'bin'/'pip'} install -r {clone_dir_path/'requirements.txt'}"
+        if check_if_current_os_windows():
```
kingraphaii: Still not dry, check first comment

kingraphaii (changes requested):

automarker_2/automarker_app/lib/steps.py
```diff
@@ -626,9 +633,8 @@ def run(self, project_uri, clone_dir_path, self_test, config, fail_fast):
         if len(stderr):
             self.set_outcome(STEP_STATUS_ERROR, message=stderr)
         else:
-            command = (
-                f"{clone_dir_path/'automarker_venv'/'bin'/'pip'} install --upgrade pip"
-            )
+            python_executable_path_virtual_env = f'{get_python_executable_path()}/python' if is_current_os_windows() else f'{get_python_executable_path()}/pip'
```
kingraphaii: we can cache the value of `get_python_executable_path()` so that we can reuse it in the else block to keep our code tidy. Plus we're repeating calling "is_current_os_windows"

```python
# atop:
def get_python_executable_path():
    if platform.system() == "Windows":
        return 'Scripts/python.exe'
    return  'bin/python'

# inside PythonCreateVirtualEnv (here)
command = f"{clone_dir_path/ 'automarker_venv'/ get_python_executable_path()} -m pip install --upgrade pip"
```

We can use the same approach when we want to pip install, no reason why we should differentiate how we're calling it in `def run(**)`

```python
# like this
command = f"{clone_dir_path/ 'automarker_venv'/ get_python_executable_path()} -m pip install -r {clone_dir_path/'requirements.txt'}"
```

kingraphaii (changes requested):

automarker_2/automarker_app/lib/steps.py
```diff
@@ -673,11 +685,16 @@ class PythonRunPytests(Step):
...
     def run(self, project_uri, clone_dir_path, self_test, config, fail_fast):
+        python_executable_path_tests = (
+            f"{clone_dir_path}/{PYTHON_EXECUTABLE_PATH}"
+            if is_current_os_windows()
+            else PYTHON_EXECUTABLE_PATH
+        )
+
         command = (
-            f"cd {clone_dir_path} && automarker_venv/bin/python -m pytest --tb=line"
+            f"cd {clone_dir_path} && {python_executable_path_tests} -m pytest --tb=line"
```
kingraphaii: Something is off here, original logic was like this:

```python
command = f"cd {clone_dir_path} && automarker_venv/bin/python -m pytest --tb=line"
```

Meaning we'd first `cd` into the cloned repo, then run pytest using the python executable. So to do the same we can just replace the "bin/python" part as this the executable does not sit directly in the `clone_dir_path` as we've assumed here.

```python
command = f"cd {clone_dir_path} && automarker_venv/{PYTHON_EXECUTABLE_PATH} -m pytest --tb=line"
```

automarker_2/automarker_app/lib/steps.py
```diff
@@ -28,6 +29,19 @@ def get_all_file_paths(directory):
             yield os.path.join(path, filename)
 
 
+def is_current_os_windows():
+    return platform.system() == "Windows"
+
```
kingraphaii: Let's get rid of this two liner, we can do this check inside: `get_python_executable_path` like this

```python
def get_python_executable_path():
    if platform.system() == "Windows"
        return "Scripts/python"
    return "bin/python"
```

kingraphaii (changes requested):

automarker_2/automarker_app/lib/steps.py
kingraphaii: I just realized you are right, `&&` won't work on Windows. So, I think to clean up our code, we can cater to it this way:

```python
command = f"{clone_dir_path/'automarker_venv'/PYTHON_EXECUTABLE_PATH} -m pytest {clone_dir_path/tests --tb=line"
```

I'm just trying not to repeat some stuff, so I'm thinking we can curb that by passing the tests directory to the same exec. path we use in other places. Let me know what you think.

Unfortunately I've got no Windows box I can use to test this right away

kingraphaii (approved):

sheenarbw (approved):

PR #672, opened 2023-11-18
Add assertions for running tag, no repeated tags, and import learner code

kingraphaii:
Ref: [https://umuzi-projects.monday.com/boards/1254743632/pulses/1261667339](https://umuzi-projects.monday.com/boards/1254743632/pulses/1261667339)

## Description:

## I solemnly swear that:

### DIFF
--- a/automarker_2/automarker_app/lib/automarker_test_runner.py
+++ b/automarker_2/automarker_app/lib/automarker_test_runner.py
@@ -11,7 +11,6 @@ from .utils import (
     TAG_IMPORT_LEARNER_CODE,
     TAG_COMMAND_DESCRIPTION,
     AdapterCommandOutput,
-    test_runner_expects_code_imports,
 )
 from .constants import (
     STEP_STATUS_PASS,
@@ -120,11 +119,12 @@ class _TestRunner:
             )
 
         self.assert_setup_empty()
-        self.assert_command_description_present()
-        self.assert_running_tag_present()
         self.assert_no_repeated_tags()
-        self.assert_import_learner_code_present()
+        self.assert_no_unfinished_tags()
+        self.assert_required_tags_present()
+
         self.assert_no_import_errors()
+
         if assert_no_errors:
             self.assert_no_errors()
         if assert_no_import_side_effects:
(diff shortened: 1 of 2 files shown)


Ngoakor12 (approved):

automarker_2/automarker_app/lib/automarker_test_runner.py
```diff
@@ -187,6 +191,17 @@ def assert_command_description_present(self):
             TAG_COMMAND_DESCRIPTION
         ], f"expected command description to be present. There is something wrong with the automarker project configuration\n\nstderr={self.last_command_output.stderr}\n\nstdout={self.last_command_output.stdout}"
 
+    def assert_running_tag_present(self):
+        assert bool(
+            re.search(rf"<{TAG_RUNNING}>", self.last_command_output.stdout)
```
sheenarbw: Why does this work differently to assert_command_description_present? Why only check for opening tags and not closing tags?

automarker_2/automarker_app/lib/utils.py
```diff
@@ -115,3 +123,15 @@ def run_command(Cls, command):
                 )
 
         return Cls(stdout=stdout, stderr=stderr)
+
+
+def test_runner_expects_code_imports(cls):
+    class NewRunnerCls(cls):
```
sheenarbw: KISS

Surely there is a simpler way to do inheritance? 

If most things expect import tags then this should be in the base class, then just override the method where needed

sheenarbw (changes requested):
Some stuff in here is much more complicated than it needs to be. Always aim for the simplest code possible.

automarker_2/automarker_app/lib/automarker_test_runner.py
kingraphaii: ### Sample `__dict__` of `self.last_command_output`:
```python
{
    "stdout": '<setup>\n</setup>\n<command_description>\nconstruct a two Homes (home1, home2); adopt the animal (cat1) in both homes; then make all sounds in both homes\n</command_description>\n<import_learner_code>\n</import_learner_code>\n<running>\n</running>\n<returned>\n{"adopt_home_one":1,"adopt_home_two":1,"sounds_home_one":["Meow"],"sounds_home_two":["Meow"]}\n</returned>\n',
    "stderr": "",
    "setup": "",
    "import_learner_code": "",
    "running": "",
    "returned": {
        "adopt_home_one": 1,
        "adopt_home_two": 1,
        "sounds_home_one": ["Meow"],
        "sounds_home_two": ["Meow"],
    },
    "command_description": "construct a two Homes (home1, home2); adopt the animal (cat1) in both homes; then make all sounds in both homes",
}
```

If we do `assert self.last_command_output[TAG_RUNNING]` it will fail as per:

```python
>>> d = {"a": ""}
>>> assert d["a"]
  File "<stdin>", line 1, in <module>
AssertionError
```

but for command description it will work because we actually expect some output in there.

So some tags' blocks will not have any output, as shown in the sample even though they're properly setup. This is why I resorted to checking the `stdout` instead. I know even in doing so, there's room for improvement, but I just wanted to confirm If I'm getting this right.

sheenarbw: Ok cool. That makes sense.
Maybe just check that there is an opening and closing tag.

automarker_2/automarker_app/lib/automarker_test_runner.py
```diff
@@ -182,10 +186,47 @@ def register_test_error(self, error_message, status=STEP_STATUS_NOT_YET_COMPETEN
...
 
+    def assert_no_repeated_tags(self):
+        assert (
+            len(self.last_command_output.repeating_tags()) == 0
+        ), f"expected no repeating tags but the following tags appear more than once: {self.last_command_output.repeating_tags()}"
+
+    def assert_no_unfinished_tags(self):
+        if not self.last_command_output.stderr:
+            assert (
+                len(self.last_command_output.unfinished_tags()) == 0
```
Ngoakor12, line 197: I think the assertion message is a bit confusing. How about something like `expected tags to be closed but the following tags appear to be left open:...`

Ngoakor12 (changes requested):

Ngoakor12 (approved):

sheenarbw (approved):

PR #673, opened 2023-11-21
card component: Add information fields to card component

Ngoakor12:

## Description:

## Screenshots/videos

image

## I solemnly swear that:


backend/curriculum_tracking/models.py
```diff
@@ -1630,6 +1630,16 @@ def request_user_can_start(self):
...
+    
+    def request_user_is_assignee(self):
+        """should assignee email be shown on card?
+        This function is only used in template rendering, that is why we need to avoid argument parameters and we are using threadlocals
+        """
+        from threadlocal_middleware import get_current_user
+
+        user = get_current_user()
+        
+        return user.email==self.assignee_names[0]
```
sheenarbw: Why do you need to call the assignee_names function? Why not just use the orm? 

Something like:
`return self.assignees.first() == user`

backend/curriculum_tracking/models.py
```diff
@@ -1626,10 +1626,22 @@ def request_user_can_start(self):
...
+    def request_user_is_assignee(self):
+        """
+        Checks if current user is assignee
+        This function is only used in template rendering, that is why we need to avoid argument parameters and we are using threadlocals
+        """
+        from threadlocal_middleware import get_current_user
+
+        user = get_current_user()
+
+        return self.assignees.first() == user
```
kingraphaii, line 1668: Not sure if there will be projects with mltiple assignees, but the relationship is defined as many to many, so maybe just checking is user is in the assignees queryset will be sufficient here. Just a thought, let me know what you think

backend/frontend/templates/frontend/user/component_card.html
```diff
@@ -1,25 +1,144 @@
...
+            <p class="bold">
+                {{ card.content_type_nice }} 
+            </p>
+        </div>
+        {% if card.flavour_names %}
+            <div class="flex gap-1 items-center flex-wrap"> 
+                {% for flavour_name in card.flavour_names %}
+                    {%if flavour_name == "javascript"%}
+                        <span class="{{styles.flavour_chip}}" title="{{flavour_name}}">
+                            <svg  xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path fill="currentColor" d="M3 3h18v18H3V3m4.73 15.04c.4.85 1.19 1.55 2.54 1.55c1.5 0 2.53-.8 2.53-2.55v-5.78h-1.7V17c0 .86-.35 1.08-.9 1.08c-.58 0-.82-.4-1.09-.87l-1.38.83m5.98-.18c.5.98 1.51 1.73 3.09 1.73c1.6 0 2.8-.83 2.8-2.36c0-1.41-.81-2.04-2.25-2.66l-.42-.18c-.73-.31-1.04-.52-1.04-1.02c0-.41.31-.73.81-.73c.48 0 .8.21 1.09.73l1.31-.87c-.55-.96-1.33-1.33-2.4-1.33c-1.51 0-2.48.96-2.48 2.23c0 1.38.81 2.03 2.03 2.55l.42.18c.78.34 1.24.55 1.24 1.13c0 .48-.45.83-1.15.83c-.83 0-1.31-.43-1.67-1.03l-1.38.8Z"/></svg>
```
kingraphaii: Can we put svg illustrations for py,md,java as well?

backend/frontend/templates/frontend/user/component_card.html
```diff
@@ -1,25 +1,144 @@
...
-    {% if card.request_user_can_start %}
+    {% if card.due_time %}
+        <div class="flex gap-1" title="Due time">
+            <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path fill="currentColor" d="M12 20a7 7 0 0 1-7-7a7 7 0 0 1 7-7a7 7 0 0 1 7 7a7 7 0 0 1-7 7m0-16a9 9 0 0 0-9 9a9 9 0 0 0 9 9a9 9 0 0 0 9-9a9 9 0 0 0-9-9m.5 4H11v6l4.75 2.85l.75-1.23l-4-2.37V8M7.88 3.39L6.6 1.86L2 5.71l1.29 1.53l4.59-3.85M22 5.72l-4.6-3.86l-1.29 1.53l4.6 3.86L22 5.72Z"/></svg>
+            {% now "Y-m-d H:i:s" as current_datetime %}
+            {% if card.due_time|date:"U" > current_datetime|date:"U" %}
+                <p class="text-red-600">
+                    {{card.due_time|naturalday}} 
+                </p>
+                {% else %}
```
kingraphaii, line 130: Not seeing a difference here in this conditional check

backend/frontend/templates/frontend/user/component_card.html
```diff
@@ -1,25 +1,144 @@
...
+            {% for reviewer_name in card.reviewer_names %}
+                <p class="truncate" title="{{reviewer_name}}">{{reviewer_name}}</p> 
+            {% endfor %}
+        </div>
+    {% endif %}
 
+    {% if card.repo_url %}
+        <div class="flex gap-1">
+            <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path fill="currentColor" d="M12 2A10 10 0 0 0 2 12c0 4.42 2.87 8.17 6.84 9.5c.5.08.66-.23.66-.5v-1.69c-2.77.6-3.36-1.34-3.36-1.34c-.46-1.16-1.11-1.47-1.11-1.47c-.91-.62.07-.6.07-.6c1 .07 1.53 1.03 1.53 1.03c.87 1.52 2.34 1.07 2.91.83c.09-.65.35-1.09.63-1.34c-2.22-.25-4.55-1.11-4.55-4.92c0-1.11.38-2 1.03-2.71c-.1-.25-.45-1.29.1-2.64c0 0 .84-.27 2.75 1.02c.79-.22 1.65-.33 2.5-.33c.85 0 1.71.11 2.5.33c1.91-1.29 2.75-1.02 2.75-1.02c.55 1.35.2 2.39.1 2.64c.65.71 1.03 1.6 1.03 2.71c0 3.82-2.34 4.66-4.57 4.91c.36.31.69.92.69 1.85V21c0 .27.16.59.67.5C19.14 20.16 22 16.42 22 12A10 10 0 0 0 12 2Z"/></svg>
+            <a class="{{styles.link}}" href="{{card.recruit_project.repository.full_name|convert_git_url:card.recruit_project.repository.owner}}">Repository</a>
```
kingraphaii: cant we just use the owner/repo-name here

kingraphaii (changes requested):
image

Due date showing in red even though it's not overdue yet. Also can we put the time itself in the "title"? 

Can't see PR info on card as well

Can we also make the repository link's target _blank?

image

I have a feeling if we have more than 4 tags per card the ui wont look so good. Maybe we can place them elsewhere? Also just show the tag name in full. Having to hover over it to see the name is not ideal.

backend/curriculum_tracking/models.py
Ngoakor12: It looks like only one assignee is expected as far as I've seen currently. But this seems more ideal given relationships though

backend/frontend/templates/frontend/user/component_card.html
Ngoakor12: I'll double check

backend/frontend/templates/frontend/user/component_card.html
Ngoakor12: My thinking behind this was that the repo name didn't really provide any useful info and that it was too long and hard to present visually without cutting it. I may be wrong, I've never needed to know or see th repo name

Ngoakor12:
@kingraphaii 
> Due date showing in red even though it's not overdue yet. Also can we put the time itself in the "title"?

I'll test it out some more, the colours seemed to change as expected before
I will try and get the time to show too.

> Can't see PR info on card as well

Which PR info?

> I have a feeling if we have more than 4 tags per card the ui wont look so good. Maybe we can place them elsewhere? Also just show the tag name in full. Having to hover over it to see the name is not ideal.

I agree regarding the flavours and hovering. I just haven't found a good way display everything...

kingraphaii:
> Which PR info?
> 
> >

as per:

image

backend/frontend/templates/frontend/user/component_card.html
kingraphaii: I meant on extracting the https:// version of the repo. We have owner and repo name as fields on each Repository instance.

Ngoakor12: The owner is currently stored as `Umuzi.org` but the owner for valid links is `Umuzi-org` so I couldn't find any easy way to replace the dot with a dash to make the links valid :grimacing:

kingraphaii (approved):

backend/frontend/templates/frontend/components/main_header.html
```diff
@@ -47,12 +47,12 @@
                         From: "transform opacity-100 scale-100"
                         To: "transform opacity-0 scale-95"
                     -->
-                    <div class="absolute right-0 z-10 mt-2 w-48 origin-top-right rounded-md bg-white py-1 shadow-lg ring-1 ring-black ring-opacity-5 focus:outline-none" role="menu" aria-orientation="vertical" aria-labelledby="user-menu-button" tabindex="-1">
+                    <!-- <div class="absolute right-0 z-10 mt-2 w-48 origin-top-right rounded-md bg-white py-1 shadow-lg ring-1 ring-black ring-opacity-5 focus:outline-none" role="menu" aria-orientation="vertical" aria-labelledby="user-menu-button" tabindex="-1"> -->
```
sheenarbw: Why are you editing the main header in your PR?

backend/frontend/theme.py
```diff
@@ -12,4 +12,8 @@
     "heading3": f"text-l font-bold text-{TEXT_COLOUR}-600",
     # link
     "link": f"text-{PRIMARY_COLOUR}-600 hover:text-{PRIMARY_COLOUR}-500 focus:text-{PRIMARY_COLOUR}-500 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-{PRIMARY_COLOUR}-500 underline underline-offset-2",
+    "flavour_chip": "flex px-2 py-1 max-h-8",
+    "good_review_counter_chip": "font-bold bg-blue-600 text-white rounded-full flex justify-center items-center h-5 px-2 py-0 text-xs -translate-x-1/2 -translate-y-2/4 scale-100 top-0 right-0",
+    "bad_review_counter_chip": "font-bold bg-red-600 text-white rounded-full flex justify-center items-center h-5 px-2 py-0 text-xs -translate-x-1/2 -translate-y-2/4 scale-100 top-0 right-0",
```
sheenarbw: good_review_counter_chip and bad_review_counter_chip are only used in a few limited places and their classes are mostly the same.

It would be best to use this object for more generic things, eg links are used all over the place, headings too. These counters are only used in one place.

If you think we'll have a reason to show a lot of icons then it could make sense to add an "icon" to this, but it seems like it might still be too specific. 

If a class string is going to be used all over the place then it should be here. If a class name is likely to just be used in one place then it should be in that once place.

backend/frontend/theme.py
```diff
@@ -12,4 +12,8 @@
...
+    "small_bold_text": "text-xs font-bold text-gray-600",
```
sheenarbw: Even this seems quite specific. Do you think we'll be using this a lot? 

Try to make things as general and reusable as possible. Making one entry here too specific makes it less useful. 

There is nothing stopping us from doing something like this in a template:

```
<button 
    type="button" 
    class="{{ styles.button_secondary_small }} w-full"
   ...
```

Here we use the generic button_secondary_small class string then make it full width. We dont need to make a "button_secondary_small_full_width" entry and then use it only once.

backend/frontend/templatetags/component_card_extras.py
```diff
@@ -0,0 +1,13 @@
+# Custom template filters according to the docs here: https://docs.djangoproject.com/en/4.2/howto/custom-template-tags/
+
+from django import template
+
+register = template.Library()
+
+
+@register.filter(name="convert_git_url")
```
sheenarbw: This seems overly complicated and not cohesive. This filter is only going to be used once, right? On one specific field of one specific model. 

It should be defined in the model. Eg:

```
class Repository:
    ...
    def get_github_repo_link(self):
        return f"https://github.com/{self.owner.replace('.','-')}/{self.repo_name}"
```

Always remember KISS.

backend/frontend/templates/frontend/user/component_card.html
```diff
@@ -1,25 +1,173 @@
...
+            <p class="{{styles.small_bold_text}}">Reviewers</p>
+            {% for reviewer_name in card.reviewer_names %}
+                <p class="truncate" title="{{reviewer_name}}">{{reviewer_name}}</p> 
+            {% endfor %}
+        </div>
+    {% endif %}
 
+    {% if card.repo_url %}
+        <div class="flex gap-1">
+            <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path fill="currentColor" d="M12 2A10 10 0 0 0 2 12c0 4.42 2.87 8.17 6.84 9.5c.5.08.66-.23.66-.5v-1.69c-2.77.6-3.36-1.34-3.36-1.34c-.46-1.16-1.11-1.47-1.11-1.47c-.91-.62.07-.6.07-.6c1 .07 1.53 1.03 1.53 1.03c.87 1.52 2.34 1.07 2.91.83c.09-.65.35-1.09.63-1.34c-2.22-.25-4.55-1.11-4.55-4.92c0-1.11.38-2 1.03-2.71c-.1-.25-.45-1.29.1-2.64c0 0 .84-.27 2.75 1.02c.79-.22 1.65-.33 2.5-.33c.85 0 1.71.11 2.5.33c1.91-1.29 2.75-1.02 2.75-1.02c.55 1.35.2 2.39.1 2.64c.65.71 1.03 1.6 1.03 2.71c0 3.82-2.34 4.66-4.57 4.91c.36.31.69.92.69 1.85V21c0 .27.16.59.67.5C19.14 20.16 22 16.42 22 12A10 10 0 0 0 12 2Z"/></svg>
```
sheenarbw: Could we not use font awesome of something like that instead of writing these paths by hand?

If not then I would recommend ng up a library of icons ourselves so we can do things like say:

```
{{ githubIcon }}
```

I'm sure we will use a github icon many times over. 

One benefit to using something like fontawesome is that all our icons will have the same kind of style.  It would also be easy to look at the HTML and immediately know what is intended.

backend/frontend/templates/frontend/user/component_card.html
```diff
@@ -1,25 +1,173 @@
...
+                <p>
+                    {{card.due_time|naturalday}}, {{card.due_time|time}} 
+                </p>
+            {% endif %}
+        </div>
+    {% endif %}
+    
+    {% if card.request_user_is_assignee == False %}
+        <div class="w-full" >
+            <p class="{{styles.small_bold_text}}">Assignees</p>
```
sheenarbw: Should this not say "Assignee"? 

Tilde was initially built with the intention of allowing group work projects, but we are not actually using that. Maybe we will one day

backend/frontend/templates/frontend/user/component_card.html
```diff
@@ -1,25 +1,173 @@
...
+        <div class="flex gap-2" title="Due time">
+            <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path fill="currentColor" d="M12 20a7 7 0 0 1-7-7a7 7 0 0 1 7-7a7 7 0 0 1 7 7a7 7 0 0 1-7 7m0-16a9 9 0 0 0-9 9a9 9 0 0 0 9 9a9 9 0 0 0 9-9a9 9 0 0 0-9-9m.5 4H11v6l4.75 2.85l.75-1.23l-4-2.37V8M7.88 3.39L6.6 1.86L2 5.71l1.29 1.53l4.59-3.85M22 5.72l-4.6-3.86l-1.29 1.53l4.6 3.86L22 5.72Z"/></svg>
+            {% now "U" as current_timestamp %}
+            {% if current_timestamp > card.due_time|date:"U" %}
+                <p class="text-red-600">
+                    {{card.due_time|naturalday}}, {{card.due_time|time}} 
+                </p>
+                {% else %}
+                <p>
+                    {{card.due_time|naturalday}}, {{card.due_time|time}}
```
sheenarbw, line 132: This is a good use of filters

backend/frontend/templates/frontend/user/component_card.html
```diff
@@ -1,25 +1,173 @@
...
+                        {{card.code_review_competent_since_last_review_request}} 
+                    </div>
+                </div>    
+            {% endif %}
+            {% if card.code_review_excellent_since_last_review_request %}
+                <div class="flex" title="Number of competent code reviews since last review request">
+                    <div title="Number of excellent code reviews since last review request">
+                        <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path fill="currentColor" d="M19 10c0 1.38-2.12 2.5-3.5 2.5s-2.75-1.12-2.75-2.5h-1.5c0 1.38-1.37 2.5-2.75 2.5S5 11.38 5 10h-.75c-.16.64-.25 1.31-.25 2a8 8 0 0 0 8 8a8 8 0 0 0 8-8c0-.69-.09-1.36-.25-2H19m-7-6C9.04 4 6.45 5.61 5.07 8h13.86C17.55 5.61 14.96 4 12 4m10 8a10 10 0 0 1-10 10A10 10 0 0 1 2 12A10 10 0 0 1 12 2a10 10 0 0 1 10 10m-10 5.23c-1.75 0-3.29-.73-4.19-1.81L9.23 14c.45.72 1.52 1.23 2.77 1.23s2.32-.51 2.77-1.23l1.42 1.42c-.9 1.08-2.44 1.81-4.19 1.81Z"/></svg> 
+                    </div>
+                    <div class="{{styles.good_review_counter_chip}}">{{card.code_review_excellent_since_last_review_request}}
```
sheenarbw: This isn't a chip. It's an icon with a count. 

Hre are some examples of chips: https://www.google.com/search?sca_esv=587603400&sxsrf=AM9HkKmzFwAn0k3xCBWY3kq-yAgXp6_L5g:1701676906552&q=ui+chips&tbm=isch&source=lnms&sa=X&ved=2ahUKEwj3hqa8qPWCAxXnT0EAHScWBuYQ0pQJegQIDBAB&biw=1850&bih=976&dpr=1 

Chips are text surrounded by a friendly rounded border. Chips can have icons.

backend/curriculum_tracking/models.py
```diff
@@ -1626,10 +1626,25 @@ def request_user_can_start(self):
...
-        # TODO: return True if:
-        # the card is Ready
-        # the user is allowed to start the card (ie. they are an assignee, or have permission)
-        return True
+
+        has_manage_perm = [
+            user.has_perm(Team.PERMISSION_MANAGE_CARDS, team) for team in user.teams()
+        ]
+
+        return (self.can_start()) and (
```
sheenarbw: Think of database calls as expensive. They are slow. We dont want to make all the db calls and then make a decision if we could make a decision based on fewer calls.

Eg: If the current user is a card assignee then we could just return true. We dont need to know what teams the user is in, or what permissions it has. 

It would also be good to test this function.
You can make a few cards in a few different states and a few different users with different permissions and just make sure that the function returns True under the right circumstances.

sheenarbw (changes requested):
You are on the right track but there are a few things to learn

backend/frontend/templates/frontend/components/main_header.html
Ngoakor12: I was removing the distracting profile dropdown. I'll change it back when I'm done

backend/curriculum_tracking/models.py
```diff
@@ -1619,17 +1619,46 @@ def get_users_that_reviewed_open_prs(self):
...
+        """
+        if user is None:
+            from threadlocal_middleware import get_current_user
+
+            user = get_current_user()
+
+        if user is not None:
+            request_user_is_assignee = self.assignees.first() == user
+
+            if request_user_is_assignee and self.can_start():
```
kingraphaii: can_start() takes in force=Bool, for when user has super rights, do we not need that here as well?

backend/curriculum_tracking/models.py
```diff
@@ -1619,17 +1619,46 @@ def get_users_that_reviewed_open_prs(self):
...
+                return True
+
+            has_manage_cards_permission = any(
+                [
+                    user.has_perm(Team.PERMISSION_MANAGE_CARDS, team)
+                    for team in self.assignees.first().teams()
+                ]
+            )
+
+            if has_manage_cards_permission and self.can_start():
```
kingraphaii: We can cache the value of self.can_start() somewhere up top

backend/curriculum_tracking/models.py
```diff
@@ -1619,17 +1619,46 @@ def get_users_that_reviewed_open_prs(self):
...
+                return True
+
+        return False
+
+    def request_user_is_assignee(self):
```
kingraphaii, line 1659: Not seeing where we are making use of this. We manually did this in `request_user_can_start`:

```
request_user_is_assignee = self.assignees.first() == user
```

backend/curriculum_tracking/tests/test_request_user_can_start.py
kingraphaii, line 1: I think we also need to make sure that super users can start and test for it in here

kingraphaii (changes requested):

backend/curriculum_tracking/models.py
Ngoakor12: I've accounted for cards that can be "force started" and allowed super users and team managers to force start cards that can be "force started."

backend/curriculum_tracking/models.py
Ngoakor12: I'm doing it manually because there were clashes with the test environments and how we get the requesting user.

However, the actual method is still being used in the template.

I need to pass in a user and do the user checks I'm doing in the `request_user_can_start` to call it in without issues. I opted not to because that `request_user_is_assignee` is not being tested.

Maybe I'm thinking about it wrong, any suggestions are welcome.

kingraphaii (approved):

sheenarbw (approved):

PR #674, opened 2023-11-24
V2 login flow

kingraphaii:

## Description:

## I solemnly swear that:


backend/frontend/templates/frontend/auth/base.html
```diff
@@ -0,0 +1,22 @@
...
+      integrity="sha384-FhXw7b6AlE/jyjlZH5iHa/tTe9EpJ1Y55RjcgPbjeWMskSxZt1v9qkxLJWNJaGni"
+      crossorigin="anonymous"></script>
+
+    <script src="https://cdn.tailwindcss.com"></script>
+    <script src="https://cdn.jsdelivr.net/npm/chart.js"></script>
+    <script defer src="https://cdn.jsdelivr.net/npm/alpinejs@3.x.x/dist/cdn.min.js"></script>
+
+    {% block head %}{% endblock %}
+  </head>
+  <body class="h-full w-full">
```
sheenarbw: This looks like it's mostly a copy paste of the main template. Is there any way we can make it mre DRY using template inheritance?

backend/frontend/templates/frontend/auth/page_forgot_password.html
```diff
@@ -0,0 +1,22 @@
+{% extends "frontend/auth/base.html" %}
+<!DOCTYPE html>
```
sheenarbw: Why do you need the doctype here again?

backend/frontend/templates/frontend/auth/page_login.html
```diff
@@ -0,0 +1,24 @@
...
+<div class="mx-auto w-1/3 p-6 bg-white rounded">
+  <h1 class="text-lg">Login</h1>
+  {% if error %}
+  <p>{{ error }}</p>
+  {% endif %}
+
+  <form action="{% url 'user_login' %}" method="post">
+    {% csrf_token %}
+    <label for="username">Username</label>
+    <input class="border" type="text" name="username" id="username" />
```
sheenarbw: Django has form functionality doesn't it? Why not use that?

kingraphaii: Oh, yeah, meant to clean this up before marking it ready :)

backend/frontend/templates/frontend/auth/base.html
kingraphaii: There's some stuff I didn't want in there, like the main header and charting assets etc.

backend/frontend/templates/frontend/auth/page_password_reset_done.html
```diff
@@ -0,0 +1,6 @@
+{% extends "frontend/auth/base.html" %} {% block content %}
+<div class="mx-auto w-1/3 p-6 bg-white rounded">
+  <h1 class="text-lg">Check your inbox</h1>
+  <p>We've sent you an email with a link to reset your password.</p>
+</div>
+{% endblock %}
```
Ngoakor12: Regarding the copy on this page can we maybe mention that you'll get the email if you previously had an account or have an existing account too? Just so it's clearer. I tried to reset the password on a non-existant user and I was not fully satisfied with the feedback message.

I don't think it's a big issue and might only make sense in my head :sweat_smile:

Ngoakor12 (changes requested):
The login and logout functionality seem to work fine according to my understanding.

The forgot password one however does not send an actual email. What do I need to do to get an actual password rest email? Is it the same as the SIS project?

backend/frontend/templates/frontend/auth/base.html
sheenarbw: So why not change the base template and then put the chart stuff in where it is needed? This is a lot of copy-paste

backend/frontend/templates/frontend/components/main_header.html
```diff
@@ -1,43 +1,69 @@
...
-                </div>
-                <div class="hidden md:block">
-                  <div class="ml-10 flex items-baseline space-x-4">
-                    <!-- Current: "bg-gray-900 text-white", Default: "text-gray-300 hover:bg-gray-700 hover:text-white" -->
-                    <a href="#" class="bg-gray-900 text-white rounded-md px-3 py-2 text-sm font-medium" aria-current="page">Dashboard</a>
-                    <a href="#" class="text-gray-300 hover:bg-gray-700 hover:text-white rounded-md px-3 py-2 text-sm font-medium">Team</a>
-                    
-                  </div>
-                </div>
+<div class="bg-gray-800">
```
sheenarbw, line 2: Did you just change the indentation on this file? Or is there something to review here?

sheenarbw (changes requested):

backend/frontend/templates/frontend/auth/page_password_reset_done.html
kingraphaii: Regarding emails, the dev environment is currently set to use File Backend, and it dumps the emails sent as text files inside gitignore folder.

backend/frontend/templates/frontend/components/main_header.html
kingraphaii: I put sign_out actions in there, prettier must've made the diff huge on save

Ngoakor12 (changes requested):
I was able to locate the sent emails as mentioned :ok_hand: 

But...
- the URL in for reset was using example.com as the base URL, is that normal or can we point it to a global base URL variable? So that in dev mode we use localhost and the right base URL in prod?
- after updating/going to the right url, I get met with this page:
image
  - I know styling is not the main priority but can we make text boxes look like text boxes and buttons look like buttons so that we don't have to guess where the text goes and where the buttons are.
  - Regarding the errors,  I don't know if they just show up on first render or if I did something weird. Ideally I don't think they should show up without me having entered a password

kingraphaii:
> * the URL in for reset was using example.com as the base URL, is that normal or can we point it to a global base URL variable? So that in dev mode we use localhost and the right base URL in prod?

So django gets the current site from the Sites module. You should be able to update your site in the admin

> Regarding the errors, I don't know if they just show up on first render or if I did something weird. Ideally I don't think they should show up without me having entered a password

SO those werent errors but help text hence their showing before submission :). I've since removed them on new commits

Ngoakor12:
>> * the URL in for reset was using example.com as the base URL, is that normal or can we point it to a global base URL variable? So that in dev mode we use localhost and the right base URL in prod?

> So django gets the current site from the Sites module. You should be able to update your site in the admin

Ah I see, I hoped there would be a configuration of some sort in the settings file.

Ngoakor12 (approved):

backend/frontend/templates/frontend/auth/page_forgot_password.html
```diff
@@ -0,0 +1,23 @@
...
+  <hr class="my-2" />
+  {% for field, error in form.non_field_errors.items %}
+    <p>{{ error }}</p>
+  {% endfor %}
+
+  <form action="{% url 'user_forgot_password' %}" method="post">
+    {% csrf_token %}
+    <div class="flex mb-3 flex-col">
+      <label class="{{ styles.label_small }}" for="{{ form.email.id_for_label }}">{{ form.email.label }}</label>
+      <input placeholder="" value="{{ form.email.value|default:''}}" class="{{ styles.input_small }}" id="{{ form.email.id_for_label }}" name="email" type="email" required />
```
sheenarbw: Why are you writing the form html by hand? Django has a way to turn Form instances into html.

I assume you are just doing it this way because you want to use tailwind, but people have been styling django forms for as long as Django has been around. If there isn't a DRY way to do it I would be very surprised.

Crispy forms styles django forms. We shouldnt be using crispy forms because it's bootstrap I think, but it's definitely possible to get forms to look good without having to write all the hrml by hand.


For now you can just use as_div or whatever and leave it ugly, and then we can make a seperate issue about getting all forms to render nicely forevermore.

backend/frontend/templates/frontend/base.html
```diff
@@ -1,28 +1,15 @@
...
-<!doctype html>
-<html lang="en-US" class="h-full bg-gray-100">
-  <head>
-    <meta charset="utf-8" />
-    <title>Tilde ~</title>
-    <script src="https://unpkg.com/htmx.org@1.9.6" integrity="sha384-FhXw7b6AlE/jyjlZH5iHa/tTe9EpJ1Y55RjcgPbjeWMskSxZt1v9qkxLJWNJaGni" crossorigin="anonymous"></script> 
-    {# TODO https://htmx.org/docs/#installing #}
-
-    <script src="https://cdn.tailwindcss.com"></script>
+{% block main_head %}
```
sheenarbw: Now we have `_base` which is the base for everything and `base` which is the base of some stuff... what stuff? can you think of a better name? 

Is this file only used for pages where the user is already logged in? Maybe logged_in_base or something like that would be a better name. 

This app is already becoming unweildly and I'll be tidying up a bunch of things myself, but it would be good to at least do the obvious clean things.

sheenarbw (changes requested):
I'm concerned about writing the form html by hand. Django has all sorts of form functionality built in.

There is also a weird thing with template names.

backend/frontend/forms.py
```diff
@@ -0,0 +1,48 @@
...
+
+class ThemedFormMixin:
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        for field in self.fields:
+            self.fields[field].widget.attrs.update(
+                {
+                    "class": styles["input_small"],
+                }
+            )
```
Ngoakor12, line 21: This is a workaround for adding certain styles/classes on the fly?

Ngoakor12 (approved):
It would be nice for the error messages in `/reset-password` page styled like the `/login` ones.

Otherwise the functionality seems to be working and changes look addressed.

backend/frontend/forms.py
kingraphaii: yes, so that even if we leave django to render the form, we'd still get our custom classes

backend/frontend/templates/frontend/auth/page_forgot_password.html
```diff
@@ -0,0 +1,15 @@
+{% extends "frontend/auth/base.html" %} {% block content %}
+<div class="mx-auto w-1/3 p-6 bg-white rounded">
+  <h1 class="{{ styles.heading1 }}">Forgot Password</h1>
+  <hr class="my-2" />
+  {% for field, error in form.non_field_errors.items %}
```
sheenarbw: Will this work? Do non-field errors have fields? 

You do this differently in the different templates, eg: 
```
{% if form.non_field_errors %}
    <div class="border border-red-500 bg-red-100 rounded p-3 mb-4 text-sm">
      {% for error in form.non_field_errors %}
          <p class="text-red-500">{{ error }}</p>
      {% endfor %}
    </div>
  {% endif %}
```

I wonder if there is a way to put the repetitive form error stuff in another template and just include that as needed.

backend/frontend/templates/frontend/auth/page_login.html
```diff
@@ -0,0 +1,30 @@
+{% extends "frontend/auth/base.html" %} {% block content %}
+<div class="mx-auto w-1/3 p-6 bg-white rounded">
+  <h1 class="{{ styles.heading1 }}">Login</h1>
+  <hr class="my-2" />
+  {% if messages %}
```
sheenarbw: Hmm, would this not be better in a more foundational base template? Then all pages will have message functionality.  Eg I'm sure it would make sense to have messages when a user adds a review to a card, updates their user profile or any number of other things. Or do you think that kind of thing will be displayed completely differently? 

I'm not actually sure at this point.

backend/frontend/views.py
```diff
@@ -50,13 +64,111 @@
...
+        form = CustomAuthenticationForm(request=request, data=request.POST)
+        context.update({"form": form})
+
+        if form.is_valid():
+            login(
+                request=request,
+                user=form.user_cache,
+            )
+            return redirect(
+                reverse_lazy("user_board", kwargs={"user_id": form.user_cache.id})
```
sheenarbw: There might be a `next` get parameter in the url.

With the current deployed version, if I try to access the url `https://backend.tilde.umuzi.org/users_and_teams_nav/` and I am not logged in then I get redirected to 
https://backend.tilde.umuzi.org/accounts/login/?next=/users_and_teams_nav/.

The idea is that when I log in I should get redirected back to where I wanted to be. It's good to have a default location in case `next` is empty, but it would be a better user experience if we pay attention to what the user was trying to do.

backend/frontend/templates/frontend/auth/page_login.html
```diff
@@ -0,0 +1,30 @@
...
+    {% for message in messages %}
+        <p>
```
sheenarbw: a paragraph inside a paragraph? This is a new use of paragraph I have not come across before... I'm sure it'll render fine but, what are you actually trying to do here? 

Should this be an unordered list? Or a paragraph in a div? Or what?

sheenarbw (changes requested):
There are a few things that look a bit odd still. But you are getting there.

I think that it would be good to get this stuff merged and then let Sbonelo try to do an accessability audit on the code and fix up whatever is still weird in the HTML and styling. 

One thing to know is: There is already technical debt in this project that will need to be cleaned up. I hope that you will be able to spot technical debt and find ways to deal with it yourself as you work. Ideally, I shoulnd't have to point out things that look like they are getting messy. This relates to this PR and your other one.

backend/frontend/templates/frontend/auth/page_forgot_password.html
kingraphaii: Makes sense, will implement this in a part_ template

kingraphaii: Scratched this because {{ form.as_p }} will render the erorrs in a list with custom classes, maybe we can just write styles for those

Ngoakor12 (approved):

Ngoakor12:
@kingraphaii merge conflicts

sheenarbw (approved):

sheenarbw (approved):

PR #676, opened 2023-12-01
Restrict access to boards

kingraphaii:

## Description:

restricted access to user boards

## I solemnly swear that:


Ngoakor12 (approved):

backend/frontend/views.py
```diff
@@ -50,40 +54,93 @@
...
-from django.contrib.auth.decorators import user_passes_test
-
-
 def is_super(user):
     return user.is_superuser
 
 
-@user_passes_test(is_super)
-def user_board(request, user_id):
+def user_has_view_access(user, request):
```
sheenarbw: This is not super clear...
is `user` the logged in user? Or is it the user being viewed?

sheenarbw: And it looks like the only reason we are passing in request is so that we can access request.user. Why not just pass in the logged in user as an instance? 

eg: `user_has_view_access(viewed_user,logged_in_user)` or something like that?

backend/frontend/views.py
```diff
@@ -50,40 +54,93 @@
...
+        request.user.has_perm(Team.PERMISSION_ASSIGN_REVIEWERS, team)
+        for team in user_teams
+    ]
+    can_review_cards = [
+        request.user.has_perm(Team.PERMISSION_REVIEW_CARDS, team) for team in user_teams
+    ]
+    is_trusted_reviewer = [
+        request.user.has_perm(Team.PERMISSION_TRUSTED_REVIEWER, team)
+        for team in user_teams
+    ]
```
sheenarbw: This code seems quite inefficient. If a person has "manage cards" permission on a team then it would make sense to return immediately. Instead, we do many database queries before returning.

This function is going to be called really often, it needs to be quick!

If the viewed user is in 3 teams then this is 3x5 = 15 distinct db queries for every single permission check.

Read up on django guardian and see if there is a more efficient way to do it. And if you have to make so many queries in the worst case scenario then please just return as early as possible.

backend/frontend/views.py
```diff
@@ -50,40 +54,93 @@
...
 def is_super(user):
     return user.is_superuser
 
 
-@user_passes_test(is_super)
-def user_board(request, user_id):
+def user_has_view_access(user, request):
+    user_teams = user.teams()
+    can_manage_cards = [
+        request.user.has_perm(Team.PERMISSION_MANAGE_CARDS, team) for team in user_teams
```
sheenarbw: This is not DRY, why are you typing out all the view permissions one by one? They all exist in a data structure that can be imported and looped over. There is a good chance that we will be adding new permissions to that data structure in the near future, when we do that then this code will bug out unless someone remembers to come and update it. It's better to be DRY

backend/frontend/views.py
```diff
@@ -50,40 +54,93 @@
...
+        or any(can_view_all)
+        or any(can_assign_reviewers)
+        or any(can_review_cards)
+        or any(is_trusted_reviewer)
+    )
+
+    return has_view_access or user.id == request.user.id or request.user.is_superuser
+
+
+class UserBoard(LoginRequiredMixin, UserPassesTestMixin, TemplateView):
```
sheenarbw: This code is much longer and more complicated than the function-based version. What are the benefits of this approach versus just using a user_passes_test decorator? 

My understanding is that the only real reason to make use of class based views is to make use of inheritance and it doesn't look like we are making much use of that - we aren't putting our own functionality into a base class and inheriting from that. 

If there is a legit reason to use class based views then please say what that reason is. But generally we should try to keep things simple as much as possible.

backend/frontend/views.py
```diff
@@ -50,40 +54,93 @@
...
+    ]
+
+    has_view_access = (
+        any(can_manage_cards)
+        or any(can_view_all)
+        or any(can_assign_reviewers)
+        or any(can_review_cards)
+        or any(is_trusted_reviewer)
+    )
+
```
sheenarbw, line 86: The overall algorithm should be:

```
# do the cheap checks first. Doing this wont hit the db at all so get it out of the way

If user.id == request.user.id or request.user.is_superuser:
    return True 

# now do the expensive checks if needed 
```

sheenarbw (changes requested):
This needs a bit of work. It needs to be efficient and DRY. And I suspect it can be done a lot more simply

backend/frontend/views.py
kingraphaii: couldn't pass an extra argument to `user_passes_test` and also request is not accessible in there either, so I thought the CB version could do it without much deliberation. So I was thinking either this or putting the checks inside each view which wouldn't look good. Would creating a custom `user_passes_test` decorator that takes in the request object as well be a good implementation of this?

sheenarbw: Putting the checks inside each view wouldn't do.

I made a middleware that lets you get the currrent request. 

```
from threadlocal_middleware import get_current_request
request = get_current_request()
...
```

Does that help?  Or do you still think classes are better?

kingraphaii: Noice! This should do, getting the request was the only issue I had. Thanks!

Ngoakor12 (approved):

backend/frontend/views.py
```diff
@@ -50,22 +57,43 @@
...
 
-from django.contrib.auth.decorators import user_passes_test
-
-
 def is_super(user):
     return user.is_superuser
 
 
-@user_passes_test(is_super)
+def has_view_access(logged_user):
```
sheenarbw: logged_user? You mean logged_in_user?

sheenarbw: Also, sometimes a person can have view access over a whole team. This function name can be more explicit.

Maybe `can_view_user` or something like that? Then we can do another one for `can_view_team`. Or is the intention of this function to be used on user views and team views?

sheenarbw: I did originally put the is_super function in views.py, but it was because it is a 2 line function we will probably throw away. This is getting a bit long.

Can you think of a better way to organize the code? 

I also think it would be good to write tests.

sheenarbw (changes requested):
Much better. Now to make sure the code is well organised and tested.

backend/frontend/tests/test_restricted_user_board_access.py
```diff
@@ -0,0 +1,156 @@
...
+from .frontend_test_mixin import FrontendTestMixin
+
+
+class TestUserBoardControlledAccess(FrontendTestMixin):
+    def setUp(self):
+        super().setUp()
+
+        self.viewed_user_team = TeamFactory()
+        self.viewed_user = UserFactory(
+            email="learner.umuzi.org", is_staff=True  # TODO: remove this
```
Ngoakor12: A forgotten todo? If it should be there please leave a reason or more info just like login todo on line 117.

backend/frontend/tests/test_restricted_user_board_access.py
```diff
@@ -0,0 +1,156 @@
...
+        )
+        self.viewed_user.set_password(self.viewed_user.email)
+        self.viewed_user.save()
+        self.viewed_user_team.user_set.add(self.viewed_user)
+
+        self.user_without_access = UserFactory(
+            email="user_without_access.umuzi.org", is_staff=True  # TODO: remove this)
```
Ngoakor12: same here

Ngoakor12 (changes requested):

backend/frontend/tests/test_restricted_user_board_access.py
kingraphaii: Not really, more of a reminder. After we merge the proper login urls we wont need is_staff=True. This is sort of a workaround for getting a valid session via the admin portal

Ngoakor12 (approved):
There are some merge conflicts, but the changes look addressed.

Ngoakor12 (approved):

backend/frontend/views.py
```diff
@@ -69,6 +71,28 @@ def is_super(user):
...
+    if logged_in_user.id == viewed_user_id or logged_in_user.is_superuser:
+        return True
+
+    viewed_user_obj = get_object_or_404(User, pk=viewed_user_id)
+    viewed_user_teams = viewed_user_obj.teams()
+
+    if len(viewed_user_teams):
+        checker = ObjectPermissionChecker(logged_in_user)
+        checker.prefetch_perms(viewed_user_teams)
+        for view_permission in Team.PERMISSION_VIEW:
```
sheenarbw, line 115: This can still be more efficient. Here's something to try out:

```
def work():
     answer = random.choice([True,False])
     print(f"work - {answer}")
     return answer

# what is the difference between the following:
any([work() for _ in range(10)])
any((work() for _ in range(10)))
```

One is a list comprehension, and the other one uses a tuple. The second one is more like a generator. It wont do ALL the work up front.

Now imagine if `work` was an expensive function, eg, what if it queried a database..

TLDR; a tuple is a way better choice here. Not just slightly better.

sheenarbw: Also, I'm not sure, can has_perm accept a list? Like `checker.has_perm(Team.PERMISSION_VIEW, team)`. If it can then we should definately not loop.  You've probably already checked this, but figured I would mention it just in case.

backend/frontend/tests/test_restricted_user_board_access.py
```diff
@@ -0,0 +1,159 @@
+from guardian.shortcuts import assign_perm
+from core.models import Team
+from core.tests.factories import UserFactory, TeamFactory
+from .frontend_test_mixin import FrontendTestMixin
+
+
+class TestUserBoardControlledAccess(FrontendTestMixin):
```
sheenarbw: A better name would be something like TestUserBoard Permissions or ...Authorization.

Controllers are a whole thing.

backend/frontend/tests/test_restricted_user_board_access.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        self.do_login(self.user_without_access)
+        url = self.reverse_url("user_board", kwargs={"user_id": self.viewed_user.id})
+
+        self.page.goto(url)
+
+        login_url = "/login"
+        final_page_url = self.page.url
+
+        self.assertIn(login_url, final_page_url)
```
sheenarbw: What if the user is already logged in? Should something else happen? Eg, maybe they should get a page like "you do not have permission to view this page"

sheenarbw (changes requested):
There is one UX thing that seems a bit off, and an efficiency thing to do with how Python deals with tuples versus lists.

Ngoakor12:
Merge conflicts, otherwise, changes look addressed.

backend/frontend/views.py
```diff
@@ -72,6 +76,51 @@ def is_super(user):
...
+    frontend, or at least some of the frontend views.
+    """
+
+    def decorator(view_func):
+        @wraps(view_func)
+        def _wrapped_view(request, *args, **kwargs):
+            if test_func(request.user):
+                return view_func(request, *args, **kwargs)
+
+            return HttpResponseForbidden(
```
sheenarbw, line 93: If the user is logged in then they should get a forbidden message. If they are not logged in they should be directed to the login page.

sheenarbw (approved):

PR #677, opened 2023-12-08
Only display teams that the user is allowed to see

Ngoakor12:

## Description:

## I solemnly swear that:


backend/frontend/views.py
```diff
@@ -113,16 +115,53 @@ def users_and_teams_nav(request):
...
     current_team_count = int(request.GET.get("count", 0))
 
     all_teams = Team.objects.filter(active=True).order_by(
         "name"
     )  # TODO: only show teams that the current user is allowed to see
-    teams = all_teams[current_team_count : current_team_count + limit]
-    has_next_page = len(all_teams) > current_team_count + limit
+
+    is_superuser = user.is_superuser
+    is_assignee = any([team for team in all_teams if team in user_teams])
```
sheenarbw: This shouldn't be needed.  Users dont get to see the teams they are in unless they have permission to do so.

sheenarbw: Here is a better way to do it:

```
from guardian.shortcuts import get_objects_for_user

all_teams = Team.objects.filter(active=True).order_by(
        "name"
    ) 
if user.is_superuser:
    teams = all_teams
else:
    # we have to filter the teams
    teams = get_objects_for_user(
                user, Team.PERMISSION_VIEW, all_teams,
            )
```

sheenarbw (changes requested):
Looks like you are giving too much access to people. And there is a shortcut built into Guardian that makes this a small thing.

Please look at ObjectPermissionsFilter inside backend/core/filters.py if you want to see how it is used. It's a weird looking class, but the filter_queryset function is basically what we need here.

kingraphaii (approved):

kingraphaii (changes requested):
Merge conflicts :(

kingraphaii (approved):

backend/frontend/views.py
```diff
@@ -122,6 +122,26 @@ def can_view_user_board(logged_in_user):
     return False
 
 
+def can_view_team(logged_in_user):
+    request = get_current_request()
+    viewed_team_id = request.resolver_match.kwargs.get("team_id")
+
+    if logged_in_user.id == viewed_team_id or logged_in_user.is_superuser:
```
kingraphaii: Do we really need this comparison: `logged_in_user.id == viewed_team_id`

Or just that "user.is_superuser" and maybe is part of the team?

backend/frontend/views.py
```diff
@@ -274,7 +294,7 @@ def action_start_card(request, card_id):
     )
 
 
-@user_passes_test(is_super)
+@login_required()
 def users_and_teams_nav(request):
```
kingraphaii, line 298: Do we not need to make this view filter the allowed teams as well? Or maybe it's a different task?

kingraphaii (changes requested):

backend/frontend/views.py
Ngoakor12: I don't think there's a way to access the `team_id.` for this view from the can_view_team method.

Besides, the view is not responsible for fetching any teams. The "sub-view" `view_partial_teams_list` filters only the teams a user is allowed to see.

At least that's how I understand everything :see_no_evil:

kingraphaii (changes requested):
Seems to be working as expected now, but just to take you back a bit, I just thought this is relevant for this PR. The already existing test `test_super_sees_all_teams` may fail depending on how fast we get back the teams from the htmx ajax call to `view_partial_teams_list`. To cater for this, I've found we can use this:

`self.page.wait_for_load_state("networkidle")`

before we get the body: `body = self.page.text_content("body")`

kingraphaii (approved):

sheenarbw (approved):

PR #679, opened 2024-01-09
card_component: make start button work

Ngoakor12:

## Description:

## I solemnly swear that:


backend/frontend/templates/frontend/user/board/view_partial_action_card_moved.html
```diff
@@ -1,5 +0,0 @@
-{% include "frontend/user/board/partial_card.html" %}
-
-<script>
```
sheenarbw: This should work for moving a card from any column to any other column. It would be best to just use this script as is. It does work so long as the card status is correct (and the card status should be correct)

backend/frontend/templates/frontend/user/board/js_exec_action_start_card.html
```diff
@@ -0,0 +1,5 @@
+{% include "frontend/user/board/partial_card.html" %}
+
+<script>
+  moveCardToCorrectColumn("{{card.id}}", "{{card.IN_PROGRESS}}");
```
sheenarbw: This script should not be needed. We have a general one that will work for moving the card to any column.

sheenarbw (changes requested):
One weird thing I noticed on first pass

backend/frontend/templates/frontend/user/board/js_exec_action_start_card.html
kingraphaii, line 1: Can we rename this to "js_exec_action_card_moved" instead? We may need the same js executed for when we request reviews.

So we'd have this js instead:
```js
    moveCardToCorrectColumn("{{card.id}}","{{card.status}}")
```

backend/frontend/views.py
```diff
@@ -259,15 +261,27 @@ def view_partial_user_board_column(request, user_id, column_id):
     )
 
 
-@user_passes_test(is_super)
 @csrf_exempt
 def action_start_card(request, card_id):
```
kingraphaii, line 335: Don't we need to check is the action is permitted? Like how the DRF does it with perms like:

```python
permission_classes=[
            (
                curriculum_permissions.IsCardAssignee
                & curriculum_permissions.CardCanStart
                & curriculum_permissions.AssigneeHasFinishedReviews
            )
            | (
                HasObjectPermission(
                    permissions=Team.PERMISSION_MANAGE_CARDS,
                    get_objects=_get_teams_from_card,
                )
                & curriculum_permissions.CardCanForceStart
            )
        ],
```

kingraphaii (changes requested):

backend/frontend/views.py
Ngoakor12: I will have a closer at protecting the view

backend/frontend/templates/frontend/user/board/js_exec_action_start_card.html
Ngoakor12: My bad. It wasn't working at one point. I guess I just needed to take a closer look at the implementation

backend/frontend/views.py
```diff
@@ -280,15 +311,28 @@ def view_partial_user_board_column(request, user_id, column_id):
...
-    # TODO implement this
+
+    content_item_type = card.content_item.content_type
+
+    if content_item_type == ContentItem.TOPIC:
+        card.start_topic()
+    elif content_item_type == ContentItem.PROJECT:
+        card.start_project()
+    else:
+        raise NotImplemented(
```
kingraphaii, line 346: did you mean NotImplementedError()?

backend/frontend/views.py
```diff
@@ -113,14 +121,39 @@ def can_view_user_board(logged_in_user):
...
+    request = get_current_request()
+    card_id = request.resolver_match.kwargs.get("card_id")
+
+    card = get_object_or_404(AgileCard, pk=card_id)
+    card_assignees = card.assignees.all()
+
+    if (
+        (logged_in_user in card_assignees)
+        and card.can_start()
+        and not len(agile_card_reviews_outstanding(logged_in_user))
```
kingraphaii: From your suggestion on the "Review" btn PR, we may need these inside the actual view itself so that we can return appropriate messages. I've implemented some form of alerts on that same PR, maybe you can check them out and let me know what you think

kingraphaii (changes requested):

backend/frontend/static/frontend/user_board.js
```diff
@@ -21,3 +21,61 @@ function moveCardToCorrectColumn(cardId, column) {
...
+  const alert = document.createElement("div");
+
+  const buttons = card.querySelectorAll('button');
+
+  let startButton
+  for(let i=0;i<buttons.length;i++){
+    if (buttons[i].textContent.trim()==="Start"){
+      startButton=buttons[i]
+    }
+  }
```
kingraphaii: I understand why this may be nice to have but I think we can do without it because it can pose other problems for us, like when we want to reuse the same notification util. when the "request review" button is pressed instead, or any other button that;s not the start button. So either we get rid of this or we find a "catch all" solution that will just disable all buttons inside the card.

backend/frontend/static/frontend/user_board.js
```diff
@@ -21,3 +21,61 @@ function moveCardToCorrectColumn(cardId, column) {
...
+  `;
+
+  card.parentNode.insertBefore(alert, card);
+
+  if (alertType === "error" || alertType === "warning") {
+    // shake the card to draw attention to the error
+    card.classList.add("shake");
+    alert.classList.add("shake");
+    // disable buttons
+    startButton.classList.add("disable-button")
```
kingraphaii: as for this, if we still need it, we can use the tailwind "disabled:styles" selector so that we use tailwind wherever we can in which case we'd just disable the button.

kingraphaii (changes requested):

backend/frontend/theme.py
```diff
@@ -20,4 +20,5 @@
     "alert_warning": f"bg-gray-100 border border-gray-400 text-gray-700 px-4 py-3 rounded relative",
     "alert_success": f"bg-green-100 border border-green-400 text-green-700 px-4 py-3 rounded relative",
     "alert_error": f"bg-red-100 border border-red-400 text-red-700 px-4 py-3 rounded relative",
+    "disabled_button": "disabled:text-opacity-70 disabled:bg-opacity-50",
```
kingraphaii: We can still put these on the same "button_primary_small" key.

kingraphaii (changes requested):

kingraphaii (approved):

backend/frontend/static/frontend/user_board.js
```diff
@@ -21,3 +21,54 @@ function moveCardToCorrectColumn(cardId, column) {
...
+    error: "fa-times-circle",
+    warning: "fa-exclamation-circle",
+    info: "fa-info-circle",
+  };
+
+  const card = document.getElementById(`card_${cardId}`);
+  const alert = document.createElement("div");
+  const cardButtons = card.querySelectorAll("button");
+
+  cardButtons.forEach((cardButton) => (cardButton.disabled = true));
```
sheenarbw: Why disable and re-enable the buttons? What if the user wants to immediately click the details button? They would have to wait.

backend/frontend/tests/test_card_start_button.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+
+        self.card.assignees.set([self.user])
+
+    def test_start_button_moves_topic_card_to_ip_column(self):
+        self.make_topic_card()
+
+        can_start = self.card.request_user_can_start(self.user)
+
+        if can_start:
```
sheenarbw: If the current user can't start the card then this test will always pass, and it will test nothing.

If the user cant start the card then it means the test wasn't set up properly and it should raise an error.

backend/frontend/tests/test_card_start_button.py
```diff
@@ -0,0 +1,71 @@
...
+
+            self.assertIn(topic_card_title, ip_column)
+            self.assertNotIn(topic_card_title, backlog_column)
+
+    def test_start_button_moves_project_card_to_ip_column(self):
+        self.make_project_card(ContentItem.LINK)
+
+        can_start = self.card.request_user_can_start(self.user)
+
+        if can_start:
```
sheenarbw: Same here.  If the user can't move this card then this test isn't testing anything

backend/frontend/static/frontend/user_board.js
```diff
@@ -21,3 +21,54 @@ function moveCardToCorrectColumn(cardId, column) {
...
+
+  alert.className = `mb-3 relative leading-tight flex gap-2 rounded-xl border p-3 text-sm ${alertThemes[alertType]}`;
+
+  alert.innerHTML = `
+      <i class="fas ${iconThemes[alertType]} self-center"></i>
```
sheenarbw: This works but it is not ideal that the html is embedded in a js file.

Is there a way to do this with templates rather?

Eg have a template that draws the card and the alert, and also calls a function to hide the alert after 5 seconds? 

The template would look something like:

```
{% include "frontend/user/board/partial_alert.html" %}
{% include "frontend/user/board/partial_card.html" %}

<script>
  hideCardAlert()
</script>
```

Then we could reuse the alert in different places in the app if needed.

sheenarbw: Another option is to include the alert stuff inside partial_card.html and it only shows up if there is a message.

backend/frontend/tests/test_card_start_button.py
```diff
@@ -0,0 +1,71 @@
...
+            self.page.click("text=Start")
+
+            self.page.wait_for_load_state("networkidle")
+
+            ip_column = self.page.text_content("div#column_IP")
+            backlog_column = self.page.text_content("div#column_RB")
+            project_card_title = self.card.content_item.title
+
+            self.assertIn(project_card_title, ip_column)
+            self.assertNotIn(project_card_title, backlog_column)
```
sheenarbw: Please also test that the actions are logged. We need to log that the card was started.

There are also some things you did to stop cards from being started when the user has outstanding reviews. Should that not be tested?

backend/frontend/tests/test_card_start_button.py
sheenarbw: If can_start does behave badly then it could be useful to write tests for it.

sheenarbw (changes requested):
There are a few issues with tests, and there is a code architecture decision that I thin can be better. This is looking good though, you're getting close.

Could you also please include a video/screenshots of the gui in your PR?

backend/frontend/views.py
```diff
@@ -113,14 +121,34 @@ def can_view_user_board(logged_in_user):
...
-                (checker.has_perm(view_permission, team)
-                 for team in viewed_user_teams)
+                (checker.has_perm(view_permission, team) for team in viewed_user_teams)
             ):
                 return True
 
     return False
 
 
+def user_can_start_card(logged_in_user):
```
sheenarbw: This is strange. There is a function that does this already:

AgileCard.request_user_can_start 

Why the duplication?

sheenarbw (changes requested):
missed one thing

backend/frontend/views.py
sheenarbw: I now see that it is a permission thing. But the code still isn't dry. It also needs a better naming convention.

kingraphaii:
Looking good, but I'm failing to run tests, not sure where the issue is, here is the error I get:

```traceback
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
psycopg2.errors.UndefinedTable: relation "core_user" does not exist


The above exception was the direct cause of the following exception:

  File "/Users/kingraphaii/Dev/Tilde/backend/manage.py", line 31, in <module>
    main()
  File "/Users/kingraphaii/Dev/Tilde/backend/manage.py", line 27, in main
    execute_from_command_line(sys.argv)
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/core/management/__init__.py", line 395, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/core/management/commands/test.py", line 23, in run_from_argv
    super().run_from_argv(argv)
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/core/management/base.py", line 330, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/core/management/base.py", line 371, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/core/management/commands/test.py", line 53, in handle
    failures = test_runner.run_tests(test_labels)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/test/runner.py", line 695, in run_tests
    old_config = self.setup_databases(aliases=databases)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/test/runner.py", line 614, in setup_databases
    return _setup_databases(
           ^^^^^^^^^^^^^^^^^
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/test/utils.py", line 170, in setup_databases
    connection.creation.create_test_db(
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/db/backends/base/creation.py", line 72, in create_test_db
    call_command(
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/core/management/__init__.py", line 168, in call_command
    return command.execute(*args, **defaults)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/core/management/base.py", line 371, in execute
    output = self.handle(*args, **options)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/core/management/base.py", line 85, in wrapped
    res = handle_func(*args, **kwargs)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/core/management/commands/migrate.py", line 214, in handle
    self.sync_apps(connection, executor.loader.unmigrated_apps)
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/core/management/commands/migrate.py", line 336, in sync_apps
    with connection.schema_editor() as editor:
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/db/backends/base/schema.py", line 115, in __exit__
    self.execute(sql)
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/db/backends/base/schema.py", line 142, in execute
    cursor.execute(sql, params)
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 66, in execute
    return self._execute_with_wrappers(sql, params, many=False, executor=self._execute)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 75, in _execute_with_wrappers
    return executor(sql, params, many, context)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 79, in _execute
    with self.db.wrap_database_errors:
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/db/utils.py", line 90, in __exit__
    raise dj_exc_value.with_traceback(traceback) from exc_value
  File "/Users/kingraphaii/Dev/Tilde/backend/venv/lib/python3.11/site-packages/django/db/backends/utils.py", line 84, in _execute
    return self.cursor.execute(sql, params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
django.db.utils.ProgrammingError: relation "core_user" does not exist
```

Ngoakor12:
Same issue and solution mentioned [here](https://github.com/Umuzi-org/Tilde/pull/698#issuecomment-1968345059) because the other PR has not been merged yet.

kingraphaii:
> Same issue and solution mentioned [here](https://github.com/Umuzi-org/Tilde/pull/698#issuecomment-1968345059) because the other PR has not been merged yet.

Seems to be coming from a recent PR merged into develop, the one on review claims

sheenarbw (approved):

PR #690, opened 2024-01-22
Add request review functionality to card

kingraphaii:

## Description:

This PR adds the "request review" button along with a minimal "alert" api for cards.

## Screenshots/videos

image

## I solemnly swear that:

### DIFF
--- a/backend/curriculum_tracking/models.py
+++ b/backend/curriculum_tracking/models.py
@@ -1667,6 +1667,34 @@ class AgileCard(
 
         return self.assignees.first() == user
 
+    def request_user_can_request_review(self, user=None):
+        """
+        Check if current user can request review for this card
+        """
+        if self.status not in [AgileCard.IN_PROGRESS, AgileCard.REVIEW_FEEDBACK]:
+            return False
+
+        if user is None:
+            from threadlocal_middleware import get_current_user
+
+            user = get_current_user()
+
+        if user is not None:
+            is_assignee = user in self.assignees.all()
+            is_superuser = user.is_superuser
+
+            has_manage_cards_permission = any(
+                (
+                    user.has_perm(Team.PERMISSION_MANAGE_CARDS, team)
+                    for team in self.get_teams()
+                )
+            )
+
+            if is_assignee or is_superuser or has_manage_cards_permission:
+                return True
+
+        return False
+
 
 class BurndownSnapshot(models.Model):
     MIN_HOURS_BETWEEN_SNAPSHOTS = 4
--- a/backend/curriculum_tracking/models.py
+++ b/backend/curriculum_tracking/models.py
@@ -1671,6 +1671,9 @@ class AgileCard(
         """
         Check if current user can request review for this card
         """
+        if self.content_type_nice != "project":
+            return False
+        
         if self.status not in [AgileCard.IN_PROGRESS, AgileCard.REVIEW_FEEDBACK]:
             return False
 
--- a/backend/curriculum_tracking/models.py
+++ b/backend/curriculum_tracking/models.py
@@ -1681,7 +1681,9 @@ class AgileCard(
 
         if user is not None:
             is_assignee = user in self.assignees.all()
-            is_superuser = user.is_superuser
+
+            if is_assignee:
+                return True
 
             has_manage_cards_permission = any(
                 (
@@ -1690,8 +1692,7 @@ class AgileCard(
                 )
             )
 
-            if is_assignee or is_superuser or has_manage_cards_permission:
-                return True
+            return has_manage_cards_permission
 
         return False
 
--- a/backend/frontend/static/frontend/user_board.js
+++ b/backend/frontend/static/frontend/user_board.js
@@ -1,7 +1,7 @@
 /*
 This gets called when we do anything to update the state of a card. It just moves the card into the correct column and puts it right at the top.
 
-TODO: Preserve the order of the cards in the column  
+TODO: Preserve the order of the cards in the column
 */
 function moveCardToCorrectColumn(cardId, column) {
   const card = document.getElementById(`card_${cardId}`);
@@ -22,50 +22,14 @@ function moveCardToCorrectColumn(cardId, column) {
   }
 }
 
-
 /*
-This inserts an alert above the card being acted on.
-It's used to show any errors or success messages that 
-come back from the server.
+  Hides a card alert after 5 seconds.
 */
-function showCardAlert({ cardId, message, alertType = "info", persist = true }) {
-  const alertThemes = {
-    "success": "border-green-300 bg-green-100 text-green-800",
-    "error": "border-red-300 bg-red-100 text-red-800",
-    "warning": "border-yellow-300 bg-yellow-100 text-yellow-800",
-    "info": "border-blue-300 bg-blue-100 text-blue-800",
-  }
 
-  const iconThemes = {
-    "success": "fa-check-circle",
-    "error": "fa-times-circle",
-    "warning": "fa-exclamation-circle",
-    "info": "fa-info-circle",
-  }
+function hideCardAlert(cardId) {
+  const cardAlert = document.querySelector(`[data-card-id="${cardId}"]`);
 
-  const card = document.getElementById(`card_${cardId}`);
-  const alert = document.createElement("div");
-
-  alert.className = `mb-3 relative leading-tight flex gap-2 rounded-xl border p-3 text-sm ${alertThemes[alertType]}`;
-
-  alert.innerHTML = `
-      <i class="fas ${iconThemes[alertType]} self-center"></i>
-      <p>${message}</p>
-  `;
-
-  card.parentNode.insertBefore(alert, card);
-
-  if (alertType === "error" || alertType === "warning") {
-    // shake the card to draw attention to the error
-    card.classList.add("shake");
-    alert.classList.add("shake");
-  }
-
-
-  if (!persist) {
-    // clear the alert after 5 seconds if it's not a persistent alert
-    setTimeout(() => {
-      alert.remove();
-    }, 5000);
-  }
+  setTimeout(() => {
+    cardAlert.remove();
+  }, 5000);
 }
--- a/backend/frontend/static/frontend/user_board.js
+++ b/backend/frontend/static/frontend/user_board.js
@@ -21,3 +21,51 @@ function moveCardToCorrectColumn(cardId, column) {
     destinationColumn.insertBefore(card, destinationColumn.firstChild);
   }
 }
+
+
+/*
+This inserts an alert above the card being acted on.
+It's used to show any errors or success messages that 
+come back from the server.
+*/
+function showCardAlert({ cardId, message, alertType = "info", persist = true }) {
+  const alertThemes = {
+    "success": "border-green-300 bg-green-100 text-green-800",
+    "error": "border-red-300 bg-red-100 text-red-800",
+    "warning": "border-yellow-300 bg-yellow-100 text-yellow-800",
+    "info": "border-blue-300 bg-blue-100 text-blue-800",
+  }
+
+  const iconThemes = {
+    "success": "fa-check-circle",
+    "error": "fa-times-circle",
+    "warning": "fa-exclamation-circle",
+    "info": "fa-info-circle",
+  }
+
+  const card = document.getElementById(`card_${cardId}`);
+  const alert = document.createElement("div");
+
+  alert.className = `mb-3 relative leading-tight flex gap-2 rounded-xl border p-3 text-sm ${alertThemes[alertType]}`;
+
+  alert.innerHTML = `
+      <i class="fas ${iconThemes[alertType]} self-center"></i>
+      <p>${message}</p>
+  `;
+
+  card.parentNode.insertBefore(alert, card);
+
+  if (alertType === "error" || alertType === "warning") {
+    // shake the card to draw attention to the error
+    card.classList.add("shake");
+    alert.classList.add("shake");
+  }
+
+
+  if (!persist) {
+    // clear the alert after 5 seconds if it's not a persistent alert
+    setTimeout(() => {
+      alert.remove();
+    }, 5000);
+  }
+}
--- a/backend/frontend/static/frontend/utils.js
+++ b/backend/frontend/static/frontend/utils.js
@@ -5,15 +5,3 @@ function countLoadedPageItems(parentId) {
   const parent = document.getElementById(parentId);
   return parent.childElementCount - 1; // the button is an element too
 }
-
-/*
- * Swaps in the body of error pages returned from htmx requests 
- * for example when an action is forbidden.
- */
-document.addEventListener("htmx:beforeOnLoad", function (event) {
-  const xhr = event.detail.xhr
-  if (xhr.status == 500 || xhr.status == 403 || xhr.status == 404) {
-    event.stopPropagation() // Tell htmx not to process these requests
-    document.children[0].innerHTML = xhr.response // Swap in body of response instead
-  }
-})
--- a/backend/frontend/static/frontend/utils.js
+++ b/backend/frontend/static/frontend/utils.js
@@ -5,3 +5,15 @@ function countLoadedPageItems(parentId) {
   const parent = document.getElementById(parentId);
   return parent.childElementCount - 1; // the button is an element too
 }
+
+/*
+ * Swaps in the body of error pages returned from htmx requests 
+ * for example when an action is forbidden.
+ */
+document.addEventListener("htmx:beforeOnLoad", function (event) {
+  const xhr = event.detail.xhr
+  if (xhr.status == 500 || xhr.status == 403 || xhr.status == 404) {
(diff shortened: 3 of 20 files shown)


backend/curriculum_tracking/models.py
```diff
@@ -1667,6 +1667,34 @@ def request_user_is_assignee(self):
...
+                    user.has_perm(Team.PERMISSION_MANAGE_CARDS, team)
+                    for team in self.get_teams()
+                )
+            )
+
+            if is_assignee or is_superuser or has_manage_cards_permission:
+                return True
+
+        return False
+
```
Ngoakor12: - Should we not return early if the user is the assignee before checking anything else?
- I'm not sure this button is supposed to show for superusers or team managers. Maybe something to confirm with Sheena? I just haven't seen it in the UI so far.
- Can we maybe add a test for this method similar to the one for `request_user_can_start` would be useful too?

backend/frontend/views.py
```diff
@@ -274,6 +277,59 @@ def action_start_card(request, card_id):
     )
 
 
+def user_can_request_review(logged_in_user):
+    if len(helpers.agile_card_reviews_outstanding(logged_in_user)):
+        return False
+    if len(helpers.pull_request_reviews_outstanding(logged_in_user)):
+        return False
```
Ngoakor12: I think we can ignore these checks if we want to remind the learners to review outstanding reviews before starting in the UI like we currently are. It's probably possible but I think the button should be there for us to have an opportunity to remind them.

I don't know if I'm explaining this well or not, so let me know if it doesn't make sense

Ngoakor12 (changes requested):
The `Request Review` button is showing up on all the assignee's cards in "in progress" and "review feedback". There should be an exception: topic cards that don't require a review. Those cards will have a "done" button instead.

backend/frontend/views.py
kingraphaii: I did not quite get you, but this isn't enforced at the model level, and as per the current setup, one shouldn't be able to move it to IR, hence its presence here. 

Also, the fact that this is on the "action" view, not the main one in charge of displaying the cards, means there's still that visual cue for the outstanding IR cards.

Let me know if we're talking about the same thing 😅

Ngoakor12: Lol, let me try again.

The color being a visual cue is all well and good. I just think it is better to be explicit about why your card is not able to move to review with something like a toast or a modal. Especially for new users who just know they can move cards from 1 column to the other.

Also, my thinking was around how easy it would be to implement that "explicit UI feedback" mentioned above. My thinking is we would still have to do some checks with the above helper functions again since the ones used in the above function only checks a lot of stuff including the other permissions.

Ngoakor12: I just saw the logic for the permission classes in the api_views disregard the above review, I guess? :sweat_smile:

kingraphaii: Got it

backend/curriculum_tracking/models.py
kingraphaii: > * I'm not sure this button is supposed to show for superusers or team managers. Maybe something to confirm with Sheena? I just haven't seen it in the UI so far.

I did this following the existing implementation here (in the current DRF api views):

```python
permission_classes=[
    (
        curriculum_permissions.IsCardAssignee
        & curriculum_permissions.AssigneeHasFinishedReviews
    )
    | HasObjectPermission(
        permissions=Team.PERMISSION_MANAGE_CARDS,
        get_objects=_get_teams_from_card,
    )
]
```

Refer to the `or` expression after the `|`

Ngoakor12: Apologies, I thought your response was related to my last comment, but now I see that it did not send. I did actually end up seeing the logic for API view access.

image

backend/frontend/tests/test_card_request_review_button.py
```diff
@@ -0,0 +1,85 @@
...
+        self.page.click("text=Request Review")
+
+        self.page.wait_for_load_state("networkidle")
+
+        rf_column = self.page.text_content("div#column_RF")
+        review_column = self.page.text_content("div#column_IR")
+        card_title = self.card.content_item.title
+
+        self.assertIn(card_title, review_column)
+        self.assertNotIn(card_title, rf_column)
```
Ngoakor12: Can you add a test for the card movements when there are outstanding reviews. I can't seem to move a card even after adding a review on an outstanding card

backend/frontend/views.py
```diff
@@ -1,4 +1,3 @@
-import json
```
Ngoakor12: This module should not be removed,  it's used by the dashboard

Ngoakor12 (changes requested):

backend/frontend/static/frontend/utils.js
```diff
@@ -5,3 +5,15 @@ function countLoadedPageItems(parentId) {
...
+
+/*
+ * Swaps in the body of error pages returned from htmx requests 
+ * for example when an action is forbidden.
+ */
+document.addEventListener("htmx:beforeOnLoad", function (event) {
+  const xhr = event.detail.xhr
+  if (xhr.status == 500 || xhr.status == 403 || xhr.status == 404) {
+    event.stopPropagation() // Tell htmx not to process these requests
+    document.children[0].innerHTML = xhr.response // Swap in body of response instead
```
sheenarbw: Let's say we have a page that uses htmx to fetch 10 graphs from the backend as it loads. If graph number 8 comes up with an error then the entire page will show that wrror. But the other 9 graphs could be useful.

Generally with frontend dev you want to make sure that errors are well contained. If one little widget is broken you dont want the whole page to break.   Here you seem to be explicitly breaking the whole page. 

What are you trying to achieve with this?

backend/frontend/templates/frontend/user/board/partial_card.html
```diff
@@ -166,5 +166,17 @@
             >Start
         </button>
     {% endif %}
+    
+    {% if card.content_type_nice == "project" or card.content_type_nice == "topic" and card.topic_needs_review %}
```
sheenarbw: We need to get rid of topic_needs_review. It's debt. Please don't build it into new logic

backend/frontend/views.py
```diff
@@ -295,6 +297,77 @@ def action_start_card(request, card_id):
...
+            {
+                "card": card,
+                "message": "Please make sure you have reviewed assigned pull requests.",
+                "alert_type": "error",
+            },
+        )
+
+    if card.recruit_project:
+        card.recruit_project.request_review(force_timestamp=timezone.now())
+    elif card.topic_progress:
```
sheenarbw: The request review button should not show up for topics. If there is a topic_progress instance it would be an error.

For topics you should just be able to say that they are finished, it's a different button

backend/frontend/views.py
```diff
@@ -295,6 +297,77 @@ def action_start_card(request, card_id):
...
+        card.finish_topic()
+    else:
+        raise NotImplementedError()
+
+    log_creators.log_card_review_requested(card=card, actor_user=request.user)
```
sheenarbw, line 352: it would be a bug to log that a review was requested if a topic was finished.  We need to be careful to be explicit with these logs

sheenarbw (changes requested):
I just had a quick look. I didn't review your tests.

backend/frontend/static/frontend/utils.js
kingraphaii: Had not thought about this, thank you for pointing it out. I wanted a way to communicate errors for example we have this guard on a view that gets called via ajax:

```python
@user_passes_test_or_forbidden(user_can_request_review)
def action_request_review(request, card_id):
    ...
```

if the request user fails the test, then the user won't know what happened, the error will only be visible in the browser console. 

Is this something we should worry about on this PR or we can leave it out for later?

sheenarbw: That's a good point. Bt I do think having it fail in this way is much better than breaking hydration. We use hydration already and I'll sure we'll use it more.

I think the thing to do is treat things differently per error. Eg a 500 error is something that really shouldn't happen, it means our backend code fell over. If those happen we can just have an alert saying "something went wrong" or something like that. 

500 errors will get picked up in our error logs and I'll get an email if there are any new ones so things will get fixed.

404s really shouldn't happen in an xhr request because we'll only really be fetching things we are allowed to fetch, with known ids. This would also actually be a bug I think. Same for 403s.  

A 403/4 can also come from a user navigating to a page they are not allowed to see but that should not be handled here.

I think for now it's ok to remove this event handler.

Ngoakor12 (approved):
Changes addressed and the tests are passing

backend/frontend/static/frontend/user_board.js
```diff
@@ -21,3 +21,51 @@ function moveCardToCorrectColumn(cardId, column) {
...
   }
 }
+
+
+/*
+This inserts an alert above the card being acted on.
+It's used to show any errors or success messages that 
+come back from the server.
+*/
+function showCardAlert({ cardId, message, alertType = "info", persist = true }) {
```
sheenarbw: This is similar to some code from Ngoako that I just reviewed.  You shouldn't be implementing the same thing in 2 different PRs.

backend/frontend/templates/frontend/user/board/js_exec_action_card_moved.html
```diff
@@ -0,0 +1,5 @@
+{% include "frontend/user/board/partial_card.html" %}
+
+<script>
```
sheenarbw, line 3: Hmm.. same comment as above. Also there is already a file that does this, now the code isn't DRY.

backend/frontend/templates/frontend/user/board/js_exec_action_show_card_alert.html
```diff
@@ -0,0 +1,5 @@
+{% include "frontend/user/board/partial_card.html" %}
+
+<script>
+  showCardAlert({cardId: "{{card.id}}", message: "{{ message }}", alertType: "{{ alert_type }}", persist: {{ persist|default:"false" }} });
```
sheenarbw: feedback given to Ngoako already

backend/frontend/views.py
```diff
@@ -295,6 +298,75 @@ def action_start_card(request, card_id):
     )
 
 
+def user_can_request_review(logged_in_user):
```
sheenarbw: I think we need a naming convention for these permission functions. It's really not clear what the intention of this code is until I go search for where you used it. 

Can you think of a better naming convention? It's also not clear from the name that a card_id is required here. 

Ideally I should know what a function is supposed to do just by looking at the name.

backend/frontend/views.py
```diff
@@ -295,6 +298,75 @@ def action_start_card(request, card_id):
...
+    """The card is in progress or review feedback and the user has chosen to request review"""
+    card = get_object_or_404(AgileCard, id=card_id)
+
+    if len(helpers.agile_card_reviews_outstanding(request.user)):
+        return render(
+            request,
+            "frontend/user/board/js_exec_action_show_card_alert.html",
+            {
+                "card": card,
+                "message": "You have outstanding card reviews.",
```
sheenarbw: This kind of logic also exists in Ngoako's PR. It wont be DRY. We need functionality to be centralised and referenced, not repeated.

backend/frontend/templates/frontend/user/board/page.html
```diff
@@ -1,14 +1,37 @@
 {% extends "frontend/user/base.html" %}
 
+{% block head %}
+    {{ block.super }}
+    <style>
+        @keyframes shake {
+            10%, 90% {
```
sheenarbw: also the same across PRs

backend/frontend/views.py
```diff
@@ -295,6 +298,75 @@ def action_start_card(request, card_id):
...
 
 
+def user_can_request_review(logged_in_user):
+    request = get_current_request()
+    card_id = request.resolver_match.kwargs.get("card_id")
+
+    card = get_object_or_404(AgileCard, pk=card_id)
+    card_assignees = card.assignees.all()
+
+    if logged_in_user in card_assignees:
```
sheenarbw: This functionality seems to be repeated in AgileCard.request_user_can_request_review

sheenarbw (changes requested):
It might make sense for you or Ngoako to pause while we get one thing merged. There is a lot of stuff that is repeated across PRs and it's going to get messy to merge these things

kingraphaii:
@Ngoakor12 

As suggested in previous reviews, I'm proposing we use the check_[verb]_[optional_qualifier] for naming our permission checking functions.

Example:
```python
def check_user_can_request_review_on_card(logged_in_user):
    return [bool]

@user_passes_test_or_forbidden(check_user_can_request_review_on_card)
def decorated_view():
    ...
```
I've used it on the "request_review" action only until we all agree on a naming style for them.

Sheena also mentioned how we do not make it known that some of these expect a "card_id" to be present in the `kwargs,` but I haven't been able to come up with a clean solution so far; feel free to suggest a better way to handle this.

backend/frontend/tests/test_card_request_review_button.py
```diff
@@ -0,0 +1,100 @@
...
+        self.page.click("text=Request Review")
+
+        self.page.wait_for_load_state("networkidle")
+
+        ip_column = self.page.text_content("div#column_IP")
+        review_column = self.page.text_content("div#column_IR")
+        card_title = self.card.content_item.title
+
+        self.assertIn(card_title, review_column)
+        self.assertNotIn(card_title, ip_column)
```
Ngoakor12, line 76: I think the functionality is missing a test for checking if the right action is logged as mentioned [here](https://github.com/Umuzi-org/Tilde/pull/679#discussion_r1477965261) on the other PR

Ngoakor12 (changes requested):
This implementation seems much cleaner to me :smile: :ok_hand: 


> As suggested in previous reviews, I'm proposing we use the check_[verb]_[optional_qualifier] for naming our permission checking functions.

I think the naming convention makes sense.

> I've used it on the "request_review" action only until we all agree on a naming style for them.

We just need to remember to create a separate task for changing the rest once this PR gets merged.

> Sheena also mentioned how we do not make it known that some of these expect a "card_id" to be present in the kwargs, but I haven't been able to come up with a clean solution so far; feel free to suggest a better way to handle this.

I can't think of a better way for making this apparent. I think the "on_card" sort of makes it known that we need a way to identify the card, same thing for team related checks.

backend/frontend/tests/test_card_request_review_button.py
kingraphaii: Missed that one, thanks!

Ngoakor12 (approved):

backend/frontend/templates/frontend/user/board/js_exec_action_card_moved.html
```diff
@@ -0,0 +1,5 @@
...
+  moveCardToCorrectColumn("{{card.id}}", "{{card.status}}");
+</script>
```
sheenarbw, line 5: There is a file called view_partial_action_card_moved. It's content is:

```
{% include "frontend/user/board/partial_card.html" %}

<script>
    moveCardToCorrectColumn("{{card.id}}","{{card.status}}")
</script>
```

It is exactly the same as this file.

Can you please make it DRY

backend/curriculum_tracking/models.py
```diff
@@ -1667,6 +1667,35 @@ def request_user_is_assignee(self):
 
         return self.assignees.first() == user
 
+    def request_user_can_request_review(self, user=None):
+        """
+        Check if current user can request review for this card
+        """
+        if self.status not in [AgileCard.IN_PROGRESS, AgileCard.REVIEW_FEEDBACK]:
```
sheenarbw, line 1677: would it not make sense to return false if this is not a project card? Because nobody can request a review on a topic?

like: 
```
if card.content_type_nice !== project:
    return False
```

This would simplify code elsewhere and make this function more explicit and cohesive.

backend/frontend/templates/frontend/user/board/partial_card.html
```diff
@@ -166,5 +166,17 @@
             >Start
         </button>
     {% endif %}
+    
+    {% if card.content_type_nice == "project" %}
+        {% if card.request_user_can_request_review %}
```
sheenarbw: Shouldn't `{% if card.request_user_can_request_review %}` be enough here? 
Ideally all the logic around figuring out if a user can request a review should be inside the `card.request_user_can_request_review` function.

Right now, most of the logic is inside the function and the rest is inside the template. It's a strange way to split things up

sheenarbw (changes requested):
Very good. This looks solid.

Just a few little issues of cleanliness.  You can send this straight back to me once you have fixed that stuff and I'll merge it.

backend/frontend/templates/frontend/user/board/js_exec_action_card_moved.html
kingraphaii: I have removed it, had been renamed to `js_exec_[]` so we'll keep the that one

sheenarbw (approved):

PR #692, opened 2024-01-24
Making the card component look good

Sbonelo01:
Related issues: https://umuzi-projects.monday.com/boards/1274815998/pulses/1363777379

## Description:

Making the card component look good

## Note:
I am aware that a user profile picture can be uploaded in the backend. But for now, I am using static pictures for the avatars.

## Screenshots/videos

Before:

image

After:

tilde-card-7

## I solemnly swear that:


Ngoakor12 (changes requested):
Implementation-wise, it looks very similar to the designs :ok_hand: 

- The card content doesn't seem to be centered. There is no breathing room on the right. The card details lean to the right leading to some of the card details being hidden. Is there no way to add breathing room and avoid hidden content? E.g. long project names?
image

- The content button is not centered. I'm guessing that's because you wanted to add an icon there, but if there isn't one available, I think the text should be centered.
image

- The whole button should be clickable, not just the text

- The due date shows in a funny way
image
- The tooltips are not there anymore. Maybe they should be in some places? E.g. three dot icon, avatars


General questions:
- What does the date at the top represent? It's not obvious and seems confusing maybe it shouldn't be there?
image

- What does this icon represent? What's meant to happen when it's clicked?
image

backend/frontend/templates/frontend/user/board/partial_card.html
```diff
@@ -1,170 +1,223 @@
 {% load humanize %}
 
+<!-- TODO: Extract constants and store them theme.py -->
```
Ngoakor12: should this be here?

Ngoakor12 (changes requested):
image
- The review count icons are bunched up together, I'm not sure if that's intended or not.
- Also, the competent review count does not have an icon for some reason. 
- And these icons need tooltips, otherwise it's hard to make sense of them.

image
- This PR section needs a tooltip too


Side note:
- Regarding the tags. I understand you felt they were extra busy with the icon and description inside a chip, but I'd argue we still have the same thing with the hashtags in place of the icons. Personally I think the Tilde 2 original(tag inside a chip without an icon or hashtag) looks better and is less busy.

Sbonelo01:
@Ngoakor12 I've addressed your feedback
The tooltip for a bunch of reviewers doesn't make that much sense right now. I have an idea of what needs to be done. (Here is an issue on MoM: https://umuzi-projects.monday.com/boards/1274815998/pulses/1387241235)

Ngoakor12:
> * The card content doesn't seem to be centered. There is no breathing room on the right. The card details lean to the right leading to some of the card details being hidden. Is there no way to add breathing room and avoid hidden content? E.g. long project names?
>   image

This point :point_up:  seems to have been undone or not fully resolved.

Screenshot from the last review:
image

Screenshot now:
image

Ngoakor12:
Nevermind, it seems to be a responsiveness/out of date chrome on linux browser issue

Ngoakor12 (approved):

backend/frontend/templates/frontend/user/board/partial_card.html
```diff
@@ -1,170 +1,225 @@
...
+             {% endif %}
+             {% if card.code_review_red_flag_since_last_review_request %}
+                <div class="relative inline-flex items-center p-2 text-sm font-medium text-center bg-opacity-10 rounded-lg">
+                  <i class="fa-regular fa-face-sad-tear" aria-hidden="true" title="Red-flag reviews"></i>
+                  <div class="absolute inline-flex items-center justify-center w-5 h-5 text-xs text-white bg-red-600 border-2 border-white rounded-full -top-2 -end-2">{{ card.code_review_red_flag_since_last_review_request }}</div>
+                </div>   
+             {% endif %}
+             {% if card.code_review_excellent_since_last_review_request %}
+                <div class="relative inline-flex items-center p-2 text-sm font-medium text-center bg-opacity-10 rounded-lg">
+                  <i class="fa-regular fa-face-laugh-wink" aria-hidden="true" title="Excellent reviews"></i>
```
sheenarbw: This is looking very repetitive. Can you try make it more DRY?

This could help: https://github.com/carltongibson/django-template-partials

backend/frontend/templates/frontend/user/board/partial_card.html
```diff
@@ -1,170 +1,225 @@
...
+                  <div class="absolute inline-flex items-center justify-center w-5 h-5 text-xs text-white bg-red-600 border-2 border-white rounded-full -top-2 -end-2">{{ card.code_review_excellent_since_last_review_request }}</div>
+                </div>
+             {% endif %}
+         </div>
+
+        {% if card.open_pr_count %}
+         <a href={{ card.repo_url }} title="Pull request info">
```
sheenarbw: There was a request to be able to navigate straight to the repo from the card even if there are no open PRs. This request was made with Tilde1, might as well implement it now.

backend/frontend/templates/frontend/user/board/partial_card.html
```diff
@@ -1,170 +1,225 @@
...
-        <button type="button" class="{{ styles.button_primary_small }}"
+    <div class="grid grid-cols-2 grid-rows-1 gap-x-10 gap-y-3 w-full h-[auto] pt-2">
+
+      <div class="Assignee-avatar col-span-1 row-span-1 group inline-block">
+        <div class="mb-2 text-xs text-indigo-950 font-['Poppins']">Assignee</div>
+        {% if card.assignee_names %}
+          {% for assignee in card.assignee_names %}
+            <img
+              class="w-[30px] h-[30px] rounded-full"
+              src="https://images.unsplash.com/photo-1594751543129-6701ad444259?q=80&w=1000&auto=format&fit=crop&ixlib=rb-4.0.3&ixid=M3wxMjA3fDB8MHxzZWFyY2h8Mnx8ZGFyayUyMHByb2ZpbGV8ZW58MHx8MHx8fDA%3D"
```
sheenarbw: for now can you please make this an anonomous image/icon. We are unlikely to implement profile pics before we go live.

backend/frontend/templates/frontend/user/board/partial_card.html
```diff
@@ -1,170 +1,225 @@
...
+                {% endif %}
+            {% endif %}
+        </div>
+      </div>      
+    
+      <!-- Buttons section -->
+      
+      <div class="grid grid-cols-2 grid-auto-rows-1fr gap-x-10 gap-y-10 w-full h-[auto]">
+        <div class="Start-button col-span-1 row-span-1 w-full">
+          <button
```
sheenarbw: Clicking on the content button should not start the card. It should rather link to the card content and open it in a new tab.   

Alternatively,  `alert("TODO")` so it's very obvious that it needs to be fixed, and you can also make a Mom task so it can be done

Please don't make buttons do random things!

backend/frontend/theme.py
```diff
@@ -1,10 +1,12 @@
...
+CONTENT_BTN_COLOUR = "indigo-950"
+
 TEXT_COLOUR = "gray"
 
 styles = {
+    # button
     "button_primary_small": f"rounded bg-{PRIMARY_COLOUR}-600 px-2 py-1 text-xs font-semibold text-white shadow-sm hover:bg-{PRIMARY_COLOUR}-500 focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-{PRIMARY_COLOUR}-600",
-    #
-    #
+    "button_content_small": f"rounded bg-{CONTENT_BTN_COLOUR} px-2 py-1 text-xs font-semibold text-white shadow-sm hover:bg-{CONTENT_BTN_COLOUR} focus-visible:outline focus-visible:outline-2 focus-visible:outline-offset-2 focus-visible:outline-{CONTENT_BTN_COLOUR}",
```
sheenarbw: Is this only used in one place? If yes, it does not belong in this file.

This file is about creating things we will use over and over again.  We can have many small primary buttons on many pages all over the place.  But content buttons are very rare.

sheenarbw: Why not just make use of primary and secondary buttons? You can update the definitions. We don't want a new button style defined for every function that a button might do. We should reuse things.

backend/frontend/theme.py
```diff
@@ -1,10 +1,12 @@
 PRIMARY_COLOUR = "indigo"
+CONTENT_BTN_COLOUR = "indigo-950"
```
sheenarbw: Please aim to be consistent with the things that are already in place. If you are defining a new standard it's good to be explicit about it. This just seems a bit random. This one colour constant will behave differently from all the rest.

I'm also not convinced it's necessary.

backend/frontend/templates/frontend/user/board/partial_card.html
sheenarbw: If it seems like a difficult thing to do or it will cause too many delays, please add a new Mom task for it. You can label it as technical debt. We really should try hard not to write repetitive code.

sheenarbw (changes requested):
This is looking good but there are a few rough edges.

Also, please always run the unit tests, even if you are doing pure frontend work. This is a big piece of code and there is a chance it'll break something.

backend/frontend/templates/frontend/user/board/partial_card.html
Sbonelo01: Technical debt issue added: https://umuzi-projects.monday.com/boards/1274815998/pulses/1398535402

backend/frontend/templates/frontend/user/board/partial_card.html
Ngoakor12: @Sbonelo01 The current workaround for this issue only works when there are open PRs. The previous card had a "Repository" text to achieve this. I don't know what needs to happen because the card design you are following does not consider this.

I think a GitHub icon that links to the repo if it exist should be fine for now, maybe there's a better way to do it.

backend/frontend/templates/frontend/user/board/partial_card.html
```diff
@@ -1,180 +1,201 @@
...
+               title="Excellent reviews"></i>
+            <div class="absolute inline-flex items-center justify-center w-5 h-5 text-xs text-white bg-red-600 border-2 border-white rounded-full -top-2 -end-2">
+              {{ card.code_review_excellent_since_last_review_request }}
+            </div>
+          </div>
+        {% endif %}
+      </div>
+      {% if card.open_pr_count %}
+        {% with modified_url=card.repo_url|slice:"25:-4" %}
+          <a href="https://github.com/umuzi-org/{{ modified_url }}/pulls"
```
Ngoakor12: I think it's best if we reuse the `get_github_repo_link` method that was used before and just append "/pulls" at the end and avoid hardcoding the URL here.

Ngoakor12 (changes requested):

Ngoakor12 (approved):

sheenarbw (approved):

PR #693, opened 2024-01-25
authentication page layout -- make sure the pages are laid out in a tidy and professional way

mfanaKoPitori:

## Description:

## Basic Login Form
Screenshot 2024-01-30 113140

## Login Page Empty Email and Password Field Errors
Screenshot 2024-01-30 113154

## Login Page Invalid Email Address Field Error
Screenshot 2024-01-30 113221

##  Login Page Incorrect Email Address or Password Non Field Error
Screenshot 2024-01-30 122732

##  Basic Forgot Password Form
Screenshot 2024-01-30 114339

## Forgot Password Page Empty Email Address Field Error
Screenshot 2024-01-30 113302

## Forgot Password Page Invalid Email Address Field Error
Screenshot 2024-01-30 113315

## Forgot Password Page Success Page
Screenshot 2024-01-30 113332

## Reset Password Basic Page
Screenshot 2024-01-30 191037

## Reset Password Page Empty Password and Confirm Password Field Errors
Screenshot 2024-01-30 191053

## Reset Password Page Mismatching Password and Confirm Password Field Errors
Screenshot 2024-01-30 191129

## Reset Password Page Redirects to Login Page with Alert Message
Screenshot 2024-01-30 191156

## I solemnly swear that:


Ngoakor12 (changes requested):
Well done on getting this far Raymond...since this is a design-heavy task, some or most of the feedback will be subjective, but I'll try my best to justify it and provide some resources.

- Field errors should be displayed like field errors. Adding an icon followed by red coloured text should be enough. Having a border and background colour for inline errors does not look right to me, I think that should be left to the main page error, maybe everything is correct but the form failed to process the info or something unexpected happened.
- I think input fields look better when they span full-width of the container especially if the container is a fixed and small size box like we have.
- Regarding the password reset failed page, I think whoever is entering an email should still be given a chance to correct their email rather than the whole thing just disappearing So you don't need to create a whole new page, just display the error inline or on the same form. I think the done page should have a link to go back to the login page

Have a look at the following regarding all that I've said and more:
- https://www.nngroup.com/articles/errors-forms-design-guidelines/
- https://adamsilver.io/blog/where-to-put-buttons-on-forms/

You'll let me know if there's anything you're not sure or disagree with.

Ngoakor12 (changes requested):
Much improved Raymond

- There is however no need for you to check that the user exists or if their password is correct. That logic is already handled. The field errors should be concerned with whether the details the user entered are valid or make sense. E.g. password can't be two characters, email has all the things required of an email. That should also be covered already, the main thing was that they were supposed to render correctly in the right places is how I understand the task.
- Also I think the errors being normal font-weight is fine, bold is just too in your face

Also have a look at why specifying if the password or username is wrong may not be the best thing [here](https://stackoverflow.com/questions/14922130/which-error-message-is-better-when-users-entered-a-wrong-password) and [here](https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#authentication-and-error-messages)

Ngoakor12 (approved):

Ngoakor12 (approved):

backend/frontend/templates/frontend/auth/page_login.html
```diff
@@ -1,12 +1,35 @@
...
     </div>
+
+    {% for field in form %}
+      <label for="{{ field.id_for_label }}" class="{{ styles.label_small }}">{{ field.label_tag }}</label>
+      {{ field }}
+      {% if field.errors %}
+          <ul class="{{ styles.field_error }}">
+              {% for error in field.errors %}
+                  <li>
+                      <i class="fa-solid fa-triangle-exclamation"></i>
```
sheenarbw: This is not DRY. And I'm sure that if we had to write this kind of thing out manually for every single form we build forever then that would turn into a big mess. We need to centralise functionality and use the tools at our disposal.

There are 2 options here:
1. Look here: https://docs.djangoproject.com/en/5.0/topics/forms/#reusable-form-templates OR
2. Upgrade the ThemedFormMixin so it does all the rendering

Currently it seems that setting up a form template is the way to go

sheenarbw (changes requested):
This isn't DRY at all.

Please make use of Django's built in machinary to write reusable code. We wont want to have to hard-code forms when we build new ones.

backend/frontend/templates/frontend/auth/base.html
```diff
@@ -1,14 +1,14 @@
...
-    <div class="mx-auto w-1/3 p-6 bg-white rounded">
+    <div class="mx-auto w-1/4 p-6 bg-white rounded my-4">
       <h1 class="{{ styles.heading1 }}">
         {% block auth_heading %}{% endblock %}
       </h1>
       <hr class="my-2" />
       {% if messages %}
         {% for message in messages %}
-          <div class="{{ message.extra_tags }}" role="alert">
+          <div class="{{styles.alert_success}}" role="alert">
```
kingraphaii: Why do we assume every message sent is a success one?

backend/frontend/templates/frontend/auth/page_login.html
```diff
@@ -1,12 +1,14 @@
 {% extends "frontend/auth/base.html" %}
 {% block auth_heading %}Login{% endblock %}
 {% block auth_content %}
-  <form action="" method="post">
+  <form action="" method="post" class="my-4" novalidate>
```
kingraphaii, line 4: Any reason why we are using "novalidate" here?

backend/frontend/templates/frontend/auth/form_template.html
```diff
@@ -0,0 +1,24 @@
+
+{% if form.non_field_errors %}
+    <ul class="my-4 {{styles.alert_error}}">
+        {{ form.non_field_errors.as_ul }}
+    </ul>
+{% endif %}
+
+{% for field in form %}
+    <div class="fieldWrapper">
```
kingraphaii: Where is this class used? "fieldWrapper"

kingraphaii (changes requested):

backend/frontend/templates/frontend/auth/page_login.html
mfanaKoPitori: @kingraphaii, this is what I am supposed to do:

```
The main challenge with this is that we are using Django forms functionality to render the forms so you'll need to get the backend to render things in a useful way.

Also please make sure that field errors get rendered correctly.
```
- If I let HTML do its default validation, then there is no point in me addressing the last point, because they are doing the same thing. Or am I missing something?

Ngoakor12 (approved):

sheenarbw (approved):

PR #695, opened 2024-01-31
Add page for link card details

mfanaKoPitori:

## Description:

## I solemnly swear that:


backend/frontend/templates/frontend/user/board/page_card_details.html
```diff
@@ -0,0 +1,117 @@
+{% extends "frontend/user/base.html" %}
+
+{% block main %}
+
+    {% if card.project_submission_type_nice == "link"%}
```
sheenarbw: If we totally delete the card then the project will still exist. We need to access projects that have no card associated with them.  

The word `card` should not be here. It should just be something like `if project.submission_type == link`

backend/frontend/templates/frontend/user/board/page_card_details.html
```diff
@@ -0,0 +1,117 @@
+{% extends "frontend/user/base.html" %}
```
sheenarbw: We don't want a page called card details. 

It is not a part of the board, it should not be inside the board directory. It should also not be inside the user directory.

If we totally delete the card then the project will still exist. We need to access projects that have no card associated with them.  Same for topics. So this is not a card.

We also don't always access these pages via the board. We can get to them from a code review dashboard or similar.

A better path would be something like:

`backend/frontend/templates/frontend/{something}`

Either we would need separate pages for topic and projects in which case we would have:

`backend/frontend/templates/frontend/page_project_details`

Or we would need to combine both of the topic and project pages into one template in which case this might make more sense:

`backend/frontend/templates/frontend/page_course_component_details` or something like that.

backend/frontend/theme.py
```diff
@@ -20,4 +20,6 @@
     "alert_warning": f"bg-gray-100 border border-gray-400 text-gray-700 px-4 py-3 rounded relative",
     "alert_success": f"bg-green-100 border border-green-400 text-green-700 px-4 py-3 rounded relative",
     "alert_error": f"bg-red-100 border border-red-400 text-red-700 px-4 py-3 rounded relative",
+    # background
+    "main_background":"bg-orange-300"
```
sheenarbw: This file is for things that will get used over and over. It's not for things that get used once. Think of it as utilities or like components and widgets. 

I'm also sure you meant to remove this line because it is not used anywhere at all.

backend/frontend/urls.py
```diff
@@ -39,6 +39,11 @@
         views.action_start_card,
         name="action_start_card",
     ),
+    path(
+        "cards/<int:card_id>/details",
```
sheenarbw: We need to navigate to projects even when they don't have cards. `projects/<int:card_id>` would be better. The `/details` at the end seems unnecessary.

backend/frontend/urls.py
```diff
@@ -39,6 +39,11 @@
...
+        views.link_card_details_page,
+        name="link_card_details_page",
```
sheenarbw: It's very likely that we'll use the same view for all project cards. So a better name would be "project_details_page".

If it is a link project then the template should show link stuff, and if it is a repo project, it should show repo stuff. We can handle this in the template or in the view.

backend/frontend/views.py
```diff
@@ -294,6 +294,24 @@ def action_start_card(request, card_id):
         },
     )
 
+@user_passes_test(is_super)
```
sheenarbw: Percy will be able to help you with the permissions. He wrote a bunch of stuff. 
We will need anyone with view permission to see the card details.

backend/frontend/views.py
```diff
@@ -294,6 +294,24 @@ def action_start_card(request, card_id):
...
+@csrf_exempt
+def link_card_details_page(request, card_id):
+    """The card is in the backlog and the user has chosen to start it"""
+    card = get_object_or_404(AgileCard, id=card_id)
```
sheenarbw: get the project

backend/frontend/views.py
```diff
@@ -294,6 +294,24 @@ def action_start_card(request, card_id):
...
+
+    recruit_project_reviews = get_list_or_404(RecruitProjectReview)
```
sheenarbw: only get the reviews that are associated with the current project. Getting ALL the reviews is not going to work out.

```
project.project_reviews.order_by("- timestamp")
```

And I believe that get_list_or_404 will only raise a 404 if there are literally zero reviews in the database. This is a misuse of get_list_or_404

backend/frontend/views.py
```diff
@@ -294,6 +294,24 @@ def action_start_card(request, card_id):
...
+    reviews = [{"timestamp":review.timestamp,"reviewer": get_object_or_404(User,id=review.reviewer_user_id).get_username(),"comments":review.comments,"status":review.status} for review in recruit_project_reviews if review.recruit_project_id == card.recruit_project_id]
```
sheenarbw: You are over-using the whole get...or_404 thing. 

We can just get a review user like this: `review.reviewer_user`. There is a constraint on the database so it's impossible for there to be a review without a reviewer user. Unless postgres or Django have massive bugs which is very unlikely.

Also, why are you doing this at all? You can just pass the reviews to the template. The context could be:
```
{
"project": project,
"reviews": project.project_reviews.order_by("- timestamp")
}
```

Then loop over the reviews inside the template in the normal way.

sheenarbw: Also, if a project has no reviews then we really should not raise a 404.  All projects start out with no reviews, and then people review them by accessing the project page.

If this page raises a 404 if there are no reviews on a project then nobody will be allowed to add the first review. Nobody will be allowed to see the link that was submitted even.

Please spend some time understanding what happens whien a 404 is raised on a page. It literally stops the whole page from working. It's a proper exception

sheenarbw (changes requested):
You're on the right track mostly. There is some non-obvious stuff worth knowing about. 

It looks like I forgot to mention that stuff on the initial task. Basically, the whole AgileCard mechanism needs to change at some point so we need to only rely on cards when we have to. Projects can exist without cards.

You also need to get a bit more comfortable with getting data out of models, and how to access related objects and that kind of thing.

Lastly, the 404 functions have specific use cases. You are doing very funny things there. They are for when a url points to a resource that doesn't exist. You are using them to make whole pages raise exceptions even when the data is doing perfectly normal things.

backend/frontend/views.py
```diff
@@ -294,6 +295,40 @@ def action_start_card(request, card_id):
         },
     )
 
+@user_passes_test_or_forbidden(can_view_team)
+@user_passes_test_or_forbidden(can_view_user_board)
+def project_details_page(request, project_id):
+    project = RecruitProject.objects.get(pk=project_id)
```
kingraphaii: Let's use [get_object_or_404](https://docs.djangoproject.com/en/5.0/topics/http/shortcuts/#get-object-or-404) instead of `.get()` to avoid 500 (server) error when the project does not exist.

backend/frontend/views.py
```diff
@@ -294,6 +295,40 @@ def action_start_card(request, card_id):
...
+
+    reviews = [{
+        "timestamp":review.timestamp, 
+        "status":  PROJECT_STATUS.get(review.status), 
+        "comments": review.comments, 
+        "reviewer_user":review.reviewer_user
+        } for review in project.project_reviews.order_by("-timestamp")
+    ]
+
+    form = SubmissionLinkForm()
```
kingraphaii: Do we always have the submission form on RecruitProjects? If not, let's have the form set up only when we need it.

backend/frontend/views.py
```diff
@@ -294,6 +295,40 @@ def action_start_card(request, card_id):
...
+    context = {
+        "form": form,
+        "project": project,
+        "reviews": reviews,
+        "board_status": BOARD_STATUS[project.agile_card_status],
+    }
+    if request.method == "POST":
+        submission_link = request.POST.get('submission_link')
+
+        if project.link_submission_is_valid(submission_link):
```
kingraphaii: Maybe add a `TODO` for when the link is invalid to show that there's some furth work we need to add later on here

backend/frontend/constants.py
```diff
@@ -0,0 +1,14 @@
...
+}
+
+BOARD_STATUS = {
+    "B": "Blocked",
+    "R": "Ready",
+    "IP": "In Progress",
+    "RF": "Review Feedback",
+    "R": "Review",
+    "C": "Complete"  
+}
```
kingraphaii: `PROJECT_STATUS[ES]` > `PROJECT_STATUS`, same goes for `BOARD_STATUS`, makes more sense for the context they are used in.

backend/frontend/templates/frontend/page_course_component_details/page_project_details.html
```diff
@@ -0,0 +1,110 @@
...
+                    Content URL: 
+                    <a class="{{styles.link}}" href="{{project.content_url}}">{{project.content_url}}</a>
+                </p>
+            </div>
+
+            <div class="grid bold p-2 border-2 rounded-md gap-2">
+                <h6>Link to your work</h6>
+                
+                <div>Current link submission: 
+                    <a class="{{styles.link}}" href="{{project.link_submission}}">
```
kingraphaii: If we have no link submission it results in `project/None` so maybe we can use the `|default` filter with a dummy href like `#`

or better yet, only show link when we know it's there

backend/frontend/views.py
kingraphaii: I understand we are only working on "link" projects but we will be reusing the same view for other project types as well (repo)

backend/frontend/views.py
```diff
@@ -294,6 +295,40 @@ def action_start_card(request, card_id):
         },
     )
 
+@user_passes_test_or_forbidden(can_view_team)
+@user_passes_test_or_forbidden(can_view_user_board)
```
kingraphaii: We can stick to one perm checker here; we've duplicated the logic if you check their definitions. So let's keep one, perhaps the `can_view_user_board`

backend/frontend/views.py
```diff
@@ -294,6 +295,40 @@ def action_start_card(request, card_id):
...
+    form = SubmissionLinkForm()
+
+    context = {
+        "form": form,
+        "project": project,
+        "reviews": reviews,
+        "board_status": BOARD_STATUS[project.agile_card_status],
+    }
+    if request.method == "POST":
+        submission_link = request.POST.get('submission_link')
```
kingraphaii: We can use Django forms all the way here. We've used it to render HTML and left it at that, but we have much more stuff we can do with it, like our validation logic and even the db logic.

backend/frontend/static/frontend/page_project_details.js
```diff
@@ -0,0 +1,11 @@
+function toggleIcon(){
```
kingraphaii: Since the file name this resides in is `page_project_details.js`, perhaps we can be more specific with this. `toggleIcon` doesn't say much about which icon it is.

kingraphaii: same as for our displayForm function below

backend/frontend/templates/frontend/page_course_component_details/page_project_details.html
kingraphaii: This may not be a priority for now, but it's good to know what sort of tools we have at our disposal

kingraphaii (changes requested):
We need to rename our template base for the details functionality. "page_" prefix is reserved for the actual "pages" not dirs.

backend/frontend/templates/frontend/page_course_component_details/base.html
```diff
@@ -0,0 +1,16 @@
...
+{% block main_content %}
+  {% block content %}
+    <div class="mx-auto w-1/2 p-6 bg-white rounded my-4">
+      <h1 class="{{ styles.heading1 }}">
+        {% block project_heading %}{% endblock %}
+      </h1>
+
+      {% block project_content %}{% endblock %}
+    </div>
+    <script src="{% static "frontend/page_project_details.js" %}"></script>
```
kingraphaii: lets extend the "head" and add our scripts in there. See how its done in `users/base.html`

backend/frontend/templates/frontend/page_course_component_details/page_project_details.html
mfanaKoPitori: @kingraphaii I've heard this type of comment before from sheena, which was mainly about leveraging the tools that django provides, in that particular instance it was about using django built in forms rather than manually creating forms using html.

What is comment in reference to? is it about the link_submission mechanism?

backend/frontend/templates/frontend/page_course_component_details/base.html
mfanaKoPitori: @kingraphaii when I do that, the link_submission stops working, am i missing something? I extended head in this commit: https://github.com/Umuzi-org/Tilde/pull/695/commits/66e5836b8218034bb94756746060b9f1c773dfce

backend/frontend/templates/frontend/page_course_component_details/page_project_details.html
kingraphaii: > What is this comment in reference to?

 To using template filters like `|default`

mfanaKoPitori: @kingraphaii oh I see, I understand. I thought it was in reference to another issue.

backend/frontend/templates/frontend/page_course_component_details/base.html
kingraphaii: Please don't forget to use `{{ block.super }}` to ensure we get whatever scripts and other assets are loaded in the inherited templates. Maybe that's where your problem is coming from, but I can't say exactly since I do not have complete information on what you did exactly. You can ping me on Slack when you work on it so we can see what's up together.

mfanaKoPitori: @kingraphaii  thanks for the help, the problems was that the name used to defined head in base.html, it is called `main_head ` and not `head` as i was referencing it on my side.

backend/frontend/constants.py
```diff
@@ -0,0 +1,14 @@
+PROJECT_STATUSES = {
+    "NYC":"Not Yet Competent",
```
sheenarbw: These constants are already defined elsewhere. Why are you defining them here as well? 

See: Tilde/backend/curriculum_tracking/constants.py

backend/frontend/constants.py
```diff
@@ -0,0 +1,14 @@
...
+    "C":"Complete",
+    "RF":"Red Flag"
+}
+
+BOARD_STATUSES = {
```
sheenarbw: Here is a code snippet from elsewhere. Please keep things DRY:

```
    STATUS_CHOICES = [
        (BLOCKED, "Blocked"),
        (READY, "Ready"),
        (IN_PROGRESS, "In Progress"),
        (REVIEW_FEEDBACK, "Review Feedback"),
        (IN_REVIEW, "Review"),
        (COMPLETE, "Complete"),
    ]
```

backend/frontend/static/frontend/page_project_details.js
```diff
@@ -0,0 +1,11 @@
+function toggleEditLinkSubmissionIcon(){
```
sheenarbw: Please use alpine.js for showing and hiding things.

backend/frontend/templates/frontend/page_course_component_details/base.html
```diff
@@ -0,0 +1,20 @@
+{% extends "frontend/base.html" %}
```
sheenarbw: Should we not be extending base_logged_in? Only logged in users should be able to see this.

backend/frontend/templates/frontend/page_course_component_details/base.html
```diff
@@ -0,0 +1,20 @@
...
+{% block main_head %}
+    {{ block.super }}
+    <script src="{% static 'frontend/page_project_details.js' %}"></script>
+{% endblock %}
+
+{% block main_content %}
+  {% block content %}
+    <div class="mx-auto w-1/2 p-6 bg-white rounded my-4">
+      <h1 class="{{ styles.heading1 }}">
+        {% block project_heading %}{% endblock %}
```
sheenarbw: Is this base template only going be used for projects? Or topics as well? 

If the plan is to use it for both, then it would be best not to say "project_heading" here. Same for "project_content". Topics are not projects so the name would be misleading

backend/frontend/templates/frontend/page_course_component_details/page_project_details.html
```diff
@@ -0,0 +1,108 @@
...
+                <div class="bold p-2 border-2 rounded-md ">
+                    <p class="bold">
+                        Status : {{board_status}}
+                    </p>
+                </div>
+            </div>
+        
+            <div class="bold p-2 border-2 rounded-md grid gap-2">
+                {% if user.email != project.recruit_user_names|first %}
+                    <div class="bold p-2 border-2 rounded-md ">
```
sheenarbw: You don't need to hide the assignee. This would probably confuse people

backend/frontend/templates/frontend/page_course_component_details/page_project_details.html
```diff
@@ -0,0 +1,108 @@
...
+                <div>Current link submission: 
+                    {% if project.link_submission == None %}
+                        No link submitted yet
+                    {% else %}
+                    <a class="{{styles.link}}" href="{{project.link_submission}}">
+                        {{project.link_submission}}
+                    </a>
+                    {% endif%}
+                </div>
+                <i id="edit_link_submission_icon" class="fa-solid fa-pen-to-square" onclick="displayLinkSubmissionForm()"></i>
```
sheenarbw: rather use alpine.js for showing and hiding

sheenarbw (changes requested):
On the right track, but please make it DRY. You also apparently don't know about the awesome power of alpine.js. 

I think it would be good to focus on just displaying the correct info for now. We can handle making link submissions work in another task.  There can also be a separate task for displaying reviews.  I'll make those tasks now.

It would also be good to add some tests. Just make sure that the frontend shows up and that information appears somewhere in the page. We are using Playwright for tests that involve the browser.

backend/frontend/tests/test_course_component_details_pages.py
```diff
@@ -0,0 +1,70 @@
...
+from django.utils import timezone
+
+
+class TestCourseComponent(FrontendTestMixin):
+    def setUp(self):
+        super().setUp()
+        self.user = UserFactory(
+            email="learner@umuzi.org",
+            is_staff=True,
+            is_superuser=True,  # TODO: remove this once "restricted access" PR is merged
```
kingraphaii: we no longer need the is_superuser entry, that PR was merged

backend/frontend/tests/test_course_component_details_pages.py
```diff
@@ -0,0 +1,70 @@
...
+
+        self.agile_card_for_link_project = factories.AgileCardFactory(
+            status=AgileCard.READY,
+            recruit_project=self.link_project,
+            content_item=self.content_item,
+            assignees=[self.user],
+        )
+        self.agile_card_for_link_project.start_topic()
+
+        self.link_project_url = f"{self.live_server_url}/project/{self.link_project.id}"
```
kingraphaii: lets use `reverse` to revolve our relative URLs like:

```python
 f"{self.live_server_url}/reverse('...')"
```

If our links change we wont need to change the tests too

backend/frontend/tests/test_course_component_details_pages.py
```diff
@@ -0,0 +1,70 @@
...
+        self.assertIn("learner@umuzi.org", body)
+        self.assertIn("In Progress", body)
+        self.assertIn("Feb. 12, 2024, 2:06 p.m.", body)
+        self.assertIn("Feb. 13, 2024, 2:06 p.m.", body)
+        self.assertIn("learner_r1@umuzi.org", body)
+        self.assertIn("learner_r2@umuzi.org", body)
+        self.assertIn(
+            "https://raw.githubusercontent.com/Umuzi-org/tech-department/master/content/projects/tdd/simple-calculator/part-465/_index.md",
+            body,
+        )
```
kingraphaii: This fails, why did we say its part-465 and not part-1

backend/frontend/templates/frontend/course_component_details/page.html
kingraphaii, line 1: Nav bar missing, we did not inherit our base templates correctly

backend/frontend/templates/frontend/course_component_details/base.html
kingraphaii, line 1: Is there anything this base adds that the base_logged_in does not offer

backend/frontend/templates/frontend/course_component_details/page.html
kingraphaii, line 1: Also, reviews are missing. Did you remove them? They used to show on last revision

kingraphaii (changes requested):

backend/frontend/templates/frontend/course_component_details/page.html
mfanaKoPitori: @kingraphaii The reviews and the form part of link_submission part were moved to another task, so the functionality will be added later. Here are the newly created mom tasks: [link_submission_form](https://umuzi-projects.monday.com/boards/1274815998/pulses/1391272478), [reviews](https://umuzi-projects.monday.com/boards/1274815998/pulses/1391273049)

backend/frontend/tests/test_course_component_details_pages.py
mfanaKoPitori: @kingraphaii  when I remove the superuser entry my tests start failing. am I missing something?

backend/frontend/templates/frontend/course_component_details/page.html
```diff
@@ -0,0 +1,92 @@
+{% extends "frontend/base_logged_in.html" %}
+
+{% block content %}
+    <div class="mx-auto w-1/2 p-6 bg-white my-4">
+        <h1 class="{{ styles.heading1 }}">
+            {% block heading %}Course Component Details{% endblock %}
+        </h1>
```
kingraphaii: In the in the inherited base template there's not "Heading" block to extend. Not sure what you meant to do here, otherwise it can just be a normal heading

backend/frontend/views.py
```diff
@@ -17,14 +17,14 @@
...
-from curriculum_tracking.models import AgileCard, ContentItem
+from curriculum_tracking.models import AgileCard, ContentItem, User, RecruitProject
 
 from taggit.models import Tag
 from guardian.core import ObjectPermissionChecker
 
 from threadlocal_middleware import get_current_request
 
-from .forms import ForgotPasswordForm, CustomAuthenticationForm, CustomSetPasswordForm
+from .forms import ForgotPasswordForm, CustomAuthenticationForm, CustomSetPasswordForm,SubmissionLinkForm
```
kingraphaii: spacing here

backend/frontend/tests/test_course_component_details_pages.py
```diff
@@ -0,0 +1,77 @@
...
+
+from django.utils import timezone
+
+from core.tests.factories import UserFactory
+from .frontend_test_mixin import FrontendTestMixin
+from curriculum_tracking.tests import factories
+from curriculum_tracking.models import AgileCard
+
+
+class TestCourseComponent(FrontendTestMixin):
```
kingraphaii: We nay need to rename tbis to show its about "link" projects

kingraphaii (changes requested):

kingraphaii (dismissed):

kingraphaii (approved):

backend/frontend/tests/test_course_component_details_pages.py
```diff
@@ -0,0 +1,68 @@
...
+        self.page.goto(self.link_project_url)
+
+        body = self.page.text_content("body")
+
+        self.assertIn("Course Component Details", body)
+        self.assertIn("foo.1@example.com", body)
+        self.assertIn("In Progress", body)
+        self.assertIn("Feb. 12, 2024, 2:06 p.m.", body)
+        self.assertIn("Feb. 13, 2024, 2:06 p.m.", body)
+        self.assertIn("foo.2@example.com", body)
```
kingraphaii: Lets not use static values for our users. we can attach them to self in setup

kingraphaii (changes requested):

kingraphaii (approved):

backend/frontend/tests/test_course_component_details_pages.py
```diff
@@ -0,0 +1,73 @@
...
+        self.content_item: ContentItem = ContentItemFactory(
+            content_type=ContentItem.PROJECT,
+            project_submission_type=project_submission_type,
+        )
+
+        self.learner_reviewer = UserFactory(
+            email="learner_reviewer@umuzi.org",
+        )
+
+        self.card = AgileCardFactory(
```
sheenarbw: This is a little bit of a bad habit.

You are declaring a variable inside a function in a class. That doesn't mean that it makes sense to make it into an instance variable. 

Use instance variables if you actually intend to expose the variable in this way. You don't use self.card anywhere else in this class. This seems to happen quite a lot.

The code is more explicit when you use self intentionally. If you use self then that should mean you want people to be able to access the thing

backend/frontend/templates/frontend/course_component_details/page.html
```diff
@@ -0,0 +1,92 @@
...
+                </div>
+
+                <div class="grid bold p-2 border-2 rounded-md gap-2">
+                    <h6>Link to your work</h6>
+                    
+                    <div>Current link submission: 
+                        {% if project.link_submission == None %}
+                            No link submitted yet
+                        {% else %}
+                        <a class="{{styles.link}}" href="{{course_component.link_submission}}">
```
sheenarbw: when linking to something external, can you get it to open in a new page? 

https://www.freecodecamp.org/news/how-to-use-html-to-open-link-in-new-tab/

sheenarbw (changes requested):
one tiny thing I want you to know about classes.

Once you have made the changes you can ask for another review from me and I'll merge it. Very close now

PR #696, opened 2024-02-01
column scroll down

TheMaja:
- scrollable column
- scrollable columns

## Description:

## Screenshots/videos
[screen-capture.webm](https://github.com/Umuzi-org/Tilde/assets/73611175/402a2255-0eaf-4149-bf16-df2fe649147c)

## I solemnly swear that:


Ngoakor12 (changes requested):
The columns scroll as expected :+1: 

This is how it looks on my machine with normal zoom:
image
- Ideally, the column should take the whole height, excluding the headers. I think the [Github projects page](https://github.com/Umuzi-org/Tilde/projects/2?query=is%3Aopen+sort%3Aupdated-desc) is a good example. Have a look and see if you can't reverse engineer its board's scroll look/functionality.
- I'm not sure about the use of plain CSS when we're also using tailwind, but I guess you can do it that way if the tailwind way is too messy

Ngoakor12 (changes requested):
https://github.com/Umuzi-org/Tilde/assets/54069197/89862ba1-7e1c-46c2-bcac-1a0f6f2608f1



- The the screenshot attached seems fine, but the actual functionality does not match. See the attached video. It might just be my browsers(I also tried firefox), so please attach a video if everything is fine on your end.

Also, since the code changes more than the scroll behavior here's more feedback:
- The columns/cards should have a max size so that the board still looks reasonably good for people with extremely large screens. For those with extremely small screens you can also have a min with for the columns/cards so that it doesn't shrink for forever, you can add a horizontal scroll once it reaches that point.

Just a reminder to use the github projects page as a reference and don' try to copy everything but the scroll behavior is top notch in my opinion.

Ngoakor12 (changes requested):
- The the screenshot attached seems fine, but the actual functionality does not match. See the attached video. It might just be my browsers(I also tried firefox), so please attach a video if everything is fine on your end.

Also, since the code changes more than the scroll behavior here's more feedback:
- The columns/cards should have a max size so that the board still looks reasonably good for people with extremely large screens. For those with extremely small screens you can also have a min with for the columns/cards so that it doesn't shrink for forever, you can add a horizontal scroll once it reaches that point.

Just a reminder to use the github projects page as a reference and don' try to copy everything but the scroll behavior is top notch in my opinion.

Ngoakor12 (changes requested):
Two things:

- The board looks broken on smaller screen sizes. It should ideally have a min-width and scroll horizontally if the min-width is reached. Have a look at the current Tilde and the Github projects page for reference



https://github.com/Umuzi-org/Tilde/assets/54069197/94cfb424-1daf-4278-acd0-c5b679096fe8



- I also don't think there any need for the board to scroll at all. I know you said you were recreating the current Tilde, but it looks like a bug to me, because it even hides cards at the end





https://github.com/Umuzi-org/Tilde/assets/54069197/0920b350-db57-4fc8-b4f7-98c5e93aa004

Ngoakor12 (approved):
Much better Themba, make sure to update the PR description with the right demo video/screenshots

sheenarbw:
It's good but there is a weird thing happening at the bottom of the screen. I made a video: https://www.youtube.com/watch?v=XkSNgNZ7TS0 

If you really can't get the bottom gap to behave itself then please make another task on Mom about it so we can fix it.

backend/frontend/templates/frontend/user/board/js_resize_scroll_page.html
```diff
@@ -0,0 +1,12 @@
+<script>
+  window.addEventListener("resize", function () {
+    var columns = document.querySelectorAll(".grid-cols-5 > div");
+    var windowHeight = window.innerHeight;
```
Ngoakor12: use the correct keywords

Ngoakor12 (commented):
I think this could probably be done via CSS only, but I don't have the exact answer so I think what works works for now.

- The only major issue now is that the horizontal scroll bar is not visible. It works, you can scroll horizontally in smaller screen with `shift+scroll` shortcut but I think that's bad UX, if a screen can scroll let's show the scroll bar.
[Screencast from 08-03-2024 10:01:35.webm](https://github.com/Umuzi-org/Tilde/assets/54069197/606e2ac3-59cf-41c3-afa7-4174df7c150f)


I believe the main task and changes have been addressed so maybe creating a ticket for the issue above is acceptable.

backend/frontend/templates/frontend/user/board/js_resize_scroll_page.html
```diff
@@ -0,0 +1,12 @@
...
+    const columns = document.querySelectorAll(".grid-cols-5 > div");
+    const windowHeight = window.innerHeight;
+
+    columns.forEach(function (column) {
+      column.style.height = windowHeight - 160 + "px";
+    });
+  });
+
+  window.dispatchEvent(new Event("resize"));
+</script>
```
Ngoakor12: forgot to mention that the function and event handling should be moved to the static folder, and maybe even one of the existing files in the static folder

Ngoakor12 (commented):
you didn't address nor say anything about the other feedback

TheMaja:
> you didn't address nor say anything about the other feedback

I will create a new ticket mom ticket for it

backend/frontend/templates/frontend/user/board/js_resize_scroll_page.html
TheMaja: > forgot to mention that the function and event handling should be moved to the static folder, and maybe even one of the existing files

existing files you mean the page.html  template ?

Ngoakor12 (approved):

PR #705, opened 2024-03-05
card component: Add stop button and make it work - topics and projects

Ngoakor12:
https://umuzi-projects.monday.com/boards/1274815998/pulses/1301320332
## Description:
- I added a new `stop_project` function to the Agilecard model
- I added another `request_user_can_stop_card` function for checking request user's permission
- I then added everything around making the button actually show up and work with accompanying tests
- Fixed some tests that were failing
- Small UI updates to the board with and fixed hidden card text 

## Screenshots/videos
image

image

## I solemnly swear that:

### DIFF
--- a/backend/curriculum_tracking/models.py
+++ b/backend/curriculum_tracking/models.py
@@ -1693,6 +1693,7 @@ class AgileCard(
 
         if user is None:
             from threadlocal_middleware import get_current_user
+
             user = get_current_user()
 
         if user is not None:
@@ -1724,6 +1725,7 @@ class AgileCard(
 
         if user is None:
             from threadlocal_middleware import get_current_user
+
             user = get_current_user()
 
         if user is not None:
@@ -1740,6 +1742,36 @@ class AgileCard(
             )
             return has_manage_cards_permission
         return False
+
+    def request_user_can_stop_card(self, user=None):
+        """
+        Check if current user can stop card
+        """
+        if self.status != AgileCard.IN_PROGRESS:
+            return False
+
+        if user is None:
+            from threadlocal_middleware import get_current_user
+
+            user = get_current_user()
+
+        if user is not None:
+            is_assignee = user in self.assignees.all()
+
+            if is_assignee:
+                return True
+
+            has_manage_cards_permission = any(
+                (
+                    user.has_perm(Team.PERMISSION_MANAGE_CARDS, team)
+                    for team in self.get_teams()
+                )
+            )
+
+            return has_manage_cards_permission
+
+        return False
+
 
 class BurndownSnapshot(models.Model):
     MIN_HOURS_BETWEEN_SNAPSHOTS = 4
--- a/backend/curriculum_tracking/models.py
+++ b/backend/curriculum_tracking/models.py
@@ -1713,6 +1713,35 @@ class AgileCard(
 
         return False
 
+    def request_user_can_stop_card(self, user=None):
+        """
+        Check if current user can stop card
+        """
+        if self.status != AgileCard.IN_PROGRESS:
+            return False
+
+        if user is None:
+            from threadlocal_middleware import get_current_user
+
+            user = get_current_user()
+
+        if user is not None:
+            is_assignee = user in self.assignees.all()
+
+            if is_assignee:
+                return True
+
+            has_manage_cards_permission = any(
+                (
+                    user.has_perm(Team.PERMISSION_MANAGE_CARDS, team)
+                    for team in self.get_teams()
+                )
+            )
+
+            return has_manage_cards_permission
+
+        return False
+
 
 class BurndownSnapshot(models.Model):
     MIN_HOURS_BETWEEN_SNAPSHOTS = 4
--- a/backend/curriculum_tracking/models.py
+++ b/backend/curriculum_tracking/models.py
@@ -1629,10 +1629,9 @@ class AgileCard(
         if not user:
             return False
 
-        return (
-            (self.request_user_is_assignee(user) and self.can_start())
-            or (self.user_has_permission(user, Team.PERMISSION_MANAGE_CARDS)
-            and (self.can_start() or self.can_force_start()))
+        return (self.request_user_is_assignee(user) and self.can_start()) or (
+            self.user_has_permission(user, Team.PERMISSION_MANAGE_CARDS)
+            and (self.can_start() or self.can_force_start())
         )
 
     def request_user_is_assignee(self, user):
--- a/backend/curriculum_tracking/models.py
+++ b/backend/curriculum_tracking/models.py
@@ -1622,35 +1622,18 @@ class AgileCard(
         This function is only used in template rendering and we are using threadlocals,
         that is why we need to avoid argument parameters unless they're for testing purposes
         """
-        if user is None:
-            from threadlocal_middleware import get_current_user
+        from threadlocal_middleware import get_current_user
 
-            user = get_current_user()
+        user = user or get_current_user()
 
-        if user is not None:
-            can_start = self.can_start()
-            is_assignee = self.assignees.first() == user
+        if not user:
+            return False
 
-            if is_assignee and can_start:
-                return True
-
-            is_superuser = user.is_superuser
-            can_force_start = self.can_force_start()
-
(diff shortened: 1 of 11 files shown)


backend/curriculum_tracking/models.py
```diff
@@ -1465,6 +1465,29 @@ def stop_topic(self):
         self.status = AgileCard.READY
         self.save()
 
+    def stop_project(self):
+        assert self.status == AgileCard.IN_PROGRESS
+        assert self.recruit_project != None, f"Project hasn't been started"
+        assert (
+            self.content_item.content_type == ContentItem.PROJECT
+        ), f"Expected content_type to be 'project', but got {self.content_item.content_type}"
```
kingraphaii, line 1472: Shouldn't we make this check first before the one it precedes?

backend/frontend/views.py
```diff
@@ -509,6 +509,47 @@ def action_finish_topic(request, card_id):
...
+    content_type = card.content_item.content_type
+
+    if content_type == ContentItem.TOPIC:
+        card.stop_topic()
+    elif content_type == ContentItem.PROJECT:
+        card.stop_project()
+    else:
+        raise NotImplementedError("Only topics and projects can be stopped")
+
+    log_creators.log_card_review_request_cancelled(card=card, actor_user=request.user)
```
kingraphaii: Arent we supposed to log using this instead:

"log_card_stopped"

backend/frontend/tests/test_card_stop_button.py
```diff
@@ -0,0 +1,103 @@
...
+        self.page.wait_for_load_state("networkidle")
+
+        self.assertEqual(LogEntry.objects.count(), 1)
+        entry = LogEntry.objects.first()
+
+        self.assertEqual(entry.actor_user, self.user)
+        self.assertEqual(entry.effected_user, self.card.assignees.first())
+        self.assertEqual(entry.object_1, self.card.recruit_project)
+        self.assertEqual(entry.object_2, None)
+        self.assertEqual(entry.event_type.name, creators.CARD_REVIEW_REQUEST_CANCELLED)
```
kingraphaii: If we're only accepting status == IP, do we ever need to concern ourselves with review stuff?

kingraphaii (changes requested):

Ngoakor12:
I'm a little unsure about the details button bug issue. It was either this or commenting it out to prevent issues

backend/frontend/tests/test_card_stop_button.py
```diff
@@ -0,0 +1,106 @@
...
+    def make_topic_card(self):
+        self.card = AgileCardFactory(
+            content_item=ContentItemFactory(content_type=ContentItem.TOPIC),
+            status=AgileCard.IN_PROGRESS,
+        )
+
+        self.card.assignees.set([self.user])
+
+        # update field needed for an in-progress topic without logging the start event
+        self.card._create_topic_progress_if_not_exists()
```
kingraphaii: What if we trigger a start and adjust our log-checking logic?

backend/frontend/tests/test_card_stop_button.py
```diff
@@ -0,0 +1,106 @@
...
+    def test_project_stop_button_logs_card_stopped_event(
+        self,
+    ):
+        self.make_project_card(ContentItem.LINK)
+
+        self.page.click("text=Stop")
+
+        self.page.wait_for_load_state("networkidle")
+
+        self.assertEqual(LogEntry.objects.count(), 1)
```
kingraphaii, line 69: If we trigger a start then we would expect 2 entries here. We can also filter by event type name and assert an expected count and continue the rest of checks

kingraphaii (changes requested):

backend/frontend/tests/test_card_stop_button.py
Ngoakor12: It turned out not to be an issue because the `self.card.start_topic()` method does not trigger the start event to be logged, only the relevant view(`action_start_card`) does that.

kingraphaii (approved):

backend/curriculum_tracking/models.py
```diff
@@ -1465,6 +1465,29 @@ def stop_topic(self):
...
+        assert self.status == AgileCard.IN_PROGRESS
+        assert (
+            self.content_item.content_type == ContentItem.PROJECT
+        ), f"Expected content_type to be 'project', but got {self.content_item.content_type}"
+        assert self.recruit_project != None, f"Project hasn't been started"
+
+        project_submission_type = self.content_item.project_submission_type
+
+        if project_submission_type == ContentItem.LINK:
+            self.recruit_project.link_submission = ""
```
sheenarbw: Why are you removing the link submission?

backend/curriculum_tracking/models.py
```diff
@@ -1465,6 +1465,29 @@ def stop_topic(self):
...
+        elif project_submission_type not in ContentItem.PROJECT_SUBMISSION_TYPES:
+            raise NotImplemented(
+                f"Can't stop project of with submission type: {project_submission_type}"
+            )
+
+        self.recruit_project.review_request_time = None
+        self.recruit_project.start_time = None
+        self.recruit_project.due_time = None
```
sheenarbw: So if I was a learner and I had a deadline, I could start a card then stop it again and the deadline would go away...  

Try to change as little as possible. The card just needs to end up in the correct column.

backend/curriculum_tracking/models.py
```diff
@@ -1465,6 +1465,29 @@ def stop_topic(self):
...
+        self.recruit_project.save()
+        self.status = AgileCard.READY
+        self.save()
```
sheenarbw, line 1480: instead of setting self.status = AgileCard.READY do this:

```
self.status = self.derive_status_from_project(project)
```

This is more robust because then all the status calcs happen in one place. If self.status is not READY after you call this function then we know that if we regen the learner's board, the card will end up in the wrong place.

backend/curriculum_tracking/models.py
```diff
@@ -1741,6 +1766,36 @@ def request_user_can_finish_topic(self, user=None):
...
+            user = get_current_user()
+
+        if user is not None:
+            is_assignee = user in self.assignees.all()
+
+            if is_assignee:
+                return True
+
+            has_manage_cards_permission = any(
+                (
```
sheenarbw: This looks familiar... 

We often check if the logged-in user is the assignee or has management permissions. I've seen this code many times in recent PRs. 

Can you think of a way to make this a bit more DRY? 

I'd suggest something like:

```
def request_user_can_stop_card(self, user=None):
    from threadlocal_middleware import get_current_user
    if self.status != AgileCard.IN_PROGRESS:
            return False
    user = user or get_current_user()
    if not user:
        return False
    return (
          self.user_is_assignee(user) or 
          self.user_has_permission(user,Team.PERMISSION_MANAGE_CARDS)
    )
```

backend/frontend/tests/test_card_stop_button.py
```diff
@@ -0,0 +1,104 @@
...
+        ip_column = self.page.text_content("div#column_IP")
+        project_card_title = self.card.content_item.title
+
+        self.assertNotIn(project_card_title, ip_column)
+        self.assertIn(project_card_title, backlog_column)
+
+    def test_project_stop_button_logs_card_stopped_event(
+        self,
+    ):
+        self.make_project_card(ContentItem.LINK)
```
sheenarbw, line 63: I didn't go over all these tests in detail because I'm a bit short on time. But I have a concern.

Have you noticed that the Playwright tests are really really slow to run? Ideally we should only use playwright to test things when there is no other way, most of the time we should be able to test our logic using standard unit tests. 

Try to do unit tests first and then if there is something you dont quite trust, then reach for playwright. So it shouldn't be playwright-first. It should be playwright-last. 

There's a thing called the "testing pyramid" that is useful to know about. Here's an article about it: https://martinfowler.com/articles/practical-test-pyramid.html . The article is Java flavoured but it does a good job of explaining things. The test pyramid is also very googlable.

sheenarbw (changes requested):

kingraphaii (approved):

sheenarbw (approved):

PR #706, opened 2024-03-06
Link project: allow user to update the submission link

mfanaKoPitori:
Related mom task: [Link project: Allow user to update the link](https://umuzi-projects.monday.com/boards/1274815998/pulses/1391272478)

## Description:

### Course Component Details: With no link submitted.
screenshot1

### Course Component Details: With the input field filled.
screenshot2

### Course Component Details: When the form is submitted and the link is updated successfully.
screenshot

### Course Component Details: When the form is submitted but there is a field error due to empty input.
screenshot4

### Course Component Details:  When the form is submitted but there is a field error due to invalid url.
screenshot5

## I solemnly swear that:

### DIFF
--- a/backend/frontend/templates/frontend/base.html
+++ b/backend/frontend/templates/frontend/base.html
@@ -23,6 +23,11 @@
       src="https://kit.fontawesome.com/6e84d8828f.js"
       crossorigin="anonymous"
     ></script>
+    <script 
+      defer 
+      src="https://cdn.jsdelivr.net/npm/alpinejs@3.x.x/dist/cdn.min.js">
+    </script>
+
 
     {% block main_head %}{% endblock %}
   </head>
--- a/backend/frontend/templates/frontend/base.html
+++ b/backend/frontend/templates/frontend/base.html
@@ -23,11 +23,6 @@
       src="https://kit.fontawesome.com/6e84d8828f.js"
       crossorigin="anonymous"
     ></script>
-    <script 
-      defer 
-      src="https://cdn.jsdelivr.net/npm/alpinejs@3.x.x/dist/cdn.min.js">
-    </script>
-
 
     {% block main_head %}{% endblock %}
   </head>
--- a/backend/frontend/templates/frontend/base.html
+++ b/backend/frontend/templates/frontend/base.html
@@ -31,7 +31,7 @@
 
     {% block main_head %}{% endblock %}
   </head>
-  <body class="{% block body_classes %}h-full{% endblock %}">
+  <body class="{% block body_classes %}h-full{% endblock %} overflow-hidden">
     {% block main_content %} {% endblock %}
   </body>
 </html>
--- a/backend/frontend/tests/test_course_component_details_pages.py
+++ b/backend/frontend/tests/test_course_component_details_pages.py
@@ -94,6 +94,25 @@ class TestLinkProjectDetailsPage(FrontendTestMixin):
         body = self.page.text_content("body")
         self.assertIn("https://google.com", body)
 
+    def test_link_submission_form_disappears_after_successful_submission(
+        self,
+    ):
+        self.make_ip_project_card(ContentItem.LINK)
+
+        self.link_project_url = self.reverse_url(
+            "course_component_details", kwargs={"project_id": self.recruit_project.id}
+        )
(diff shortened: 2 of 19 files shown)


Ngoakor12 (changes requested):
Functionality seems good...

- The edit icon needs a tool-tip because it won't be obvious to everyone or maybe just use a normal text button
- Clicking the open space on the right side of the button should not click the button
- The page should be responsive in smaller screens:
image


- Another stylistic change I think should be made is replacing the "Course Component Details" text with the actual title of the project and removing it from the field below so it doesn't come up twice 
image

backend/frontend/tests/test_course_component_details_pages.py
```diff
@@ -69,7 +69,58 @@ def test_link_project_page_displays_correct_details(self):
         self.assertIn("Feb. 13, 2024, 2:06 p.m.", body)
         self.assertIn("learner_reviewer@umuzi.org", body)
         self.assertIn(
-            "https://raw.githubusercontent.com/Umuzi-org/tech-department/master/content/projects/tdd/simple-calculator/part-1/_index.md",
+            self.recruit_project.content_url,
             body,
         )
         self.assertIn("No link submitted yet", body)
+
+    def test_link_submission_form_correctly_updates_link_submission_after_link_update(
```
Ngoakor12: change `test_link_submission_form_correctly_updates_link_submission_after_link_update` to `test_link_submission_form_correctly_updates_link_submission` or make it less confusing

backend/frontend/tests/test_course_component_details_pages.py
```diff
@@ -69,7 +69,58 @@ def test_link_project_page_displays_correct_details(self):
...
+        self,
+    ):
+        self.make_ip_project_card(ContentItem.LINK)
+
+        self.link_project_url = self.reverse_url(
+            "course_component_details", kwargs={"project_id": self.recruit_project.id}
+        )
+        self.page.goto(self.link_project_url)
+
+        body = self.page.text_content("body")
```
Ngoakor12, line 87: You may also want to preemptively wait for everything to load before selecting any text_content because it tends to cause failing tests if something is selected while things haven't fully loaded. You can look up examples in other tests or maybe there's a better way

Ngoakor12 (changes requested):

Ngoakor12 (changes requested):
Mostly good, just make sure the edit button is also responsive and does not continue to shrink in the way it's doing.

image


Adding a `max-w-max` instead of `w-1/6` resolves the issue

Ngoakor12 (approved):

backend/frontend/templates/frontend/base.html
```diff
@@ -23,6 +23,11 @@
       src="https://kit.fontawesome.com/6e84d8828f.js"
       crossorigin="anonymous"
     ></script>
+    <script
```
sheenarbw: This is interesting. Don't we also include alpine.js in other templates? 

So now those templates that inherit from this one will include alpinejs twice. 

Please keep it DRY.

sheenarbw (changes requested):
Please keep script tags dry. 

Look at the duplicated script tags in all logged in pages.  If we inherit from a template, and a thing is defined in that template, we dont need to define it again.

Ngoakor12 (approved):

sheenarbw (commented):
This seems a bit overly complicated in a few ways. Implementing and testing the hiding and showing of forms was complex and unnecessary. I'm not sure that the UX makes sense - I mean, why have the extra clicks? The showing and hiding doesn't seem to add much value. 

The form is literally one field and one button. It's pretty small. It's not a big, bulky thing taking up all the space and making the page all cluttered. Replacing it with a big giant button and making the user click extra is a bit of a funny thing to do. 

If there is a link filled in: 
- Just display the link
- also display a pre-filled form for changing the link 

if there is no link filled in:
- display a pre-filled form for setting the link

There is a thing in coding called KISS. It's worth looking up.

Can you please simplify this. 

The one bit of fanciness that probably does make sense would be to use HTMX to submit the forms and then just swap the form out in place instead of reloading the entire page. But I would be ok with deferring that

sheenarbw (approved):

PR #707, opened 2024-03-11
Fix: Intermittent Test Failures

kingraphaii:
Related issues:

A couple of tests fail after exhausting the default 30-second timeout.

By temporarily introducing a delay before each request/action:

```python
class FrontendTestMixin(StaticLiveServerTestCase):
    def setUp(self):
        super().setUp()

        self.context = self.browser.new_context();
        def delay_request(route, request):
            time.sleep(4)
            route.continue_()
        
        self.context.route("", delay_request)
        
        self.page = self.context.new_page()

```

I came to this conclusion:

On the board page(s), there are a couple of external resources that are loaded (font-awesome, google-fonts, htmx cdn, tailwind cdn) and may delay painting for some elements, so depending on the tilde server's network connection or the external platforms' servers, things like `self.page.click("text=Start")` may timeout.

## Description:

This PR removes the timeout constraint by setting it to `0` for all problematic statements, and also increases the card alert display time to `10s`.

## I solemnly swear that:


Ngoakor12 (approved):

backend/frontend/tests/frontend_test_mixin.py
```diff
@@ -8,6 +8,7 @@
 """
 
 import os
+import time
```
sheenarbw: Why the import?

backend/frontend/tests/test_card_cancel_review_request_button.py
```diff
@@ -32,7 +32,7 @@ def make_project_card(self, project_submission_type):
     def test_cancel_review_request_button_moves_project_card_to_ip_column(self):
         self.make_project_card(ContentItem.LINK)
 
-        self.page.click("text=Cancel review request")
+        self.page.click("text=Cancel review request", timeout=0)
```
sheenarbw: This is a bit sketchy. Timeouts exist for a reason. What if one of our things just seriously doesn't work, and we get something that just loads forever? 

Timeouts save lives. 

This is a hack that would make the tests pass. But Django has a way of dealing with static resources. 

Can we not just do things the Django way?

sheenarbw (changes requested):
This looks like a bit of a hack. Can you think of reasons a timeout would be useful? They sure do get built into a lot of things.

Django has ways of serving static resources. Can we leverage that machinery to serve the problematic static files?

How about using a tool like https://github.com/kevin1024/django-npm?

backend/frontend/tests/test_card_start_button.py
```diff
@@ -56,7 +56,7 @@ def test_start_button_moves_project_card_to_ip_column(self):
```
Ngoakor12, line 55: I got this failed test and a other similar failures after a few runs. Maybe it's related to not using playwright `expect` like you added in other buttons? I also came across [this issue](https://stackoverflow.com/questions/77259434/python-testing-assertions-unittest-playwright) which seems related and encourages the use of `expect`

[More on expect](https://playwright.dev/python/docs/writing-tests#first-test:~:text=Playwright%20includes%20assertions%20that%20will%20wait%20until%20the%20expected%20condition%20is%20met.%20Using%20these%20assertions%20allows%20making%20the%20tests%20non%2Dflaky%20and%20resilient) - I'm reading in between the lines at this point, but `expect` might be the way to go?

```sh
FAIL: test_start_button_moves_topic_card_to_ip_column (frontend.tests.test_card_start_button.TestCardStartButton)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/frontend/tests/test_card_start_button.py", line 51, in test_start_button_moves_topic_card_to_ip_column
    self.assertIn(topic_card_title, ip_column)
AssertionError: 'something awesome part 31' not found in ''

Ran 38 tests in 132.532s

FAILED (failures=1)
```

Ngoakor12 (changes requested):

backend/frontend/tests/test_card_start_button.py
kingraphaii: I have the same sentiments; I just wanted different eyes on this. Thanks!

Ngoakor12 (changes requested):
- I'm still getting some failures related to `test_component_details_pages.py`, I had to add them to a file because I couldn't see the whole output: [output.txt](https://github.com/Umuzi-org/Tilde/files/14913908/output.txt)

Ngoakor12 (approved):

Ngoakor12 (approved):
Frontend tests passing

sheenarbw (approved):

PR #709, opened 2024-03-14
Topic shows correct info

mfanaKoPitori:
Related issues: Mom task: [Topic shows correct info](https://umuzi-projects.monday.com/boards/1274815998/pulses/1304803615)

## Description:

## Screenshots/videos

screenshot

## I solemnly swear that:


backend/frontend/views.py
```diff
@@ -368,16 +377,63 @@ def action_start_card(request, card_id):
 
 
 @user_passes_test_or_forbidden(can_view_user_board)
-def course_component_details(request, project_id):
-    project = get_object_or_404(RecruitProject, id=project_id)
+def topic(request, id):
```
Ngoakor12: maybe `topic_course_component_details` instead? same with the project view

backend/frontend/views.py
```diff
@@ -368,16 +377,63 @@ def action_start_card(request, card_id):
...
+    if relevant_logs:
+        duration = timezone.now() - relevant_logs[-1]
+        seconds = duration.total_seconds()
+        days, remainder = divmod(seconds, 86400)
+        hours, remainder = divmod(remainder, 3600)
+        minutes, remainder = divmod(remainder, 60)
+
+        formatted_time_difference = (
+            f"{int(days)} days, {int(hours)} hours, {int(minutes)} minutes"
+        )
```
Ngoakor12: Correct me if I'm wrong, but I think most of this duration stuff is related to [this ticket](https://umuzi-projects.monday.com/boards/1274815998/pulses/1407162151)?

So it should probably be moved to that model or just skip the duration field but create a separate task so that it can be added back. I think moving it is better because you've already done the bulk of the work

Ngoakor12 (changes requested):
- Slight responsiveness issues. Some text may need to wrap to a new line
image

- I'm wondering if we even need to have separate view functions for topics or projects since we are using one template file. Maybe we can use an if check. If whatever card we are dealing with is a topic: do topic-related stuff otherwise do project-related stuff and at the end do whatever the topics have in common with the projects.

backend/frontend/views.py
mfanaKoPitori: @Ngoakor12, I agree with moving this to the TopicProgress model, but this will have to be a different property method from the one you referenced. Our one calculates how long the card has been sitting in the current tilde column, while the one you referenced calculates the time it took for the card to move from start to completion.

Ngoakor12: Ah okay, my bad. I see it in the task description now.

backend/frontend/views.py
```diff
@@ -368,16 +377,63 @@ def action_start_card(request, card_id):
...
+    for log_entry in log_entries:
+        event_type = EventType.objects.get(id=log_entry.event_type_id)
+
+        if (
+            course_component.agile_card.status == AgileCard.IN_PROGRESS
+            and event_type.name == CARD_STARTED
+        ) or (
+            course_component.agile_card.status == AgileCard.IN_PROGRESS
+            and event_type.name == CARD_REVIEW_REQUEST_CANCELLED
+        ):
```
Ngoakor12: More things...

- This if check should be rewritten so that we don't have to the following twice:
```
course_component.agile_card.status == AgileCard.IN_PROGRESS
```

- I feel like there should be a way to just filter LogEntries without needing to use if checks or event types. You just need eventtypes that match `card started` and `card review request cancelled`. My brain might just be tired so maybe not, just think about it or have a closer look.

mfanaKoPitori: @Ngoakor12, here you mean something like the code in the image below, right?

Screenshot 2024-03-27 084436

Ngoakor12: Yeah, but maybe let's leave the sorting out for now and maybe combine some things. See:
```py
relevant_logs = LogEntry.objects.filter((Q(event_type__name=CARD_STARTED)|Q(event_type__name=CARD_REVIEW_REQUEST_CANCELLED)),object_1_id=self.id)

# then the sorting or just sort using while doing the query if possible
```

Ngoakor12: I'm just freestyling, so please make sure whatever you come up with works and feel free to use any other approaches if you feel this is unreasonable

mfanaKoPitori: @Ngoakor12, this is now done. ✅. I will move the log to the dedicated property method and make the rest of the changes. Thanks.

backend/frontend/templates/frontend/course_component_details/page.html
```diff
@@ -1,15 +1,72 @@
...
         </h1>
 
-        {% if course_component.submission_type_nice == "link"%}
-            <div class="grid gap-4 my-4">
+        {% if course_component.content_item.content_type_nice == "topic"%}
+            <div class="grid gap-4 my-4 w-full">
                 <div class="p-2 border-2 rounded-md grid gap-2">
-                    <div class="bold p-2 border-2 rounded-md ">
+                    <div class="p-2 border-2 rounded-md ">
+                        Tag names:
```
Ngoakor12: Some shared fields like tags, status and assignees,etc. seem to be repeated. Is there no way to only write them once?

backend/frontend/urls.py
```diff
@@ -68,7 +68,7 @@
     ),
     # Course component details
     path(
-        "project/<int:project_id>",
+        "course_component_details/<int:id>/<str:type>",
```
Ngoakor12: The URLs can be duplicated. So have the `/project/id` and `/topic/id` back, and refer them to the same view. Then in the view, you can check if the request URL is from a topic or project and fetch the necessary id.

I think it should be possible.

Duplication might not be a bad idea here. I'm just not confident which places it should and shouldn't be the way to go at the moment. But this seems easier and more approachable than adding another `<str:type>`

backend/curriculum_tracking/models.py
```diff
@@ -941,6 +947,24 @@ def get_activity_log_summary_data(self):
             "flavour_names": self.content_item.flavour_names,
         }
 
+    @property
+    def duration_in_current_column(self):
```
Ngoakor12: `duration_in_current_column` -> `get_card_duration_in_current_column`?

backend/curriculum_tracking/helpers.py
```diff
@@ -182,3 +182,16 @@ def agile_card_reviews_outstanding(user):
 def pull_request_reviews_outstanding(user):
     # TODO: only implement this once the github webhook is healthier
     return []
+
+
+def get_formatted_duration_to_string(duration):
```
Ngoakor12: `get_formatted_duration_to_string` -> `get_formatted_duration_string`

Ngoakor12 (changes requested):
The details button text seems weirdly cut off. Also, let's style it as a secondary button

Ngoakor12:
Mostly good, but the cards in the backlog column don't load now. I get some error in the terminal and commenting out the details buttons seems to resolve the cards not appearing problem.

image

mfanaKoPitori:
> Mostly good, but the cards in the backlog column don't load now. I get some error in the terminal and commenting out the details buttons seems to resolve the cards not appearing problem.
> 
> image

Morning there, @Ngoakor12. Could you please share the error message? I am failing to reproduce the bug on my side.

Ngoakor12:
> > Mostly good, but the cards in the backlog column don't load now. I get some error in the terminal and commenting out the details buttons seems to resolve the cards not appearing problem.
> > image
> 
> Morning there, @Ngoakor12. Could you please share the error message? I am failing to reproduce the bug on my side.

The server runs fine initially, but when I navigate to a user's board with cards in the backlog, I get:
```
[Tilde-backend] [2024-04-02 07:41:06] ERROR [django.request:224] Internal Server Error: /users/17/board/RB
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/core/handlers/base.py", line 179, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/contrib/auth/decorators.py", line 21, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/frontend/views.py", line 116, in _wrapped_view
    return view_func(request, *args, **kwargs)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/frontend/views.py", line 338, in view_partial_user_board_column
    return render(
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/shortcuts.py", line 19, in render
    content = loader.render_to_string(template_name, context, request, using=using)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/loader.py", line 62, in render_to_string
    return template.render(context, request)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/backends/django.py", line 61, in render
    return self.template.render(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/base.py", line 170, in render
    return self._render(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/defaulttags.py", line 211, in render
    nodelist.append(node.render_annotated(context))
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/loader_tags.py", line 192, in render
    return template.render(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/base.py", line 172, in render
    return self._render(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/base.py", line 162, in _render
    return self.nodelist.render(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/defaulttags.py", line 312, in render
    return nodelist.render(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/base.py", line 938, in render
    bit = node.render_annotated(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/base.py", line 905, in render_annotated
    return self.render(context)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/template/defaulttags.py", line 446, in render
    url = reverse(view_name, args=args, kwargs=kwargs, current_app=current_app)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/urls/base.py", line 87, in reverse
    return iri_to_uri(resolver._reverse_with_prefix(view, prefix, *args, **kwargs))
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/django/urls/resolvers.py", line 685, in _reverse_with_prefix
    raise NoReverseMatch(msg)
django.urls.exceptions.NoReverseMatch: Reverse for 'topic' with arguments '('',)' not found. 1 pattern(s) tried: ['topic/(?P<id>[0-9]+)$']
[Tilde-backend] [2024-04-02 07:41:06] ERROR [django.server:157] "GET /users/17/board/RB HTTP/1.1" 500 212678
```

Commenting out the two details buttons temporarily resolves this as mentioned, I think you maybe need to create more cards? You can use the `python manage.py create_demo_data` script if you haven't already and then navigate to the user's who have cards' boards

mfanaKoPitori:
@Ngoakor12, thanks. I am able to reproduce the error now.

Ngoakor12 (changes requested):
This doesn't address the issue. It instead made the code slightly worse. The issue is not that a user is missing. The issue is topic cards don't have a topic_progress instance and so won't have an id.

Please undo your latest changes and instead just check that a details button has topic progress before moving trying to render that button. You can use the "and" operator on this line:
```
 {%if card.content_item.content_type_nice == "topic"}
```

Ngoakor12:
This is the fourth feedback round, please set up a pair session as agreed upon in the last sprint plan. Set up something for tomorrow, if you need help, if not, do it anyway, I'll review this in the session.

backend/curriculum_tracking/helpers.py
```diff
@@ -182,3 +182,15 @@ def agile_card_reviews_outstanding(user):
 def pull_request_reviews_outstanding(user):
     # TODO: only implement this once the github webhook is healthier
     return []
+
+
+def get_days_hours_minutes_string(timestamp):
+    if timestamp:
```
Ngoakor12: remove the if check and return none.

Ngoakor12 (approved):

backend/curriculum_tracking/tests/factories.py
```diff
@@ -238,6 +238,7 @@ class Meta:
     recruit_project = factory.SubFactory(
         RecruitProjectFactory, start_time=timezone.now()
     )
+    topic_progress = factory.SubFactory(TopicProgressFactory)
```
sheenarbw: A single card cannot be linked to both a topic and a project. This factory now creates an invalid object.

backend/curriculum_tracking/models.py
```diff
@@ -941,6 +947,28 @@ def get_activity_log_summary_data(self):
...
+        relevant_logs = sorted(
```
sheenarbw: Please remove this function.  If you find yourself putting the same code in 2 different PRs then something has gone wrong.  Please keep your git tidy.

I dont think we ever had duration information show up on the frontend here. Why is it being added into this PR?

backend/curriculum_tracking/tests/test_Topic_Progress.py
```diff
@@ -0,0 +1,58 @@
...
+
+from curriculum_tracking.tests import factories
+from curriculum_tracking.models import AgileCard, ContentItem
+from activity_log.models import LogEntry
+import curriculum_tracking.activity_log_entry_creators as log_creators
+
+
+class TestTopicProgress(TestCase):
+
+    def test_get_card_duration_in_current_column_returns_none_when_there_are_no_log_entries(
```
sheenarbw: please remove. There is a totally different issue about calculating the duration. This is in the wrong place now.

backend/curriculum_tracking/tests/test_Topic_Progress.py
```diff
@@ -0,0 +1,58 @@
...
+            content_item=factories.ContentItemFactory(content_type=ContentItem.TOPIC),
+            status=AgileCard.READY,
+        )
+
+        card.assignees.set([user])
+        card.start_topic()
+        self.assertIn("get_card_duration_in_current_column", dir(card.topic_progress))
+        self.assertEquals(card.topic_progress.get_card_duration_in_current_column, None)
+
+    def test_get_card_duration_in_current_column_returns_correct_duration(self):
```
sheenarbw: Please remove

backend/frontend/templates/frontend/course_component_details/page.html
```diff
@@ -1,33 +1,54 @@
...
+                    {% endwith %}
+                </div>
+            
+                <div class="p-2 border-2 rounded-md ">
+                    <p>
+                        Status : {{board_status}}
+                    </p>
+                </div>
+
+                {% if duration %}
```
sheenarbw: Why is this here?

sheenarbw (changes requested):
Here is the task spec:

image

Please stick to the task.

Ray, this task is simple on purpose.

There is another pr that handles duration stuff. Why is your duration code in 2 PRs? Why is the duration stuff even being displayed?

backend/curriculum_tracking/models.py
mfanaKoPitori: @sheenarbw, I was displaying duration information because of this point in the task spec:
Screenshot 2024-04-07 124924

backend/frontend/templates/frontend/course_component_details/page.html
mfanaKoPitori: @sheenarbw, I was displaying duration information because of this point in the task spec:
Screenshot 2024-04-07 124924

Ngoakor12 (approved):

backend/frontend/templates/frontend/course_component_details/page.html
```diff
@@ -1,33 +1,46 @@
 {% extends "frontend/base_logged_in.html" %}
 
 {% block content %}
-    <div class="mx-auto p-6 bg-white my-4 sm:w-11/12 lg:w-8/12">
+    <div class="mx-auto p-6 bg-white my-4 sm:w-11/12 lg:w-8/12 w-full overflow-auto">
         <h1 class="{{ styles.heading1 }}">
-            {{course_component.title}}
+            {{course_component.content_item.content_type_nice | title}}: {{ course_component.content_item.title }}
```
sheenarbw: The `| title` is going to cause problems.

The lx coaches sometimes ask me to run scripts. Those scripts take the title of the content item as input. It's case sensitive.

Te lx coaches get those titles from the Tilde frontend. So if the tilde frontend changes the case things will start to break.

I do think it's good to make the titles consistent. But I dont think it would be good to do it now. It'll cause chaos and we'll either end up making more work for the lx coaches, or we'll need to upgrade scripts to allow case insensitivity.

backend/frontend/templates/frontend/course_component_details/page.html
```diff
@@ -1,33 +1,46 @@
...
-                                {% if not forloop.last %},{% endif %}
-                            {% endfor %}
-                        {% endwith %}
-                    </div>
-                
                     <div class="p-2 border-2 rounded-md ">
-                        <p class="bold">
-                            Status : {{board_status}}
-                        </p>
+                        Assignees:
```
sheenarbw: It would be good to refactor things a bit I think. 

Both topics and projects have assignees. We should not have 2 blocks of code that do the same thing.

Try to keep this stuff DRY.

backend/frontend/templates/frontend/course_component_details/page.html
```diff
@@ -1,33 +1,46 @@
...
+                        </ul>
                     </div>
                 </div>
-            
+            {% endif %}
+
+            {% if course_component.content_item.content_type_nice == "project" %}
                 <div class="p-2 border-2 rounded-md grid gap-2">
-                    <div class="bold p-2 border-2 rounded-md ">
+                    <div class="p-2 border-2 rounded-md ">
```
sheenarbw, line 49: See, we have assignees again over here. Why?

backend/frontend/views.py
```diff
@@ -368,16 +373,19 @@ def action_start_card(request, card_id):
...
+def course_component_details(request, id):
+    request_url = request.build_absolute_uri()
+    type = request_url.split("/")[3]
 
-    board_status = [
-        value
-        for key, value in AgileCard.STATUS_CHOICES
-        if key == project.agile_card_status
-    ][0]
+    if type == "topic":
```
sheenarbw: This is a bit messy. And it also leaves space for typos and nonesensce down the road because urls chance. We should keep our url code cohesive. This cunction should not know about what the url structure is. 

We have 2 options:

1. change the url to something like this "progress_details/<content_type>/<id>"
2. make use of the resolve function to get the view name 

I think this will work:

```
   from django.core.urlresolvers import resolve
   current_url = resolve(request.path_info).url_name
```

That said, I think option 1 is better. Option 2 is brittle because it assumes that the view name is unlikely to change. We might want to change it because `project` and `topic` are not very descriptive in that context.

sheenarbw: If you are doing something that feels like a hack, it's probably a hack. 

Always think about the things that the code knows about. The knowledge of the code. What will happen if something changes? Is it likely to change?

Try to see the Django way to do things. Try to understand why best practices are in place. Why do we not hardcode urls? Since we don't hardcode urls why would you think it's ok to write code that expects a certain url structure? Does this not cause exactly the same problems as a hardcoded url? 

It's like you coded this thing with a hammer. You're working around the system in order to force things to work. Work with the framework not against it.

sheenarbw (changes requested):
This still needs some polish

backend/frontend/templates/frontend/course_component_details/page.html
```diff
@@ -2,45 +2,51 @@
 {% load custom_tags %}
 
 {% block content %}
-    <div class="mx-auto p-6 bg-white my-4 sm:w-11/12 lg:w-8/12">
+    <div class="mx-auto p-6 bg-white my-4 sm:w-11/12 lg:w-8/12 w-full overflow-auto">
         <h1 class="{{ styles.heading1 }}">
-            {{course_component.title}}
```
Ngoakor12, line 7: Remove the first part ie. `project:` Just leave the project or topic title

backend/frontend/urls.py
```diff
@@ -68,9 +68,14 @@
     ),
     # Course component details
     path(
-        "project/<int:project_id>",
+        "project_details/<int:id>/<content_type>",
```
Ngoakor12: The suggest was for the urls to looks something like: `project_details/project` with the id at the end.

The way you are currently doing it is confusing because if the first part is `project_details` what's the use for `project` at the end?

Ngoakor12 (changes requested):

backend/frontend/urls.py
Ngoakor12: I made a typo on the above review, it should be `progress_details/project` not `project_details/project` or `topic_details/project`

Ngoakor12 (changes requested):
if you are going to rename the view file also rename the template folder

Ngoakor12 (approved):

sheenarbw (approved):

PR #710, opened 2024-03-17
Upgrade project_review_coordination_unclaimed view by filtering out projects the current user has reviewed since the last request

Ngoakor12:

## Description:

## I solemnly swear that:

### DIFF
--- a/backend/project_review_coordination/models.py
+++ b/backend/project_review_coordination/models.py
@@ -47,10 +47,10 @@ class ProjectReviewBundleClaim(models.Model):
             .exclude(
                 recruit_project__project_review_bundle_claims__is_active=True
             )  # if the project is already in an active claim, skip it
-            .annotate(has_reviewed=Exists(reviewed_projects_subquery))
-            .filter(
-                has_reviewed=False
+            .annotate(
+                has_reviewed=Exists(reviewed_projects_subquery)
             )  # exclude cards with recent project reviews by user
+            .filter(has_reviewed=False)
             .order_by("recruit_project__review_request_time")[:50]  # earliest first
             .prefetch_related("content_item", "recruit_project")
         )
--- a/backend/project_review_coordination/models.py
+++ b/backend/project_review_coordination/models.py
@@ -1,7 +1,7 @@
 from django.db import models
 from django.contrib.auth import get_user_model
 from datetime import timedelta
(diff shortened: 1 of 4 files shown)


backend/project_review_coordination/models.py
```diff
@@ -31,3 +31,28 @@ def save(self, *args, **kwargs):
             # have pk
             self.due_timestamp = timezone.now() + timedelta(hours=1)
         super(ProjectReviewBundleClaim, self).save(*args, **kwargs)
+
+    def get_projects_user_can_review(user):
```
sheenarbw, line 36: is this meant to be a `@staticmethod`?

Just calling the first argument `user` instead of `self` doesn't stop it from being an instance method.

Take a look at `@staticmethod` and `@classmethod` decorators.

sheenarbw (changes requested):
Looking good. Just one thing

backend/project_review_coordination/models.py
Ngoakor12: TIL :smile:

backend/project_review_coordination/tests/test_get_projects_user_can_review.py
```diff
@@ -0,0 +1,59 @@
...
+            self.project_user_has_reviewed
+        ]
+
+    def test_get_projects_user_can_review_returns_the_right_projects(self):
+        projects_user_can_review = (
+            ProjectReviewBundleClaim.get_projects_user_can_review(self.user)
+        )
+
+        assert self.project_user_has_reviewed not in projects_user_can_review
+        assert len(self.projects_in_review) != len(projects_user_can_review)
```
kingraphaii: Because the length of the projects the user can review is the only thing we are interested in here, I am not too sure about the second assert here; maybe we can assert the length of what we expect the user can review instead. I think it's more specific.

like `assert len(projects_use_can_review) == [a specific number of projects the user can review we expect here]`.

If we have a lot of projects in the `projects_in_review` list, and we are only interested in the length of the projects the user can review, it would be better to assert the length of what we expect the user can review instead of asserting the length of the `projects_in_review` list.

kingraphaii (changes requested):

kingraphaii (approved):

sheenarbw (approved):

PR #714, opened 2024-03-31
Add total duration function to classes topic progress and recruit project

mfanaKoPitori:

## Description:

## I solemnly swear that:


backend/curriculum_tracking/helpers.py
```diff
@@ -182,3 +187,44 @@ def agile_card_reviews_outstanding(user):
...
+        days, remainder = divmod(seconds, 86400)
+        hours, remainder = divmod(remainder, 3600)
+        minutes, remainder = divmod(remainder, 60)
+
+        return f"{int(days)} days, {int(hours)} hours, {int(minutes)} minutes"
+
+    return None
+
+
+def get_blant_duration(id, card_status):
```
Ngoakor12: what does "blant" mean :sweat_smile: ? Is it not possible to use another easy-to-understand word?

backend/curriculum_tracking/helpers.py
```diff
@@ -182,3 +187,44 @@ def agile_card_reviews_outstanding(user):
...
+    card_started_logs = sorted(
+        LogEntry.objects.filter(
+            Q(event_type__name=CARD_STARTED),
```
Ngoakor12: Q objects don't seem necessary here since you are not doing any complex queries, learn more [here](https://www.scaler.com/topics/django/q-objects-and-f-objects/). You can do it the normal way here

backend/curriculum_tracking/helpers.py
```diff
@@ -182,3 +187,44 @@ def agile_card_reviews_outstanding(user):
...
+    )
+
+    if (
+        card_status != models.AgileCard.COMPLETE
+        or not card_started_logs
+        or not card_completed_logs
+    ):
+        return None
+
+    duration = card_completed_logs[-1].timestamp - card_started_logs[-1].timestamp
```
Ngoakor12: I think this duration should be kept as a timestamp. We can format it into whatever string we want when we need to, maybe in the frontend or wherever this info will be used.

backend/curriculum_tracking/helpers.py
Ngoakor12: The function should also probably just move to the agile card model, I struggle to see where it would actually be used other than when referring to the agilecards

Ngoakor12 (changes requested):
- I think since the `get total duration` is not really different in topic progress or recruit project, maybe it should just live on the agilecard model?

I decided not to comment anything on the tests since I'm still not sure everything is where it should be.

backend/curriculum_tracking/helpers.py
```diff
@@ -182,3 +187,44 @@ def agile_card_reviews_outstanding(user):
 def pull_request_reviews_outstanding(user):
     # TODO: only implement this once the github webhook is healthier
     return []
+
+
+def get_formatted_duration_string(duration):
```
sheenarbw: What is this for? I dont see it being used anywhere?

backend/curriculum_tracking/helpers.py
```diff
@@ -182,3 +187,44 @@ def agile_card_reviews_outstanding(user):
...
+    if duration:
+        duration = duration
+        seconds = duration.total_seconds()
+        days, remainder = divmod(seconds, 86400)
+        hours, remainder = divmod(remainder, 3600)
+        minutes, remainder = divmod(remainder, 60)
+
+        return f"{int(days)} days, {int(hours)} hours, {int(minutes)} minutes"
+
+    return None
```
sheenarbw: Why return a string or None? 

Surely if no argument is supplies then that would be an error?

backend/curriculum_tracking/helpers.py
sheenarbw: I fully agree with Ngoako.

Consider how the data will be used. It should only be formatted for display when it is displayed.

If we wanted to check what is the average duration for a specific project we would need to jump through some crazy hoops here.

sheenarbw (changes requested):
This looks over baked.

Take a moment to google the KISS principle.

I think this task might be confusing because it is really really quite tiny.

Unfortunately most of this code is not going to be useful. It just adds a lot of complexity.

Here is the code we need:

```
def duration(self):
    return self.end_time - self.start_time 
```

We need a function like that on the project class and on the topic class. 

Please also look up separation of concerns, cohesion and loose coupling. Making a function that is meant to return something like a number return a string that is really hard to use for anything except display purposes is a bit of a funny thing to do. 

Junior coders write fancy code. Senior coders write the simplest code possible.

Ngoakor12 (approved):

backend/curriculum_tracking/models.py
```diff
@@ -761,6 +761,10 @@ def reviewer_user_names(self):
     def agile_card_status(self):
         return self.agile_card.status
 
+    @property
```
sheenarbw, line 764: Why are these properties? 

KISS, please, Raymond. Please now. We know you are clever. Stop trying to wow us with your fanciness. It's nice that you know about properties, but how is that appropriate here?

Also think about naming.

Here are the appropriate names for a property versus a normal instance method. 

```
topic_progress.duration
topic_progress.get_duration()
```

Once you get to a point where you are able to do simple things we can ask you to do harder things. 

Junior coders think fancy code makes them fancy people. Senior coders know that the simplest code is the best code. 

Please make some perfectly normal super simple functions that just do the thing they are supposed to do without extra bells and whistles. I would rewrite this the moment I try to use it.

backend/curriculum_tracking/tests/test_RecruitProject.py
```diff
@@ -82,3 +84,23 @@ def test_too_long(self):
         self.assertIn("-test-project-", repo_name)
         self.assertIn(self.user.first_name, repo_name)
         self.assertIn(self.user.last_name, repo_name)
+
+
+class get_total_duration(TestCase):
```
sheenarbw: This name is weird. Why?  Please take a look at the names of other test cases and see if you can pattern match.

backend/curriculum_tracking/tests/test_TopicProgress.py
```diff
@@ -0,0 +1,25 @@
+from django.test import TestCase
+from django.utils import timezone
+from datetime import datetime, timedelta
+
+from curriculum_tracking.tests import factories
+from curriculum_tracking.models import AgileCard, ContentItem
+
+
+class get_total_duration(TestCase):
```
sheenarbw: naming

backend/curriculum_tracking/tests/test_TopicProgress.py
```diff
@@ -0,0 +1,25 @@
...
+
+        card.topic_progress.start_time = datetime(
+            2024, 2, 12, 14, 6, 17, 373514, tzinfo=timezone.utc
+        )
+        card.topic_progress.end_time = datetime(
+            2024, 2, 12, 15, 6, 17, 373514, tzinfo=timezone.utc
+        )
+
+        self.assertEquals(
+            card.topic_progress.get_total_duration, timedelta(seconds=3600)
```
sheenarbw: This is not readable. It looks like it should be called like this:

get_total_duration()

This is because it is a verb, a doing thing It doesn't look like an attribute, i looks like a function. Properties should be named like attribures.

backend/curriculum_tracking/tests/test_RecruitProject.py
```diff
@@ -82,3 +84,23 @@ def test_too_long(self):
...
+        card.recruit_project.start_time = datetime(
+            2024, 2, 12, 14, 6, 17, 373514, tzinfo=timezone.utc
+        )
+        card.recruit_project.end_time = datetime(
+            2024, 2, 12, 15, 6, 17, 373514, tzinfo=timezone.utc
+        )
+
+        self.assertEquals(
+            card.recruit_project.get_total_duration, timedelta(seconds=3600)
+        )
```
sheenarbw, line 123: What happens if the end_time is empty?
What happens if the start and end times are empty?

Consider edge cases.

sheenarbw (changes requested):
Still overly fancy. The tests have funny names. And edge cases are not considered.

backend/curriculum_tracking/helpers.py
```diff
@@ -1,5 +1,10 @@
 from . import models
 from core import models as core_models
+from activity_log.models import LogEntry
+from curriculum_tracking.activity_log_entry_creators import (
+    CARD_STARTED,
+    CARD_MOVED_TO_COMPLETE,
+)
```
Ngoakor12, line 7: Is this still needed?

backend/frontend/templates/frontend/base_logged_in.html
```diff
@@ -11,9 +11,6 @@
 {% block main_content %}
     <div class="min-h-full">
         {% include 'frontend/partial_logged_in_main_header.html' %}
-        {% block secondary_header %}{% endblock %}
-        <main class="mx-auto max-w-7xl py-6 sm:px-6 lg:px-8 pb-0 overflow-auto">
-            {% block content %}{% endblock %}
-        </main>
```
Ngoakor12, line 17: Why are we removing this?

Ngoakor12 (changes requested):
Please review the "Files changed" tab to verify that you are doing the right things. There should not be 21 files changed for what you're doing.

Previous features seem to be removed, notebly:


- [This feature](https://github.com/Umuzi-org/Tilde/pull/719), it's part of work that Themba did.
- Same with this one. It looks like it's related to work done on [this PR](https://github.com/Umuzi-org/Tilde/pull/718)

backend/curriculum_tracking/models.py
```diff
@@ -941,6 +951,18 @@ def get_activity_log_summary_data(self):
             "flavour_names": self.content_item.flavour_names,
         }
 
+    @property
+    def duration(self):
+        if self.agile_card.status == AgileCard.BLOCKED or self.agile_card.status == AgileCard.READY:
+            return timedelta(0)
+        
+        if self.agile_card.status == AgileCard.IN_PROGRESS:
```
Ngoakor12, line 959: What about cards in review? and review feedback?

mfanaKoPitori:
@Ngoakor12, this happened because I was reverting merged commits a lot. How would I go about fixing this? Maybe revert all the changes up to a specific commit. But I can fix it; let me try.

Update: @Ngoakor12 It is too wrangled now. I will dump the pull request and start over. I should upskill more on advanced topics on Git to handle something like this in the future.

PR #717, opened 2024-04-01
filtering out restricted projects on project review coordination

kingraphaii:

## Description:

## I solemnly swear that:


backend/project_review_coordination/models.py
```diff
@@ -56,4 +60,24 @@ def get_projects_user_can_review(user):
...
+                continue  # prefetch_perms will fail if there are no teams
+
+            perm_checker.prefetch_perms(assignee_teams)
+            for view_permission in Team.PERMISSION_VIEW:
+                if any(
+                    (perm_checker.has_perm(view_permission, team) for team in assignee_teams)
+                ):
+                    permitted_cards.append(card)
+    
+        return permitted_cards
```
Ngoakor12: Is [this](https://github.com/Umuzi-org/Tilde/blob/f9172d81893ae25e99eddc2f287868b1bd8d028a/backend/frontend/views.py#L604) not a slicker way of writing the above. Then we wouldn't need to explicitly looper over perms

Ngoakor12 (changes requested):

Ngoakor12 (approved):

backend/project_review_coordination/models.py
```diff
@@ -56,4 +60,26 @@ def get_projects_user_can_review(user):
             .prefetch_related("content_item", "recruit_project")
         )
 
-        return cards
+        if user.is_superuser:
+            return cards
+
```
sheenarbw, line 65: I worry about efficiency, we are hitting the database quite a lot. This view is already really slow.

Here is some pseudocode for a faster version:

```
@lru_cache  # I'm not sure if this will work if we use django model instances as arguments. But we should try not to fetch the same users teams over and over 
def get_user_teams(user):
      return user.teams()

...
        viewable_teams = get all the teams the current user can view before we start looping
        if viewable_teams == []:
              return [] 

        for card in cards:
            assignee = card.assignees.first()
            assignee_teams = get_user_teams(assignee)

            for team in assignee_teams:
                 if team in viewable_teams:
                      permitted_cards.append(card)
                      break

return permitted_cards 
```

sheenarbw (changes requested):
This looks good, but we need to be a bit more careful about hammering th database. This page is already way too slow

backend/project_review_coordination/models.py
```diff
@@ -56,4 +74,22 @@ def get_projects_user_can_review(user):
...
+        
+        permitted_cards = []
+
+        for card in cards:
+            assignee = card.assignees.first()
+            assignee_teams = cls.get_user_teams(assignee)
+
+            if set(viewable_teams).intersection(assignee_teams):
+                permitted_cards.append(card)
+                continue
```
Ngoakor12: this continue doesn't really do anything

Ngoakor12 (approved):

backend/project_review_coordination/models.py
```diff
@@ -33,7 +36,22 @@ def save(self, *args, **kwargs):
         super(ProjectReviewBundleClaim, self).save(*args, **kwargs)
 
     @staticmethod
-    def get_projects_user_can_review(user):
+    @lru_cache
+    def get_user_teams(user):
+        return user.teams()
```
sheenarbw: These functions seem like they are in a weird place. Should they not be on the User class since we are fetching information about the user? 

Why would anyone think to get the user permissions by checking the ProjectReviewBundleClaim static method? 

If we organise the code so that it is on the user class, would it not be much more useful? The only way the code could be reused if it is here would be:
- through spaghetti imports OR
- by first refactoring the code so that the functions are in a reusable place

Would it not be nice if this was available everwhere?

```
user.teams_cached()
user.get_teams_for_user_cached(perms=Team.PERMISSION_VIEW)
```

or even

```
user.teams(cached=True) 
user.get_teams_for_user(perms=Team.PERMISSION_VIEW,cached=True)
```

sheenarbw (changes requested):
The code looks like it will work. But there are failing tests. And the functions are in a strange place.

Ngoakor12 (approved):

sheenarbw (approved):

PR #720, opened 2024-04-05
Use better placeholder avatars

Ngoakor12:

## Description:

## Screenshots/videos

image

image

## I solemnly swear that:


kingraphaii (approved):

backend/frontend/templatetags/custom_tags.py
```diff
@@ -0,0 +1,9 @@
+from django import template
+
+register = template.Library()
+
+
+@register.inclusion_tag("frontend/user/board/partial_user_avatar.html")
+def user_avatar(name, many=False):
```
sheenarbw: What is `many`? I dont think you are using this argument anywhere. What is it meant to do?

And why does the user avatar take in a name argument? Why not just a user? Why does the calling code need to know that the avtar is derived from the user's name? WHy is that logic scattered everywhere?

Should we even use the user's name? Who knows? If we needed to change that logic we should change it only once. It shouldn't be in every single template. This must be DRY and cohesive.

sheenarbw (changes requested):
What is `many` for?

Also, the code is not cohesive or DRY. Some of the avatar logic is in the calling code, some is in the template tag. 

Remember that we will eventually have profile pics. We should be able to update the code in one place to make it work.

backend/frontend/templatetags/custom_tags.py
Ngoakor12: `Many` was used to conditionally style a group of user avatars. But I moved that to the template.

The avatar logic is now in the template tag.

The user_avatar tag now takes in a user. The user profile is likely to live on the user model, so the only thing that would need updating once we have profile pictures is the user avatar template and the context for the template. We won't need to change all the templates where the user avatar is rendered.

backend/frontend/templates/frontend/user/partial_user_avatar.html
```diff
@@ -0,0 +1,5 @@
+<div
+  class="bg-slate-400 font-bold {% if many == 'True' %} border-2 border-white {%endif%} text-slate-100 w-[30px] h-[30px] rounded-full flex items-center justify-center flex-nowrap"
```
kingraphaii: Is `many` still relevant here

kingraphaii (changes requested):

backend/frontend/templates/frontend/user/partial_user_avatar.html
Ngoakor12: nope

kingraphaii (approved):

sheenarbw (approved):

PR #715, opened 2024-04-01
Add timezone conversion to claim due timestamp

kingraphaii:
Related issues: [please specify]

## Description:
Currently, due date on claims is show in UTC (the default).

What are you up to? Fill us in :)
This smol change will make SAST (Africa/Johannesburg) the display tz.

## I solemnly swear that:

- [x] My code follows the style guidelines of this project
- [x] I have merged the develop branch into my branch and fixed any merge conflicts
- [x] I have performed a self-review of my own code
- [ ] I have commented my code in hard-to-understand areas
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have tested new or existing tests and made sure that they pass

PR #721, opened 2024-04-09
Test failure fix: Made sure to use the correct time, because timezone was changed to the south african(jhb) timezone instead of default one.

mfanaKoPitori:
Related issues: Mom task: [New test failures: test_link_project_page_displays_correct_details](https://umuzi-projects.monday.com/boards/1274815998/pulses/1458534365)

## Description:

The test was not failing because the timezone used by the frontend app was recently changed by changes in this merged PR, [Linked Here](https://github.com/Umuzi-org/Tilde/pull/715)

If the timezone ever gets changed again, this test will fail again. I may need to change how I am testing if times are getting displayed altogether.

## I solemnly swear that:

### DIFF
--- a/backend/frontend/tests/test_course_component_details_pages.py
+++ b/backend/frontend/tests/test_course_component_details_pages.py
@@ -65,8 +65,8 @@ class TestLinkProjectDetailsPage(FrontendTestMixin):
 
         self.assertIn("learner_1@umuzi.org", body)
         self.assertIn("In Progress", body)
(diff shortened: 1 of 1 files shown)


Ngoakor12 (changes requested):
This works but if the timezone is different it will fail like you mentioned. Is there no way to make this test dynamic? If not I think we should be explicit that this is only testing the time in "Africa/Johannesburg" timezone

kingraphaii (approved):

sheenarbw (approved):

PR #725, opened 2024-04-17
Add review modal

kingraphaii:

## Description:

#### Add Review:
image

### Newly added review
image

#### Review Preview:
image

### Input validation
image

## I solemnly swear that:


backend/requirements.txt
```diff
@@ -85,3 +85,5 @@ playwright==1.40.0 # frontend testing
 
 # django-template-partials==23.4 # https://github.com/carltongibson/django-template-partials  possibly useful later on
 
+# rendering md reviews on the frontend
```
Ngoakor12, line 88: What about `pymdown-extensions`?

Ngoakor12 (changes requested):
This is a big task...looks clean though. It looks like it also tackles [this task](https://umuzi-projects.monday.com/boards/1274815998/pulses/1391273049)

- The add review form is missing proper form handling. We should let the user know when they try and send the form with missing info or something happened when sending the form? You can have a look at the one I did previously [here](https://github.com/Umuzi-org/Tilde/pull/628) for inspo
image

I might have missed some things since this is a bit task so I'll do another pass over this and let you know if I find something

backend/frontend/views.py
```diff
@@ -426,6 +449,64 @@ def progress_details(
...
+
+    if content_type == "project":
+        project = get_object_or_404(RecruitProject, id=id)
+        return project.request_user_can_add_review(logged_in_user)
+
+    return False
+
+@user_passes_test_or_forbidden(check_user_can_add_review)
+def action_add_review(request, content_type, id):
+    if request.method == "POST":
```
Ngoakor12: This function has a lot of nesting, which makes reading it a bit hard to follow.

I don't think checking the method is as seen with the other action related methods and the fact that this will always be a POST.

backend/frontend/views.py
```diff
@@ -426,6 +449,64 @@ def progress_details(
...
+        if content_type == "project":
```
Ngoakor12: Ideally, this should always be a project. Can we just change this to an assert instead?

Ngoakor12 (changes requested):
Can we remove the whole review section for topics?
image


I just noticed that the avatar on the modal gets squished on smaller screens and maybe we can align the icons (avatar and cancel) to the top:
image


I also got this error when running the tests?
```
ERROR: test_reviewer_by_permission_can_add_review (frontend.tests.test_add_review_button.TestProgressDetailsAddReviewButton)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/frontend/tests/test_add_review_button.py", line 126, in test_reviewer_by_permission_can_add_review
    self._submit_review("not yet competent")
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/frontend/tests/test_add_review_button.py", line 76, in _submit_review
    self.page.get_by_role("button", name="Add Review").click()
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/playwright/sync_api/_generated.py", line 15784, in click
    self._sync(
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/playwright/_impl/_sync_base.py", line 115, in _sync
    return task.result()
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/playwright/_impl/_locator.py", line 159, in click
    return await self._frame.click(self._selector, strict=True, **params)
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/playwright/_impl/_frame.py", line 484, in click
    await self._channel.send("click", locals_to_params(locals()))
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/playwright/_impl/_connection.py", line 62, in send
    return await self._connection.wrap_api_call(
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/playwright/_impl/_connection.py", line 492, in wrap_api_call
    return await cb()
  File "/home/ngoako/Desktop/code/umuzi/Tilde/backend/venv/lib/python3.10/site-packages/playwright/_impl/_connection.py", line 100, in inner_send
    result = next(iter(done)).result()
playwright._impl._errors.TimeoutError: Timeout 30000ms exceeded.

Ran 44 tests in 179.080s

FAILED (errors=1)
```

Ngoakor12 (approved):

backend/curriculum_tracking/models.py
```diff
@@ -761,6 +761,39 @@ def reviewer_user_names(self):
...
+        if not user:
+            return False
+        
+        if user.is_superuser:
+            return True
+        
+        if user in card.reviewers.all():
+            return True
+        
+        return card.request_user_is_assignee(user) or any(
```
sheenarbw, line 793: So the card assignee can add reviews on their own projects? That seems a little odd. Surely we dont want to allow people to review their own stuff?

backend/frontend/templates/frontend/base_logged_in.html
```diff
@@ -4,7 +4,12 @@
 {% block main_head %}
     <script src="https://cdn.jsdelivr.net/npm/chart.js"></script>
     <script defer src="https://cdn.jsdelivr.net/npm/alpinejs@3.x.x/dist/cdn.min.js"></script>
+    <script src="https://unpkg.com/@highlightjs/cdn-assets@11.9.0/highlight.min.js"></script>
```
sheenarbw, line 7: This script tag is in frontend/base.html already. Why does it need to be included twice?

This kind of thing comes up a lot. If we include something in a parent template then it is included already. We dont need to add it twice

backend/frontend/views.py
```diff
@@ -427,6 +453,63 @@ def progress_details(
...
+        response = render(
+            request,
+            "frontend/progress_details/partial_review_form.html",
+            {
+                "review_form": form,
+                "course_component": project,
+            }
+        )
+        response["HX-Retarget"] = "#review-form"
+        response["HX-Reswap"] = "innerHTML"
```
sheenarbw, line 509: Generally we have all the htmx stuff handled in the html. I think you are doing things this way so that it is more like the standard Django way of doing things,. but it seems a bit messy to me.

Now the view code needs to understand how the html is structured and how it is meant to behave. It seems like this function is taking on too much responsibility. It's not cohesive, it is tightly coupled.

I think it would be better to just have a different vide for the form submission and then do all the HTMX stuff in the simplest and clearest way possible. Django form code is very simple, we dont need to follow this pattern.

Here is something I did recently on a demo project for a presentation. It's a todo list app:

https://github.com/sheenarbw/demo-testing-django-with-playwright/blob/main/todo/templates/todo/index.html   < Here we render the form. Look at line 53 onwards. 

The form goes to a view called "action_add_new_todo". You can see this on line 28 here: https://github.com/sheenarbw/demo-testing-django-with-playwright/blob/main/todo/views.py 

It's not perfect code but it shows a nice simple way of oing this stuff.

backend/frontend/forms.py
```diff
@@ -51,6 +53,18 @@ def get_password_reset_url(self):
...
         )
 
+
+class RecruitProjectReviewForm(ThemedFormMixin, forms.ModelForm):
+
+    class Meta:
+        model = RecruitProjectReview
+        fields = ["status", "comments", "recruit_project", "reviewer_user"]
+        widgets = {
+            "recruit_project": forms.HiddenInput(),
```
sheenarbw, line 63: Why are these fields necessary? Is the recruit_project id not in the url we post to? And is the reviewer_user not in the session? This seems sketchy if you are using them because anyone can just do a post from their browser and add a review to any card and make it look like it came from any person. So you would need to validate this stuff anyway.

backend/frontend/templates/frontend/progress_details/js_exec_action_prepend_new_review.html
```diff
@@ -0,0 +1,10 @@
+{% include "frontend/progress_details/partial_review.html" %}
+<script>
+    const newReview = document.getElementById("review-{{ review.id }}");
```
sheenarbw, line 3: This file has a funny name. It doesn't prepend anything. The prepending happens elsewhere

backend/frontend/templates/frontend/progress_details/partial_add_review_modal_content.html
```diff
@@ -0,0 +1,27 @@
...
+            <h1 class="font-bold leading-tight text-gray-800 ">
+                Add Review for {{ course_component.content_item.content_type_nice|slice:1|title }}: {{ course_component.content_item.title }}
+            </h1>
+            <span class="text-xs text-slate-500">
+                Submitting review as <span class="font-bold">{{ request.user.email }}</span>
+            </span>
+        </div>
+    </div>
+
+    <button @click="modalOpen = false"
```
sheenarbw, line 17: Is this a part of the header area for the modal? Would it not make sense to have the modal have this button by default instead of having to implement it every time anyone uses the modal partial?

The base modal could have a block for the heading and the content, and then the heading should always be styled the same way and the close button would always be in the same place.

backend/frontend/templates/frontend/progress_details/partial_add_review_modal_content.html
```diff
@@ -0,0 +1,27 @@
...
+{% block modal_content %}
+<div class="flex items-center justify-between">
+    <div class="flex items-center gap-2">
+        <div class="hidden md:block">{% user_avatar request.user %}</div>
+        <div class="flex flex-col gap-0">
+            <h1 class="font-bold leading-tight text-gray-800 ">
+                Add Review for {{ course_component.content_item.content_type_nice|slice:1|title }}: {{ course_component.content_item.title }}
+            </h1>
+            <span class="text-xs text-slate-500">
+                Submitting review as <span class="font-bold">{{ request.user.email }}</span>
```
sheenarbw, line 12: Why are you telling the user their own email address? They are logged in and will be able to see who they are in the main header area

backend/frontend/templates/frontend/partial_modal_base.html
```diff
@@ -0,0 +1,33 @@
...
+        <div x-cloak
+             x-show="modalOpen"
+             x-transition:enter="transition ease-out duration-300 transform"
+             x-transition:enter-start="opacity-0 translate-y-4 sm:translate-y-0 sm:scale-95"
+             x-transition:enter-end="opacity-100 translate-y-0 sm:scale-100"
+             x-transition:leave="transition ease-in duration-200 transform"
+             x-transition:leave-start="opacity-100 translate-y-0 sm:scale-100"
+             x-transition:leave-end="opacity-0 translate-y-4 sm:translate-y-0 sm:scale-95"
+             class="inline-block w-full max-w-xl p-8 my-20 overflow-hidden text-left transition-all transform bg-white rounded-lg shadow-xl 2xl:max-w-2xl">
+            {% block modal_content %}
```
sheenarbw, line 28: Add a header area here. It should have space for some header text, and it should have a close button.

backend/frontend/templates/frontend/progress_details/partial_review.html
```diff
@@ -0,0 +1,27 @@
+{% load custom_tags %}
+<div id="review-{{ review.id }}">
+    <span class="text-xs">{{ review.timestamp|date:"d/m/Y, H:i:s" }}</span>
```
sheenarbw, line 3: We should standardise how we show dates and times.

YYYY MM DD is nice for numerical dates. It's a format adopted elsewhere because if you put these kinds of date strings in alphabetical order, then they are in ascending date order.
Or we can say the month as a word if it's confusing to read. Like "8 May 2024". This might be a bit better on the frontend because it's very clear.

For the time, we probably don't need to know the exact seconds, hours and minutes are enough. It's good to think about real estate on the screen, there is a lot going on so if we show unnecessary information it just makes things more cluttered.

backend/frontend/tests/test_add_review_button.py
```diff
@@ -0,0 +1,164 @@
...
+                content_type=ContentItem.PROJECT,
+                project_submission_type=ContentItem.LINK,
+            ),
+            status=AgileCard.IN_REVIEW,
+        )
+        self.card.assignees.set([self.user])
+        self.card.reviewers.add(self.reviewer_by_assignment)
+        self.card.recruit_project.recruit_users.set([self.user])
+
+    def _get_project_progress_details_url(self, project):
```
sheenarbw, line 69: Why does this start with an _ and make_in_review_project_card does not?

backend/frontend/tests/test_add_review_button.py
```diff
@@ -0,0 +1,164 @@
...
+    
+    def _submit_review(self, status):
+        self.page.get_by_role("button", name="Add Review").click()
+        self.page.wait_for_load_state()
+        self.page.select_option("select#id_status", status)
+        self.page.query_selector("textarea#id_comments").fill(self.SAMPLE_REVIEW_COMMENT)
+        self.page.query_selector("#submit_review_button").click()
+        self.page.wait_for_load_state('networkidle')
+
+    def test_super_user_can_add_review(self):
```
sheenarbw, line 82: Is there a way to speed up these tests?

If you l;ook at the time it takes to run all the tests, it's friggin long. Not absolutely every test of functionality needs to be a full on integration test. WQe can do small, fast, clear unit tests too. 

Look into the "testing pyramid". Try to only write tests that spin up a browser when necessary. 

If unit tests are slow to run then they become less useful.

You can also write tests that render html but that do not fire up a browser.

sheenarbw (commented):
You're on the right track but there is a lot of polish to add here. It would also be very good to start thinking about the testing pyramid.

They say that when the only tool you have is a hammer then every problem looks like a nail. Playwright is a great hammer, but it's being used when it isn't needed and it's making simple small tests into long, slow, complicated tests.

PR #733, opened 2024-05-02
only bundle owner can add time and unclaim bundle

TheMaja:

## Description:

## Screenshots/videos
[screen-capture (7).webm](https://github.com/Umuzi-org/Tilde/assets/73611175/efa241e5-fa89-4ca2-9663-bb1ead1725b3)

## I solemnly swear that:


Ngoakor12 (changes requested):
This looks like resolves the issue, judging by the demo video, can you also please test it?

backend/frontend/tests/test_all_claims_page.py
```diff
@@ -0,0 +1,39 @@
+from django.test import TestCase, Client
```
Ngoakor12: We are making use of Playwright for UI testing so let's make sure we are consistent.

Have a look at the other frontend tests to see how we handle the login and asserting if things are on the page or not. Eg. make use of `FrontendMixin`, etc.

backend/frontend/tests/test_all_claims_page.py
```diff
@@ -0,0 +1,39 @@
...
+        self.client.login(email=self.user.email, password=self.user.email)
+        # claimed by user 2
+        self.claim = ProjectReviewBundleClaim.objects.create(
+            claimed_by_user=self.user_2
+        )
+        self.url = reverse("project_review_coordination_all_claims")
+
+    def test_page_loads(self):
+        response = self.client.get(self.url)
+        self.assertEqual(response.status_code, 200)
```
Ngoakor12: no need to test page loads

Ngoakor12 (changes requested):

backend/frontend/tests/test_claims_page_button.py
```diff
@@ -0,0 +1,57 @@
...
+        self.page.goto(url)
+        self.page.wait_for_load_state()
+        body = self.page.locator("body")
+
+        expect(body).to_contain_text("Add 15 minutes")
+        expect(body).to_contain_text("Unclaim bundle")
+
+    def test_buttons_not_displayed_when_not_claimant(self):
+
+        super().setUp()
```
Ngoakor12: you can call this only once in your setUp method

Ngoakor12 (changes requested):
Almost there, just a few cleanup suggestions

- make use of a setUp method to put the common code between the two methods
- no need to attach variables to self when they're only useful in the function they're being declared in. E.g.
```
self.claim = ProjectReviewBundleClaim.objects.create(
            claimed_by_user=self.user_2
        )
'// can be changed to 
claim = ProjectReviewBundleClaim.objects.create(
            claimed_by_user=self.user_2
        )
```

backend/frontend/tests/test_claims_page_button.py
```diff
@@ -0,0 +1,41 @@
...
+        self.user = UserFactory(
+            email="staff_1@umuzi.org",
+            is_staff=True,
+        )
+        self.user.set_password(self.user.email)
+        self.user.save()
+
+        self.user_2 = UserFactory(email="staff_2@umuzi.org", is_staff=True)
+        self.user_2.set_password(self.user_2.email)
+        self.user_2.save()
```
Ngoakor12: Almost there...

There's no need for user_2 to:
- be in `setUp` since it will only be used in the second test
- it should, therefore, also not be attached to `self`

The difference between the 2 users is that we need to be logged in as one user before running all the tests and the first test makes use of the logged in user

Ngoakor12 (changes requested):

Ngoakor12 (approved):

PR #735, opened 2024-05-05
Allow search for users

kingraphaii:

## Description:

## I solemnly swear that:

### DIFF
--- a/backend/frontend/tests/test_page_users_and_teams_nav.py
+++ b/backend/frontend/tests/test_page_users_and_teams_nav.py
@@ -1,4 +1,5 @@
 from core.tests.factories import UserFactory, TeamFactory
+from social_auth.tests.factories import SocialProfileFactory
 from .frontend_test_mixin import FrontendTestMixin
 from guardian.shortcuts import assign_perm
 from core.models import Team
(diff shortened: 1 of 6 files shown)


Ngoakor12 (approved):

backend/frontend/tests/test_page_users_and_teams_nav.py
```diff
@@ -121,3 +122,112 @@ def test_user_sees_teams_that_match_the_search_term(self):
...
+
+    def test_all_users_show_up_onload(self):
+        self.do_login(self.user)
+        self._goto_users_page()
+
+        self.page.wait_for_load_state("networkidle")
+
+        users_div = self.page.locator("#users")
+        for user in [self.user, self.user_without_github, self.user_with_github]:
+            expect(users_div).to_contain_text(user.get_full_name())
```
sheenarbw, line 167: Do we need to spin up a whole browser for all of these tests? 

And what happens if we have 1000 users? We shouldn't expect them all to be displayed.

backend/frontend/tests/test_page_users_and_teams_nav.py
```diff
@@ -121,3 +122,112 @@ def test_user_sees_teams_that_match_the_search_term(self):
...
+    
+    def test_superusers_do_not_appear(self):
```
sheenarbw, line 169: Why? This seems like a strange choice given that we sometimes need superusers to review code and use the Tilde frontend to monitor people's performance in code review.

backend/frontend/tests/test_page_users_and_teams_nav.py
```diff
@@ -121,3 +122,112 @@ def test_user_sees_teams_that_match_the_search_term(self):
...
+
+        users_div = self.page.locator("#users")
+        expect(users_div).to_contain_text("John Doe")
+    
+    def test_users_can_search_last_name(self):
+        self.do_login(self.user)
+        self._goto_users_page()
+
+        search_input = self.page.locator("#user-search-term-box")
+        search_input.press_sequentially("Doe")
```
sheenarbw, line 228: We should have a search function that takes in a string and returns a database query then test that as a normal function instead of using playwright for everything. 

Remember the testing pyramid. Tests should be fast and specific. Favor unit tests over integration tests.

backend/frontend/views.py
```diff
@@ -672,6 +672,44 @@ def action_stop_card(request, card_id):
     )
 
 
+@login_required()
+def view_partial_users_list(request):
+    exclude = {
+        "is_superuser": True,
```
sheenarbw, line 678: Why are you introducing a data structure for this filter but not for the ` active` filter? Also, please do display superusers.

backend/frontend/views.py
```diff
@@ -672,6 +672,44 @@ def action_stop_card(request, card_id):
...
+    }
+    all_users = User.objects.filter(active=True).order_by("first_name", "last_name").exclude(**exclude)
```
sheenarbw, line 680: We should not display all users, we should display only the users that the current logged in user is allowed to see.

As things stand, any learner can log in and get a big list of users. We need to think about permissions all the time.

backend/frontend/views.py
```diff
@@ -672,6 +672,44 @@ def action_stop_card(request, card_id):
...
+    total_user_count = all_users.count()
+
+    if request.method == "POST":
+        form = SimpleSearchForm(request.POST)
+
+        if form.is_valid():
+            search_term = form.cleaned_data["search_term"]
+
+            all_users = User.objects.filter(active=True).filter(
```
sheenarbw, line 689: Rather make a base query and then filter it instead of writing the same query twice. 

So:
```
all_users = User.objects.filter(.....
...
if form.is_valid()
     all_iusers = all_users.filter(...)
```

Also, why are you only filtering by istartswith? Why not icontains? Surely we dont need the people using this to remember how things start?

Also there might be multiple search terms. Eg: "Percival Rapha" in which case we would want to search for each word.

sheenarbw (commented):
Please include some screenshots in your PR

PR #731, opened 2024-04-29
Feat/expand unclaimed bundles

kingraphaii:
Related issues: [please specify]

## Description:

It would be useful to see at least:

- the project title

- assignee

- number of positive reviews since the last review request

- a link to the details page

- a list of the staff members who reviewed.

## Screenshots/videos

image

image

## I solemnly swear that:

- [x] My code follows the style guidelines of this project
- [x] I have merged the develop branch into my branch and fixed any merge conflicts
- [x] I have performed a self-review of my own code
- [x] I have commented my code in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have tested new or existing tests and made sure that they pass

PR #736, opened 2024-05-07
Feat/add loading spinner on "Load More"

kingraphaii:

## Description:

## Screenshots/videos

image

## I solemnly swear that:

### DIFF
--- a/backend/frontend/templates/frontend/base.html
+++ b/backend/frontend/templates/frontend/base.html
@@ -25,7 +25,17 @@
       crossorigin="anonymous"
     ></script>
 
-    
+    <style>
+      .htmx-indicator {
+        display: none;
+      }
+      .htmx-request .htmx-indicator{
+          display:inline-flex;
+      }
+      .htmx-request.htmx-indicator{
+          display:inline-flex;
+      }
+    </style>
 
     {% block main_head %}{% endblock %}
   </head>
--- a/backend/frontend/templates/frontend/users_and_teams_nav/page.html
+++ b/backend/frontend/templates/frontend/users_and_teams_nav/page.html
@@ -19,9 +19,12 @@
                         Show Users
                     </button>
                 </div>
-                <form method="post" action="" class="mb-3 flex items-center gap-1">
+                <form method="post" action="" class="mb-3 relative flex items-center gap-1">
                     {% csrf_token %}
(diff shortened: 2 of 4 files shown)


backend/frontend/templates/frontend/base.html
```diff
@@ -25,7 +25,17 @@
...
+      .htmx-indicator {
+        display: none;
+      }
+      .htmx-request .htmx-indicator{
+          display:inline-flex;
+      }
+      .htmx-request.htmx-indicator{
+          display:inline-flex;
+      }
+    </style>
```
Ngoakor12, line 38: This looks a bit weird, is this not something we can add to the theme and style using Tailwind?

The last 2 selectors are almost similar and do the same thing maybe we just separate them with a comma?

Ngoakor12 (changes requested):
TIL: that you can throttle your network via dev tools, that's the only way I could confirm the loading spinner :sweat_smile: 

I'll update the PR screenshots section

Ngoakor12 (changes requested):
I think this is much better. Just more suggestions, let me know what you think:

image

- can we just make the whole group of avatars clickable when we have more than 2 reviewers?
- I agree that the avatars are potentially cluttering the UI, especially because they don't really do anything unlike the ones in the current Tilde which link to the users' boards. Plus hovering over them still brings up the emails even though they are fully visible.
- I know we can close the pop-up by clicking away, but can we also have an explicit close icon(x) for less technical users?


...
**Side note**: Eventually I think moving to an expanding pop-up similar to what @kingraphaii  did on [this PR](https://github.com/Umuzi-org/Tilde/pull/731) might be best. I think it will look cleaner than us trying to create our own tooltip. And if I'm not mistaken that's similar to what you and Ant where planning for the redesign of cards on the current Tilde, I remember seeing it on Figma

backend/frontend/templates/frontend/base.html
kingraphaii: As per the docs [here](https://htmx.org/attributes/hx-indicator/#:~:text=When%20a%20request%20is%20in%20flight%2C%20this%20will%20cause%20the%20htmx%2Drequest%20class%20to%20be%20added%20to), htmx adds that class to the loading indicator and I could not find another way of making it work without making it messy. 

as for the second part about the "duplicated" selectors, I thought that it'd be easier to read that way as opposed to separating them with a comma, because then, it'd be harder to justify while reading that they aren't duplicates.

Ngoakor12 (approved):

sheenarbw (approved):

PR #741, opened 2024-05-08
Allow search for Team

Ngoakor12:

## Description:

## Screenshots/videos

image
image
image
image

## I solemnly swear that:


backend/frontend/views.py
```diff
@@ -616,6 +617,17 @@ def view_partial_teams_list(request):
...
         )
 
+    if request.method == "POST":
+        form = SearchTeamForm(request.POST)
+
+        if form.is_valid():
+            search_term = form.cleaned_data["search_term"]
+
+            teams = Team.objects.filter(
+                active=True, name__istartswith=search_term
```
sheenarbw: Why not icontains?
And what if there are multiple words in the search term?

If there are multiple words then we should check if the name contains any of them. So we would be using Q and |

backend/frontend/tests/test_page_users_and_teams_nav.py
```diff
@@ -70,7 +70,54 @@ def test_load_more_button_appears(self):
         url = self.reverse_url("view_partial_teams_list")
         self.page.goto(url)
         self.page.wait_for_load_state()
-        
+
         body = self.page.locator("body")
 
         expect(body).to_contain_text("Load more")
+
+    def test_user_sees_teams_that_match_the_search_term(self):
```
sheenarbw: See if you can make the playwrite part of the testing as tight and simple as possible and then test the logic in a different way. Eg check that the frontend actually searches, then use normal unit tests to make sure that the search results are correct in different circumstances.

sheenarbw (changes requested):
Looks mostly good, but it needs some polish and I think we can make the tests better.

backend/frontend/tests/test_page_users_and_teams_nav.py
Ngoakor12: Not sure how to check that the frontend searches without making the API calls. Does this mean I need to mock the calls?

Ngoakor12: Nevermind, I removed the playwright test

kingraphaii (approved):

backend/frontend/templates/frontend/users_and_teams_nav/page.html
```diff
@@ -3,9 +3,13 @@
 {% block content %}
     <main>
         <div class="mx-auto max-w-7xl py-6 sm:px-6 lg:px-8">
-
-            <h1 class="{{ styles.heading1 }}">Teams</h1>
-            
+            <div class="flex gap-2 flex-col">
+                <h1 class="{{ styles.heading1 }} pb-2">Teams</h1>
+                <form method="post" action="" class="mb-3 flex items-center gap-1">
```
sheenarbw: This is what the different HTTP verbs mean:

GET: fetching and filtering
POST: Creating or updating
PUT: updating
DELETE: killing with fire
OPTIONS: header check

POST doesn't seem like the most appropriate thing here. We should always try to use things for their intended purpose.

backend/frontend/views.py
```diff
@@ -650,6 +651,15 @@ def view_partial_teams_list(request):
             user=user, perms=Team.PERMISSION_VIEW, klass=all_teams, any_perm=True
         )
 
+    if request.method == "POST":
```
sheenarbw: The algorithm here is weird:

1. get the teams that the user is allowed to view
2. if there is a post then totally ignore all the work we just did and get a new list of teams that doesn't take permissions into account at all

Better would be:

1. Get the teams
2. if the user is tiltering by name, filter the list by name
3. filter out what the user is not allowed to see 

So we should have one queryset and we should refine it over time instead of just starting over.

Right now any user can just filter by "c" to get a full list of cohorts.

sheenarbw (changes requested):
getting there, but this has a security issue and a few other problems

backend/frontend/views.py
Ngoakor12: Changed the algorithm a bit:
1. Get the teams
2. Filter the teams the user is allowed to see
3. Use the teams the user is allowed to see when filtering via name with form

backend/frontend/views.py
```diff
@@ -644,15 +645,24 @@ def view_partial_teams_list(request):
...
+        form = SearchTeamForm(request.POST)
+
+        if form.is_valid():
+            search_term = form.cleaned_data["search_term"]
+
+            permitted_teams = Team.get_teams_from_search_term(
+                search_term, permitted_teams
+            )
+            total_teams_count = permitted_teams.count()
```
kingraphaii: Since we now use `GET` in the form, we need to change this as well

kingraphaii (changes requested):

kingraphaii (approved):

sheenarbw (approved):
Nice!

PR #751, opened 2024-06-23
Add review modal

kingraphaii:

## Description:

## Screenshots/videos

image

## I solemnly swear that:


backend/frontend/views.py
```diff
@@ -460,6 +490,104 @@ def progress_details(
...
+                elif card.status == AgileCard.COMPLETE:
+                    log_creators.log_card_moved_to_complete(card, review.reviewer_user)
+
+                return render(
+                    request,
+                    "frontend/progress_details/js_exec_action_prepend_new_review.html",
+                    {
+                        "review": review,
+                    },
+                )
```
Ngoakor12: This method has the same name as another one just after it. If it is not a duplicate, then please rename it.

Ngoakor12:
[Previous feedback](https://github.com/Umuzi-org/Tilde/pull/725)

Ngoakor12 (changes requested):
image
- The reviews seem to be floating around, is there a way to make it a bit more consistent with the rest of the content on there? So maybe make use of borders to group the reviews or anything else that can make them look less out of place?
- The list of reviews also doesn't show any of the reviews
- The list is also nowhere to be found once I refresh

kingraphaii:
> The reviews seem to be floating around, is there a way to make it a bit more consistent with the rest of the content on there? So maybe make use of borders to group the reviews or anything else that can make them look less out of place?

Done

> The list of reviews also doesn't show any of the reviews

I removed some of this to keep the PR lean, old PR was 3 tasks in 1, so this is more focused on adding reviews only. The other two are dependant on this one.

> The list is also nowhere to be found once I refresh

See previous point

backend/frontend/templates/frontend/progress_details/page.html
```diff
@@ -120,7 +120,7 @@ <h6>Link to your work</h6>
                 </div>
             {% endif %}
 
-            <div id="reviews"></div>
+            <div id="reviews" class="border-2 p-2 rounded-md grid gap-2"></div>
```
Ngoakor12: If this PR is just for the form/modal, then:

- the recently added review doesn't need to show up because the one that shows is missing the actual comment
- the review container can be removed, plus it shows up in topics too which is weird

kingraphaii:
> If this PR is just for the form/modal, then:
> 
> * the recently added review doesn't need to show up because the one that shows is missing the actual comment

As per: 
If you use a modal to display the form then please use alpinejs
And once the form is submitted then **the new review should show up**.

> * the review container can be removed, plus it shows up in topics too which is weird

I have fixed it to only show for projects

Ngoakor12 (approved):

backend/curriculum_tracking/models.py
```diff
@@ -772,7 +772,38 @@ def reviewer_user_names(self):
...
+            return False
+
+        if user.is_superuser:
+            return True
+
+        if user in card.reviewers.all():
+            return True
+
+        return not card.request_user_is_assignee(user) and any(
+            card.user_has_permission(user, perm) for perm in Team.PERMISSION_VIEW
```
sheenarbw: This doesn't look quite right. You are saying that anyone who is not the assignee and who is allowed to see the card is allowed to review the card. The code can also be a bit more clear.

Rather:

```
if card.request_user_is_assignee(user):
    return False

return any(card.user_has_permission(user, perm) in [PERMISSION_REVIEW_CARDS,PERMISSION_TRUSTED_REVIEWER])
```

sheenarbw (changes requested):
Small bug

backend/frontend/templates/frontend/progress_details/partial_add_review.html
```diff
@@ -0,0 +1,16 @@
+{% extends 'frontend/partial_modal_base.html' %}
+{% load custom_tags %}
+{% block modal_header %}
+<div class="hidden md:block">{% user_avatar request.user %}</div>
```
sheenarbw, line 4: what is this for? Does it get unhidden?  Generating the avatar is a database hit. And then we hide it.

Ngoakor12 (approved):

sheenarbw (approved):

backend/frontend/templates/frontend/progress_details/partial_add_review.html
kingraphaii: On slightly smaller screens, the header looked a bit dense, so I though to hide the avatar as the screen gets smaller

PR #766, opened 2024-10-04
Freecodecamp marker

kingraphaii:

## Description:

here is the data it relies on: https://gist.github.com/kingraphaii/765509d416f587e38b367490755a8d88
It's a simple json file which the script needs somewhere on the filesystem. I had not decided on where it will reside yet.

## I solemnly swear that:


backend/curriculum_tracking/management/commands/mark_freecodecamp_projects.py
```diff
@@ -0,0 +1,180 @@
...
+        parser.add_argument(
+            "--elevated",
+            action="store_false",
+            help="If the bot should be elevated to a superuser",
+        )
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="If the bot should add reviews",
+        )
```
Ngoakor12: I don't get these arguments...

Would it not just be easier to have one that determines whether reviews are left? Maybe `prod` for leaving reviews and not leaving the reviews if the argument is not added? Or the opposite of `dry run`

I think this would also mean the bot is a superuser/trusted when the above argument is added. If not added the user is the bot_user.

Maybe I'm missing something

Ngoakor12 (changes requested):
For consistency and ease of maintance's sake, the rest of the hardcoded values should be stored in variables at the top. Same with env variables

backend/curriculum_tracking/management/commands/mark_freecodecamp_projects.py
kingraphaii: Initially, the automarker would add competent but untrusted reviews, prompting a human review to assess its efficacy. After verifying its competence, we’d run it as super to automatically close cards. NCIT cur marker does this, but it differs. I’ve reconsidered and believe we won’t need it.

